PR #20, opened 2023-08-15
Remove `@matrixai/quic` and `@matrixai/id` and separate error tree

CMCDragonkai:
### Description

Clean up after extraction.

### Tasks

- [x] 1. `@matrixai/quic` configuration should be encapsulated by the PK library, PK CLI should not have any knowledge about QUIC
- [x] 2. The `@matrixai/id` can be replaced by the new node ID utils `encodeNodeIdString`, `decodeNodeIdString` and `createNodeIdGenerator`.
- [x] 3. The error tree under `Polykey-CLI` should be separate. It should not be extending the library. It is its own thing. I believe it can be organised under `ErrorPolykeyCLI`.
  - And the naming of the errors should probably be simplified, with either a plain prefix of the relevant domains or just a suffix. I think in the case of CLI, a plain suffix is fine. Like `ErrorPolykeyCLI...`

### Final checklist

* ~Domain specific tests~
* ~Full tests~
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -23,7 +23,6 @@ const globals = {
   ),
   // Default asynchronous test timeout
   defaultTimeout: 20000,
-  polykeyStartupTimeout: 30000,
   failedConnectionTimeout: 50000,
   // Timeouts rely on setTimeout which takes 32 bit numbers
   maxTimeout: Math.pow(2, 31) - 1,
--- a/package-lock.json
+++ b/package-lock.json
@@ -7,12 +7,10 @@
     "": {
       "name": "polykey-cli",
       "version": "0.0.1",
-      "license": "Apache-2.0",
+      "license": "GPL-3.0",
       "dependencies": {
         "@matrixai/errors": "^1.1.7",
-        "@matrixai/id": "^3.3.6",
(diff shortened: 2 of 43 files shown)


CMCDragonkai:
Tasks 1. and 2. are waiting on the next prerelease @tegefaulkes.

CMCDragonkai:
Task 1. is pending an alignment of the `src/config.ts`.

tegefaulkes:
Ok so I've fixed  up the config usage and how agent is started. But this is not really testable until Polykey is in a fully working state. 

What is the scope of this PR? I've done task 1 now but the tests will still be broken.

CMCDragonkai:
The scope is whatever is necessary to get it into staging. Then just squash merge. Testing can be done on staging.

tegefaulkes:
It should be good then. Just need to squash and merge.

tegefaulkes:
This is done now. Tests still need to be fixed but with `Polykey` in an unstable state it would be hard to tell if it's a `CLi` or `Polykey` problem right now.

We'll be fixing tests in the staging branch or a separate PR.

Issue #25, opened 2023-10-05
Bundling and fixing pkg


tegefaulkes:
### Specification

This is an epic that tracks the progress of getting the CLI fully working.

I'll add the relevant issues to this epic after reviewing them.

### Additional context

* Commentary https://github.com/MatrixAI/Polykey-CLI/pull/4#issuecomment-1748016768
* Related: #23 
* Related: https://github.com/vercel/pkg-fetch/issues/285

### Tasks

1. Use a bundler to combine the source into a single js file.
2. Check that all artefacts that we need are included and any spurious stuff is left out.
3. Check that the single js file runs as expected using node.
4. package the single js file with `pkg` and ensure it runs properly.

tegefaulkes:
This really consists of two main parts. 
1. Selecting and applying a bundler and getting that working.
2. getting pkg to work with the bundled code and node20.
3. 

What bundler are we going to go with?

CMCDragonkai:
#23 and https://github.com/MatrixAI/TypeScript-Demo-Lib/issues/32

- esbuild

CMCDragonkai:
https://github.com/vercel/pkg-fetch/issues/285#issuecomment-1748056724

PR #26, opened 2023-10-06
Bundling and pkg fixes

### LINKED ISSUE #21: Update node2nix with the upstream fixes to building bin links
### Specification

The new node2nix finally fixes the bin links problems: https://github.com/svanderburg/node2nix/pull/302

It's time to check to see if this works for us and then remove our hack in `default.nix`

### Additional context

* https://github.com/svanderburg/node2nix/pull/302
* https://github.com/NixOS/nixpkgs/pull/193337

### Tasks

1. Update to the latest nixpkgs revision that contains this fix
2. Test if this works for TS demo lib
3. Merge and update all other repos

tegefaulkes:
### Description

This PR focueses on 3 things.

1. Using esbuild to bundle the polykey CLI.
2. Getting pkg to work with the bundled code.
3. Fixing all everything so that polykey CLI is in a workable state.

### Issues Fixed

* Fixes: #25 
* Fixes #21 

### Tasks

- [x] 1. Using `esbuild` to bundle the `Polykey CLI`.
- [x] 2. Getting pkg to work with the bundled code.
- [x] 3. Fixing all everything so that `Polykey CLI` is in a workable state.
- [x] 4. Replace `node2nix` with `buildNpmPackage`
- [x] 5. Replace `vercel/pkg` with `@yao-pkg/pkg`
- [x] 6. Make sure application, docker, pkg builds are building
- [x] 7. Ensure that CI is producing and uploading to ECR - will need to confirm everything on staging
- ~[ ] 8. Client verification of server certificate needs to be more robust~ - to be resolved here: https://github.com/MatrixAI/Polykey/pull/582#issuecomment-1765762674

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased


tegefaulkes:
Ok, I have modified the build script. The typescript compile stage is now in the prebuild script and compiles to the `./tmp/build` directory. The `build` script uses `esbuild` and the code in `./tmp/build` and outputs a single `polykey.js` file to the `./dist` directory.

`esbuild` can't resolve any dynamic imports, so it has a fun time with the native modules. In the case of `DB`, it's loading the `.node` modules from the directory `../../` which remains unchanged when bundled. The current fix for this is to specify these modules as `external` so that `esbuild` doesn't try to include these and change the path. That just means we'll need to tell `pkg` to include these as assets wholesale.

We can get a little more fancy with this. I think `esbuild` includes some ways to custom resolve paths, but I'm not sure it even recognises these as paths. Worst case we can be more specific with excluding and monkey patch some code and copy the files we need over.

Next steps. 
1. Make sure the worker script is included and working.
2. fix up pkg, may have to revert versions back to node 18 or 19 for this.

tegefaulkes:
Fixed the worker script problem. I just added a new `polykeyWorker.ts` that reexports the worker script from polykey. I then added this as a target for esbuild. This is then used as the worker script.

It's a little hacky, since the worker is loaded with `new Worker('./polykeyWorker')` in polykey. `esbuild` doesn't recognise it as an import so it's bundled verbatim. It will then attempt to load the the script from `./` when running the bundled code. Since we added our own it just loads that.

We can do similar things for the .node files. For example the db loads them from the `../../` directory. This is not ideal since after bundling that points outside of the root directory. But if we can monkey patch that string in the bundled code and copy over the required .node files then that should work.

tegefaulkes:
We'll worry about the `.node` files later.

Moving on to getting `pkg` working.

tegefaulkes:
After reverting to node 18, running pkg is working.

To move on from here I need to do a few things.

1. Might have to downgrade `Polykey` repo node version down to v18, and possible other dependencies need this applied as well.
2. I need a released version of `Polykey`, there is some nuance with how the `node_modules` are structured when linking locally.
3. I need to test the packaged file on some platforms to confirm that it works. 

I think at this point I can start working on fixing `Polykey` generally.

CMCDragonkai:
What about node 19?

tegefaulkes:
It was easiest to revert the changes back to 18 for now. We can try version 19 but I think that's low priority.

CMCDragonkai:
Staging fixes should be done here too.

CMCDragonkai:
If we are using the esbuild output in the docker image build. This means our docker image no longer requires all of the npm packages that was structured before. This means all the relevant JS code has been embedded into the single `polykey.js` output file.

The only thing that is required separately would be the worker script, and the `.node` native shared objects.

This should mean the image becomes MUCH smaller.

Especially if minification can work.

If minification doesn't work atm, just create an issue for it, we can revisit after we have deployed testnet 6.

CMCDragonkai:
A few things to consider here.

1. The `pkg` tool is only used as part of nix build scripts, and the expectation is that this becomes a single executable.
2. We have just introduced esbuild as an intermediate step to bundle up the JS code before going to `pkg`. This means `esbuild` execution should be part of the `release.nix` derivations right before `pkg`.
3. However `esbuild` output can also be useful for the docker image, as it can reduce the size of the docker image as well.
4. But the docker image right now uses `node2nix` and there's some derivations in `utils.nix` that sets that up. And this of course relies on `npm run build`.
5. In that case, if you incorporate `esbuild` into `npm run build` stage, that means the `dist` is no longer a regular npm package, and you'd want to move most dependencies to dev dependencies.

I think to ensure consistency - you'd want to incorporate esbuild into the nix scripts, not as part of `npm run build`.

CMCDragonkai:
Currently docker image doesn't really the esbuild output. It could continue using the regular dist.

But only pkg needs the esbuild output. I think we can incorporate it directly into the `release.nix`.

Will need to think about how to incorporate the steps separately for docker later.

CMCDragonkai:
I can have a look at this tomorrow. But please post up any build failure logs as to the reason for docker image build, if we can keep esbuild to the pkg build process, then docker image builds should be the same as before.

tegefaulkes:
Build is failing with the following.

<details>

```
> polykey-cli@0.0.1 prebuild
> shx rm -rf ./tmp/build && tsc -p ./tsconfig.build.json

src/utils/parsers.ts:46:7 - error TS2742: The inferred type of 'parseNodeId' cannot be named without a reference to 'polykey-cli/node_modules/polykey/dist/ids/types'. This is likely not portable. A type annotation is necessary.

46 const parseNodeId = validateParserToArgParser(validationUtils.parseNodeId);
         ~~~~~~~~~~~

src/utils/parsers.ts:47:7 - error TS2742: The inferred type of 'parseGestaltId' cannot be named without a reference to 'polykey-cli/node_modules/polykey/dist/ids/types'. This is likely not portable. A type annotation is necessary.

47 const parseGestaltId = validateParserToArgParser(
         ~~~~~~~~~~~~~~

src/utils/parsers.ts:53:7 - error TS2742: The inferred type of 'parseHost' cannot be named without a reference to 'polykey-cli/node_modules/polykey/dist/network/types'. This is likely not portable. A type annotation is necessary.

53 const parseHost = validateParserToArgParser(validationUtils.parseHost);
         ~~~~~~~~~

src/utils/parsers.ts:54:7 - error TS2742: The inferred type of 'parseHostname' cannot be named without a reference to 'polykey-cli/node_modules/polykey/dist/network/types'. This is likely not portable. A type annotation is necessary.

54 const parseHostname = validateParserToArgParser(validationUtils.parseHostname);
         ~~~~~~~~~~~~~

src/utils/parsers.ts:55:7 - error TS2742: The inferred type of 'parseHostOrHostname' cannot be named without a reference to 'polykey-cli/node_modules/polykey/dist/network/types'. This is likely not portable. A type annotation is necessary.

55 const parseHostOrHostname = validateParserToArgParser(
         ~~~~~~~~~~~~~~~~~~~

src/utils/parsers.ts:58:7 - error TS2742: The inferred type of 'parsePort' cannot be named without a reference to 'polykey-cli/node_modules/polykey/dist/network/types'. This is likely not portable. A type annotation is necessary.

58 const parsePort = validateParserToArgParser(validationUtils.parsePort);
         ~~~~~~~~~

src/utils/parsers.ts:59:7 - error TS2742: The inferred type of 'parseNetwork' cannot be named without a reference to 'polykey-cli/node_modules/polykey/dist/nodes/types'. This is likely not portable. A type annotation is necessary.

59 const parseNetwork = validateParserToArgParser(validationUtils.parseNetwork);
         ~~~~~~~~~~~~

src/utils/parsers.ts:60:7 - error TS2742: The inferred type of 'parseSeedNodes' cannot be named without a reference to 'polykey-cli/node_modules/polykey/dist/nodes/types'. This is likely not portable. A type annotation is necessary.

60 const parseSeedNodes = validateParserToArgParser(
         ~~~~~~~~~~~~~~

src/utils/parsers.ts:63:7 - error TS2742: The inferred type of 'parseProviderId' cannot be named without a reference to 'polykey-cli/node_modules/polykey/dist/ids/types'. This is likely not portable. A type annotation is necessary.

63 const parseProviderId = validateParserToArgParser(
         ~~~~~~~~~~~~~~~

src/utils/parsers.ts:66:7 - error TS2742: The inferred type of 'parseIdentityId' cannot be named without a reference to 'polykey-cli/node_modules/polykey/dist/ids/types'. This is likely not portable. A type annotation is necessary.

66 const parseIdentityId = validateParserToArgParser(
         ~~~~~~~~~~~~~~~

src/utils/parsers.ts:70:7 - error TS2742: The inferred type of 'parseProviderIdList' cannot be named without a reference to 'polykey-cli/node_modules/polykey/dist/ids/types'. This is likely not portable. A type annotation is necessary.

70 const parseProviderIdList = validateParserToArgListParser(
         ~~~~~~~~~~~~~~~~~~~


Found 11 errors in the same file, starting at: src/utils/parsers.ts:46

/nix/store/5s1yg5l36wzgy1dj0vv1ibarc4g7vrdr-stdenv-linux/setup: line 136: pop_var_context: head of shell_variables not a function context
error: builder for '/nix/store/l55ly6bxr44cq3s7k6a9bmbw7pgk81nd-polykey-cli-0.0.1.drv' failed with exit code 1;
       last 10 log lines:
       >
       > src/utils/parsers.ts:70:7 - error TS2742: The inferred type of 'parseProviderIdList' cannot be named without a reference to 'polykey-cli/node_modules/polykey/dist/ids/types'. This is likely not portable. A type annotation is necessary.
       >
       > 70 const parseProviderIdList = validateParserToArgListParser(
       >          ~~~~~~~~~~~~~~~~~~~
       >
       >
       > Found 11 errors in the same file, starting at: src/utils/parsers.ts:46
       >
       > /nix/store/5s1yg5l36wzgy1dj0vv1ibarc4g7vrdr-stdenv-linux/setup: line 136: pop_var_context: head of shell_variables not a function context
       For full logs, run 'nix log /nix/store/l55ly6bxr44cq3s7k6a9bmbw7pgk81nd-polykey-cli-0.0.1.drv'.
error: 1 dependencies of derivation '/nix/store/ylr6j3l8wnncdyzsby2bzn499zbv6i1i-polykey-cli-0.0.1.drv' failed to build

```

</details>

Doing prebuild normally runs fine. Not sure why it's an issue in `nix-build`.

Simplest fix would be to add the type annotations in but maybe not the best fix.

tegefaulkes:
Right now fixing up the tests and getting the build working can be done in parallel. It's possible to build and do a `Polykey agent start` just fine so that's the minimum needed for testing building and pkg.

What's broken in tests right now is how errors are serialised and deseralized. It's something I'm working on int `js-rpc`. As part of this I'm making the `toError` responsible for wrapping the errors in a `ErrorXRemote` error.

CMCDragonkai:
> Build is failing with the following.
> 
> ```
> > polykey-cli@0.0.1 prebuild
> > shx rm -rf ./tmp/build && tsc -p ./tsconfig.build.json
> 
> src/utils/parsers.ts:46:7 - error TS2742: The inferred type of 'parseNodeId' cannot be named without a reference to 'polykey-cli/node_modules/polykey/dist/ids/types'. This is likely not portable. A type annotation is necessary.
> 
> 46 const parseNodeId = validateParserToArgParser(validationUtils.parseNodeId);
>          ~~~~~~~~~~~
> 
> src/utils/parsers.ts:47:7 - error TS2742: The inferred type of 'parseGestaltId' cannot be named without a reference to 'polykey-cli/node_modules/polykey/dist/ids/types'. This is likely not portable. A type annotation is necessary.
> 
> 47 const parseGestaltId = validateParserToArgParser(
>          ~~~~~~~~~~~~~~
> 
> src/utils/parsers.ts:53:7 - error TS2742: The inferred type of 'parseHost' cannot be named without a reference to 'polykey-cli/node_modules/polykey/dist/network/types'. This is likely not portable. A type annotation is necessary.
> 
> 53 const parseHost = validateParserToArgParser(validationUtils.parseHost);
>          ~~~~~~~~~
> 
> src/utils/parsers.ts:54:7 - error TS2742: The inferred type of 'parseHostname' cannot be named without a reference to 'polykey-cli/node_modules/polykey/dist/network/types'. This is likely not portable. A type annotation is necessary.
> 
> 54 const parseHostname = validateParserToArgParser(validationUtils.parseHostname);
>          ~~~~~~~~~~~~~
> 
> src/utils/parsers.ts:55:7 - error TS2742: The inferred type of 'parseHostOrHostname' cannot be named without a reference to 'polykey-cli/node_modules/polykey/dist/network/types'. This is likely not portable. A type annotation is necessary.
> 
> 55 const parseHostOrHostname = validateParserToArgParser(
>          ~~~~~~~~~~~~~~~~~~~
> 
> src/utils/parsers.ts:58:7 - error TS2742: The inferred type of 'parsePort' cannot be named without a reference to 'polykey-cli/node_modules/polykey/dist/network/types'. This is likely not portable. A type annotation is necessary.
> 
> 58 const parsePort = validateParserToArgParser(validationUtils.parsePort);
>          ~~~~~~~~~
> 
> src/utils/parsers.ts:59:7 - error TS2742: The inferred type of 'parseNetwork' cannot be named without a reference to 'polykey-cli/node_modules/polykey/dist/nodes/types'. This is likely not portable. A type annotation is necessary.
> 
> 59 const parseNetwork = validateParserToArgParser(validationUtils.parseNetwork);
>          ~~~~~~~~~~~~
> 
> src/utils/parsers.ts:60:7 - error TS2742: The inferred type of 'parseSeedNodes' cannot be named without a reference to 'polykey-cli/node_modules/polykey/dist/nodes/types'. This is likely not portable. A type annotation is necessary.
> 
> 60 const parseSeedNodes = validateParserToArgParser(
>          ~~~~~~~~~~~~~~
> 
> src/utils/parsers.ts:63:7 - error TS2742: The inferred type of 'parseProviderId' cannot be named without a reference to 'polykey-cli/node_modules/polykey/dist/ids/types'. This is likely not portable. A type annotation is necessary.
> 
> 63 const parseProviderId = validateParserToArgParser(
>          ~~~~~~~~~~~~~~~
> 
> src/utils/parsers.ts:66:7 - error TS2742: The inferred type of 'parseIdentityId' cannot be named without a reference to 'polykey-cli/node_modules/polykey/dist/ids/types'. This is likely not portable. A type annotation is necessary.
> 
> 66 const parseIdentityId = validateParserToArgParser(
>          ~~~~~~~~~~~~~~~
> 
> src/utils/parsers.ts:70:7 - error TS2742: The inferred type of 'parseProviderIdList' cannot be named without a reference to 'polykey-cli/node_modules/polykey/dist/ids/types'. This is likely not portable. A type annotation is necessary.
> 
> 70 const parseProviderIdList = validateParserToArgListParser(
>          ~~~~~~~~~~~~~~~~~~~
> 
> 
> Found 11 errors in the same file, starting at: src/utils/parsers.ts:46
> 
> /nix/store/5s1yg5l36wzgy1dj0vv1ibarc4g7vrdr-stdenv-linux/setup: line 136: pop_var_context: head of shell_variables not a function context
> error: builder for '/nix/store/l55ly6bxr44cq3s7k6a9bmbw7pgk81nd-polykey-cli-0.0.1.drv' failed with exit code 1;
>        last 10 log lines:
>        >
>        > src/utils/parsers.ts:70:7 - error TS2742: The inferred type of 'parseProviderIdList' cannot be named without a reference to 'polykey-cli/node_modules/polykey/dist/ids/types'. This is likely not portable. A type annotation is necessary.
>        >
>        > 70 const parseProviderIdList = validateParserToArgListParser(
>        >          ~~~~~~~~~~~~~~~~~~~
>        >
>        >
>        > Found 11 errors in the same file, starting at: src/utils/parsers.ts:46
>        >
>        > /nix/store/5s1yg5l36wzgy1dj0vv1ibarc4g7vrdr-stdenv-linux/setup: line 136: pop_var_context: head of shell_variables not a function context
>        For full logs, run 'nix log /nix/store/l55ly6bxr44cq3s7k6a9bmbw7pgk81nd-polykey-cli-0.0.1.drv'.
> error: 1 dependencies of derivation '/nix/store/ylr6j3l8wnncdyzsby2bzn499zbv6i1i-polykey-cli-0.0.1.drv' failed to build
> ```
> 
> Doing prebuild normally runs fine. Not sure why it's an issue in `nix-build`.
> 
> Simplest fix would be to add the type annotations in but maybe not the best fix.

It's possible solving https://github.com/MatrixAI/Polykey/issues/532 will have a side-effect of solving this problem.

CMCDragonkai:
So you have 2 problems right now:

1. You're fixing tests for PK CLI - you need to use 1.2.1-alpha.3 that I just released with the new client domain. Or you can link it to the current staging right now.
2. The docker build should not be using the esbuild step. Move the esbuild step to the pkg outputs in `release.nix`. We can figure out the `esbuild` docker output integration later.

I can look at 2. now.

src/agent/CommandLockAll.ts
```diff
@@ -19,9 +19,8 @@ class CommandLockAll extends CommandPolykey {
       const { default: PolykeyClient } = await import(
         'polykey/dist/PolykeyClient'
       );
-      const { default: WebSocketClient } = await import(
-        'polykey/dist/websockets/WebSocketClient'
-      );
+      const { WebSocketClient } = await import('@matrixai/ws');
+      const clientUtils = await import('polykey/dist/client/utils');
```
CMCDragonkai, line 23: Seems problematic...

Surely it's possible to expose the necessary abstractions directly from PK?

src/agent/CommandLockAll.ts
```diff
@@ -46,17 +45,27 @@ class CommandLockAll extends CommandPolykey {
...
+              await clientUtils.verifyServerCertificateChain(
+                [clientOptions.nodeId],
+                certs,
+              );
+            },
+          },
           host: clientOptions.clientHost,
           port: clientOptions.clientPort,
           logger: this.logger.getChild(WebSocketClient.name),
         });
```
CMCDragonkai, line 66: Shouldn't `PolykeyClient` encapsulate the WS client setup. I don't like having every command setup its own websocket transport. That seems very incorrect.

src/agent/CommandLockAll.ts
```diff
@@ -46,17 +45,27 @@ class CommandLockAll extends CommandPolykey {
...
         pkClient = await PolykeyClient.createPolykeyClient({
-          streamFactory: (ctx) => webSocketClient.startConnection(ctx),
+          streamFactory: () => webSocketClient.connection.newStream(),
```
CMCDragonkai, line 68: It does not make sense. `PolykeyClient` should be doing this all automatically.

src/agent/CommandLockAll.ts
tegefaulkes: I think we wanted `PolykeyClient` to be agnostic to transport like the RPC?

CMCDragonkai:
To deal with docker build I need to try solve https://github.com/MatrixAI/Polykey/issues/532, then you can try and see if that fixes it.

src/agent/CommandLockAll.ts
CMCDragonkai: No of course not!

CMCDragonkai: How can PK client be agnostic, and yet PK agent be not agnostic. Doesn't make sense.

tegefaulkes:
I had to disable 2 tests for `ping.test.ts`, these are testing failing ping. It's throwing an error when it tries to refresh the timeout timer after it has timed out. I'll need to fix the RPC to ignore that since it's fully allowed to keep running after a timeout.

CMCDragonkai:
> I had to disable 2 tests for `ping.test.ts`, these are testing failing ping. It's throwing an error when it tries to refresh the timeout timer after it has timed out. I'll need to fix the RPC to ignore that since it's fully allowed to keep running after a timeout.

What does this have to do with RPC?

tegefaulkes:
It's the timer is being refreshed inside the RPC logic. It's being refreshed after timing out and thus throwing an error.

tegefaulkes:
I'm disabling `'concurrent with bootstrap results in 1 success'` in `start.test.ts`. Both the start and bootstrap are succeeding here. It's not worth looking into right now.

CMCDragonkai:
> I'm disabling `'concurrent with bootstrap results in 1 success'` in `start.test.ts`. Both the start and bootstrap are succeeding here. It's not worth looking into right now.

Why is this failing?

tegefaulkes:
Ok, all tests are passing now except for those 3 tests I disabled.

I need to write a test and fix for `js-rpc` dealing with messages sent and received after the handler has timed out.

tegefaulkes:
> > I'm disabling `'concurrent with bootstrap results in 1 success'` in `start.test.ts`. Both the start and bootstrap are succeeding here. It's not worth looking into right now.
> 
> Why is this failing?

The test starts a polykey agent and runs a bootstrap concurrently. It expects that one of these will fail but both are succeeding. If I had to guess its bootstrapping and then starting without conflicting as a race condition. A failure of the test itself essentially.

tegefaulkes:
I pushed some fixes to `Polykey` and doing another pre-release. The next version should be `1.2.1-alpha.4`.

tegefaulkes:
Ok, after doing a prerelease on `Polykey` and using that version for testing we are getting failure. Seems that `Polykey` is failing to start now. here is the output.

```
[faulkes@faulkeswork:~/matrixcode/polykey/Polykey-CLI]$ npm run polykey -- agent start --node-path ./tmp/sdfghdfasd -v

> polykey-cli@0.0.1 polykey
> ts-node src/polykey.ts agent start --node-path ./tmp/sdfghdfasd -v

✔ Enter new password … ********
✔ Confirm new password … ********
INFO:polykey.PolykeyAgent:Creating PolykeyAgent
INFO:polykey.PolykeyAgent:Setting umask to 077
INFO:polykey.PolykeyAgent:Setting node path to ./tmp/sdfghdfasd
INFO:polykey.PolykeyAgent.Status:Starting Status
INFO:polykey.PolykeyAgent.Status:Writing Status to tmp/sdfghdfasd/status.json
INFO:polykey.PolykeyAgent.Status:Status is STARTING
INFO:polykey.PolykeyAgent.Schema:Creating Schema
INFO:polykey.PolykeyAgent.Schema:Starting Schema
INFO:polykey.PolykeyAgent.Schema:Setting state path to tmp/sdfghdfasd/state
INFO:polykey.PolykeyAgent.Schema:Started Schema
INFO:polykey.PolykeyAgent.Schema:Created Schema
INFO:polykey.PolykeyAgent.KeyRing:Creating KeyRing
INFO:polykey.PolykeyAgent.KeyRing:Setting keys path to tmp/sdfghdfasd/state/keys
INFO:polykey.PolykeyAgent.KeyRing:Starting KeyRing
INFO:polykey.PolykeyAgent.KeyRing:Checking tmp/sdfghdfasd/state/keys/private.jwk
INFO:polykey.PolykeyAgent.KeyRing:Generating root key pair and recovery code
INFO:polykey.PolykeyAgent.KeyRing:Checking tmp/sdfghdfasd/state/keys/db.jwk
INFO:polykey.PolykeyAgent.KeyRing:Generating db key
INFO:polykey.PolykeyAgent.KeyRing:Started KeyRing
INFO:polykey.PolykeyAgent.KeyRing:Created KeyRing
INFO:polykey.PolykeyAgent.DB:Creating DB
INFO:polykey.PolykeyAgent.DB:Starting DB
INFO:polykey.PolykeyAgent.DB:Setting DB path to tmp/sdfghdfasd/state/db
INFO:polykey.PolykeyAgent.DB:Started DB
INFO:polykey.PolykeyAgent.DB:Created DB
INFO:polykey.PolykeyAgent:Creating TaskManager
INFO:polykey.PolykeyAgent:Starting TaskManager in Lazy Mode
INFO:polykey.PolykeyAgent:Begin Tasks Repair
INFO:polykey.PolykeyAgent:Finish Tasks Repair
INFO:polykey.PolykeyAgent:Started TaskManager
INFO:polykey.PolykeyAgent:Created TaskManager
INFO:polykey.PolykeyAgent.CertManager:Creating CertManager
INFO:polykey.PolykeyAgent.CertManager:Starting CertManager
INFO:polykey.PolykeyAgent.CertManager:Begin current certificate setup
INFO:polykey.PolykeyAgent.CertManager:Generating new current certificate
WARN:polykey.PolykeyAgent:Failed Creating PolykeyAgent
INFO:polykey.PolykeyAgent:Stopping TaskManager
INFO:polykey.PolykeyAgent:Stopped TaskManager
INFO:polykey.PolykeyAgent.DB:Stopping DB
INFO:polykey.PolykeyAgent.DB:Stopped DB
INFO:polykey.PolykeyAgent.KeyRing:Stopping KeyRing
INFO:polykey.PolykeyAgent.KeyRing:Stopped KeyRing
INFO:polykey.PolykeyAgent.Schema:Stopping Schema
INFO:polykey.PolykeyAgent.Schema:Stopped Schema
INFO:polykey.PolykeyAgent.Status:Stopping Status
INFO:polykey.PolykeyAgent.Status:Writing Status to tmp/sdfghdfasd/status.json
INFO:polykey.PolykeyAgent.Status:Status is DEAD
Error: Cannot initialize GeneralNames. Incorrect incoming arguments

```

I'm guessing the last line before the problem happens is `INFO:polykey.PolykeyAgent.CertManager:Generating new current certificate` it means that the certificate generation is failing.

I recently removed the `webcrypto` polyfill monkey patch since it wasn't required in node20, but since we're running in node18 here that might be a problem? The only other difference that could be the issue here is that the `node_module` dependencies could be different without linking `polykey`.

CMCDragonkai:
> To deal with docker build I need to try solve [MatrixAI/Polykey#532](https://github.com/MatrixAI/Polykey/issues/532), then you can try and see if that fixes it.

PR to this is here https://github.com/MatrixAI/Polykey/pull/573.

CMCDragonkai:
https://github.com/MatrixAI/Polykey/pull/573 is merged. You'll need to update all the relevant parser locations.

```
secrets/CommandUpdate.ts:22:      parsers.parseSecretPath,
secrets/CommandRename.ts:17:      parsers.parseSecretPath,
secrets/CommandMkdir.ts:17:      parsers.parseSecretPath,
secrets/CommandDelete.ts:18:      parsers.parseSecretPath,
secrets/CommandCreate.ts:22:      parsers.parseSecretPath,
secrets/CommandGet.ts:17:      parsers.parseSecretPath,
secrets/CommandStat.ts:17:      parsers.parseSecretPath,
secrets/CommandEdit.ts:18:      parsers.parseSecretPath,
utils/options.ts:54:  .argParser(binParsers.parseNodeId);
utils/options.ts:64:  .argParser(binParsers.parseHost);
utils/options.ts:74:  .argParser(binParsers.parsePort);
utils/options.ts:78:  .argParser(binParsers.parseHost)
utils/options.ts:83:  .argParser(binParsers.parsePort)
utils/options.ts:90:  .argParser(binParsers.parseInteger)
utils/options.ts:127:  .argParser(binParsers.parseSeedNodes)
utils/options.ts:135:  .argParser(binParsers.parseNetwork)
utils/options.ts:143:  .argParser(binParsers.parseCoreCount)
nodes/CommandAdd.ts:16:    this.argument('<nodeId>', 'Id of the node to add', binParsers.parseNodeId);
nodes/CommandAdd.ts:17:    this.argument('<host>', 'Address of the node', binParsers.parseHost);
nodes/CommandAdd.ts:18:    this.argument('<port>', 'Port of the node', binParsers.parsePort);
nodes/CommandClaim.ts:18:      binParsers.parseNodeId,
nodes/CommandPing.ts:16:    this.argument('<nodeId>', 'Id of the node to ping', binParsers.parseNodeId);
nodes/CommandFind.ts:17:    this.argument('<nodeId>', 'Id of the node to find', binParsers.parseNodeId);
identities/CommandAllow.ts:18:      binParsers.parseGestaltId,
identities/CommandAllow.ts:23:      binParsers.parseGestaltAction,
identities/CommandDisallow.ts:18:      parsers.parseGestaltId,
identities/CommandDisallow.ts:23:      parsers.parseGestaltAction,
identities/CommandAuthenticated.ts:17:      parsers.parseProviderId,
identities/CommandInvite.ts:18:      binParsers.parseNodeId,
identities/CommandUntrust.ts:18:      parsers.parseGestaltId,
identities/CommandClaim.ts:17:      parsers.parseProviderId,
identities/CommandClaim.ts:22:      parsers.parseIdentityId,
identities/CommandSearch.ts:28:      parsers.parseIdentityId,
identities/CommandSearch.ts:33:      parsers.parseIdentityId,
identities/CommandSearch.ts:42:      parsers.parseInteger,
identities/CommandTrust.ts:18:      parsers.parseGestaltId,
identities/CommandGet.ts:21:      parsers.parseGestaltId,
identities/CommandDiscover.ts:18:      parsers.parseGestaltId,
identities/CommandPermissions.ts:18:      parsers.parseGestaltId,
identities/CommandAuthenticate.ts:21:      parsers.parseProviderId,
notifications/CommandSend.ts:18:      binParsers.parseNodeId,
vaults/CommandShare.ts:19:      binParsers.parseNodeId,
vaults/CommandPull.ts:19:      binParsers.parseNodeId,
vaults/CommandUnshare.ts:19:      binParsers.parseNodeId,
vaults/CommandClone.ts:19:      binParsers.parseNodeId,
```

CMCDragonkai:
Then to review the docker build.

CMCDragonkai:
Can @amydevs help here?

src/agent/CommandLockAll.ts
CMCDragonkai: @amydevs can you help on this problem?

CMCDragonkai:
Will be working https://github.com/MatrixAI/Polykey-CLI/pull/26#issuecomment-1759122081

CMCDragonkai:
@amydevs will be validating the 6 steps. Make sure to use this branch.

CMCDragonkai:
So for the parsers, we have decentralized it now, so parsers aren't all just coming from `validationUtils`.

It's easy enough to change this to bring it from the relevant places, but this means `src/utils/parsers.ts` becomes a point that brings everything in, which goes against the point of dynamic imports that occur later.

src/agent/CommandLockAll.ts
CMCDragonkai: Ok once `WebSocketClient` is moved into the `PolykeyClient`. That should mean `@matrixai/ws` should be removed from the `package.json`.

CMCDragonkai:
The docker image build works:

```
[nix-shell:~/Projects/Polykey-CLI]$ ll /nix/store/q8cj0kq4zjx42sypbq55xlgs35dg1h7d-docker-image-polykey-cli-0.0.1.tar.gz
-r--r--r-- 2 root root 119M Jan  1  1970 /nix/store/q8cj0kq4zjx42sypbq55xlgs35dg1h7d-docker-image-polykey-cli-0.0.1.tar.gz

[nix-shell:~/Projects/Polykey-CLI]$ docker images
REPOSITORY   TAG       IMAGE ID   CREATED   SIZE

[nix-shell:~/Projects/Polykey-CLI]$ docker load --input ./re
README.md    release.nix  result@      

[nix-shell:~/Projects/Polykey-CLI]$ docker load --input ./result
5ce83db80d6b: Loading layer  370.1MB/370.1MB
Loaded image: polykey-cli-0.0.1:q8cj0kq4zjx42sypbq55xlgs35dg1h7d

[nix-shell:~/Projects/Polykey-CLI]$ docker run -it polykey-cli-0.0.1:q8cj0kq4zjx42sypbq55xlgs35dg1h7d
Usage: polykey [options] [command]

Polykey CLI

Options:
  -np, --node-path <path>      Path to Node State (default: "/.local/share/polykey", env:
                               PK_NODE_PATH)
  -V, --version                output the version number

Commands:
  bootstrap [options]          Bootstrap Keynode State
  agent [options]              Agent Operations
  nodes [options]              Nodes Operations
  secrets [options]            Secrets Operations
  keys [options]               Keys Operations
  vaults [options]             Vaults Operations
  identities [options]         Identities Operations
  notifications [options]      Notifications Operations
  help [command]               display help for command
```

CMCDragonkai:
We can separate out now.

@tegefaulkes to test out the docker image, and push this up to testnet.

I'm going to figure out the pkg, esbuild situation.

CMCDragonkai:
Bundling step now works. `npm run bundle` will produce a bundle inside `./bundle` which is ignored in git.

```

[nix-shell:~/Projects/Polykey-CLI]$ npm run bundle

> polykey-cli@0.0.1 bundle
> shx rm -rf ./bundle && node ./scripts/bundle.js


[nix-shell:~/Projects/Polykey-CLI]$ tree bundle
bundle
├── polykey.js
├── polykey.js.map
├── polykeyWorker.js
└── polykeyWorker.js.map

0 directories, 4 files
```

CMCDragonkai:
I've discovered 2 issues:

Minification seems to partly work but the resulting bundle seems to have weird errors. So disabled for now.

Repeat start stop of the Polykey agent results in this:

```
[nix-shell:~/Projects/Polykey-CLI]$ ./esbuild/polykey.js agent start
✔ Please enter the password … ******
pid    2621463
nodeId    "vbi7fg856ldaaunmusrcqs91ugo0c7v993v5e2q272vek8dvnlqqg"
clientHost    "::1"
clientPort    36673
agentHost    "::"
agentPort    33964
ERROR:polykey.PolykeyAgent.scheduler:Failed scheduling loop TypeError: The first argument must be of type string or an instance of Buffer, ArrayBuffer, or Array or an Array-like Object. Received an instance of Object
ErrorTaskManagerScheduler: TaskManager scheduling loop encountered an unrecoverable error
  cause: TypeError: The first argument must be of type string or an instance of Buffer, ArrayBuffer, or Array or an Array-like Object. Received an instance of Object
```

Can't really solve that for now, will need to fix this later.

CMCDragonkai:
I'm going to try `pkg` now, by integrating `npm run bundle` into the `release.nix` that produces the pkg executables.

CMCDragonkai:
So the `package.json` `pkg` spec is just the static spec before which `scripts/pkg.js` will take and augment it by dynamically finding all the `.node` files.

The result looks like:

```
  {
    "assets": [
      "node_modules/tslib/**/*.js",
      "node_modules/tsyringe/**/*.js",
      "node_modules/@matrixai/db/prebuilds/linux-x64/node.napi.node",
      "node_modules/fd-lock/prebuilds/linux-x64/node.napi.node",
      "node_modules/sodium-native/prebuilds/linux-x64/node.napi.node"
    ],
    "scripts": [
      "dist/polykeyWorker.js"
    ]
  }   
```

Notice this is missing the quic `.node` file cause quic uses a new system (as well as using optional packages) and node-gyp-build cannot find that.

I'll need to update the `pkg.js` script to find the new one, because we will transition to the new system for `@matrixai/db` and any other native packages eventually. (Maybe even koffi).

CMCDragonkai:
The pkg executable is now created from the bundle. It has shaved off 10 MiB, but much faster to make.

```
[nix-shell:~/Projects/Polykey-CLI]$ ll out 
-rwxr-x--x 1 cmcdragonkai operators 68M Oct 13 16:22 out*

[nix-shell:~/Projects/Polykey-CLI]$ file out 
out: ELF 64-bit LSB executable, x86-64, version 1 (GNU/Linux), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, BuildID[sha1]=e8e9fc449943f935d469333b1101f224a4c0164f, for GNU/Linux 2.6.32, stripped
```

I need to test this on Ubuntu to make it work.

CMCDragonkai:
So a few issues.

1. `pkg.js` auto finds `.node` through `node-gyp-build`, however this does not work for `js-quic` because it's not using it. I have to maually add this into `package.json` `pkg`. Cannot automate the finding atm.
2. The `entrypoint` has to be `bundle/polykey.js` instead of `dist/polykey.js`. I need to incorporate this into `pkg.js` script and `release.nix`. I think the `--bin` needs to change to the full path that I want.

Still plenty of optimisations available. The `esbuild` externals isn't as optimised, we want `@matrixai/db` to start using the optional packages too.. so it can made external. And then every other external native package we'd probably have to fork to make it as optimal as possible.

tegefaulkes:
Progress report. The image has been uploaded to the Elastic contain registry on AWS. The task has started using this image so far. The problem currently is that the task crashes when doing a password hash.

```
{"type":"ErrorPasswordHash","data":{"message":"","timestamp":"2023-10-13T04:42:37.909Z","data":{},"cause":{},"stack":"ErrorPasswordHash\n at hashPassword (/lib/node_modules/polykey-cli/node_modules/polykey/dist/keys/utils/password.js:89:15)\n at Object.wrapWithPassword (/lib/node_modules/polykey-cli/node_modules/polykey/dist/keys/utils/symmetric.js:161:53)\n at constructor_.writeKeyPair (/lib/node_modules/polykey-cli/node_modules/polykey/dist/keys/KeyRing.js:545:38)\n at constructor_.setupKeyPair (/lib/node_modules/polykey-cli/node_modules/polykey/dist/keys/KeyRing.js:376:28)\n at async constructor_.start (/lib/node_modules/polykey-cli/node_modules/polykey/dist/keys/KeyRing.js:106:41)\n at async /lib/node_modules/polykey-cli/node_modules/@matrixai/async-init/dist/CreateDestroyStartStop.js:88:34\n at async withF (/lib/node_modules/polykey-cli/node_modules/@matrixai/resources/dist/utils.js:17:16)\n at async KeyRing.createKeyRing (/lib/node_modules/polykey-cli/node_modules/polykey/dist/keys/KeyRing.js:59:9)\n at async PolykeyAgent.createPolykeyAgent (/lib/node_modules/polykey-cli/node_modules/polykey/dist/PolykeyAgent.js:170:23)\n at async /lib/node_modules/polykey-cli/dist/agent/CommandStart.js:190:31","description":"Failed to hash password","exitCode":69}}
```

We're assuming that the hash is running out of memory when doing it's thing.

Right now I'm working on modifying the `Polykey-Infrastructure` script to include `ENV` config that tells `Polykey` to us the minimal password has memory and ops.

Current stumbling block is that the script is not working. Still looking into it.

CMCDragonkai:
The `nix-build ./release.nix -A package.linux.x64.elf` should be working now.

CMCDragonkai:
```
function hashPassword(
  password: string,
  salt?: PasswordSalt,
  opsLimit: PasswordOpsLimit = passwordOpsLimitDefault,
  memLimit: PasswordMemLimit = passwordMemLimitDefault,
): [PasswordHash, PasswordSalt] {
  const hash = Buffer.allocUnsafeSlow(
    sodium.crypto_aead_xchacha20poly1305_ietf_KEYBYTES,
  );
  salt ??= getRandomBytes(
    sodium.crypto_pwhash_SALTBYTES,
    undefined,
    false,
  ) as PasswordSalt;
  try {
    sodium.crypto_pwhash(
      hash,
      Buffer.from(password, 'utf-8'),
      salt,
      opsLimit,
      memLimit,
      sodium.crypto_pwhash_ALG_ARGON2ID13,
    );
    return [hash as PasswordHash, salt];
  } catch (e) {
    throw new keysErrors.ErrorPasswordHash(undefined, { cause: e });
  }
}
```

This is failing on AWS. 

```
{"type":"ErrorPasswordHash","data":{"message":"","timestamp":"2023-10-13T05:43:56.111Z","data":{},"cause":{},"stack":"ErrorPasswordHash\n at hashPassword (/lib/node_modules/polykey-cli/node_modules/polykey/dist/keys/utils/password.js:89:15)\n at Object.wrapWithPassword (/lib/node_modules/polykey-cli/node_modules/polykey/dist/keys/utils/symmetric.js:161:53)\n at constructor_.writeKeyPair (/lib/node_modules/polykey-cli/node_modules/polykey/dist/keys/KeyRing.js:545:38)\n at constructor_.setupKeyPair (/lib/node_modules/polykey-cli/node_modules/polykey/dist/keys/KeyRing.js:376:28)\n at async constructor_.start (/lib/node_modules/polykey-cli/node_modules/polykey/dist/keys/KeyRing.js:106:41)\n at async /lib/node_modules/polykey-cli/node_modules/@matrixai/async-init/dist/CreateDestroyStartStop.js:88:34\n at async withF (/lib/node_modules/polykey-cli/node_modules/@matrixai/resources/dist/utils.js:17:16)\n at async KeyRing.createKeyRing (/lib/node_modules/polykey-cli/node_modules/polykey/dist/keys/KeyRing.js:59:9)\n at async PolykeyAgent.createPolykeyAgent (/lib/node_modules/polykey-cli/node_modules/polykey/dist/PolykeyAgent.js:170:23)\n at async /lib/node_modules/polykey-cli/dist/agent/CommandStart.js:190:31","description":"Failed to hash password","exitCode":69}}
```

The cause is an empty object.

Check if the password is being set.

CMCDragonkai:
The produced ELF doesn't work on Ubuntu atm.

```
vboxuser@Ubuntu-Test:~/Downloads$ ./result 
/snapshot/polykey-cli/bundle/polykey.js:66268
      throw new Error(`Failed requiring possible native bindings: ${prebuildTargets.concat(npmTargets)}`);
      ^

Error: Failed requiring possible native bindings: /snapshot/prebuild/quic-linux-x64.node,@matrixai/quic-linux-x64
    at requireBinding (/snapshot/polykey-cli/bundle/polykey.js:66268:13)
    at node_modules/@matrixai/quic/dist/native/quiche.js (/snapshot/polykey-cli/bundle/polykey.js:66310:29)
    at __require (/snapshot/polykey-cli/bundle/polykey.js:13:50)
    at node_modules/@matrixai/quic/dist/native/index.js (/snapshot/polykey-cli/bundle/polykey.js:66437:20)
    at __require (/snapshot/polykey-cli/bundle/polykey.js:13:50)
    at node_modules/@matrixai/quic/dist/QUICSocket.js (/snapshot/polykey-cli/bundle/polykey.js:67240:20)
    at __require (/snapshot/polykey-cli/bundle/polykey.js:13:50)
    at node_modules/@matrixai/quic/dist/index.js (/snapshot/polykey-cli/bundle/polykey.js:70045:24)
    at __require (/snapshot/polykey-cli/bundle/polykey.js:13:50)
    at node_modules/polykey/dist/nodes/utils.js (/snapshot/polykey-cli/bundle/polykey.js:73986:18)

Node.js v18.5.0
```

I'm getting that.

Trying to update to the latest pkg release of 3.5... but it's kind of weird why it's not realising it.

CMCDragonkai:
Remaining issues in summary.

1. On testnet, we still have a hash password problem https://github.com/MatrixAI/Polykey-CLI/pull/26#issuecomment-1760971608 - so currently the testnet nodes are just failing to run.
2. The built pkg executable doesn't work on Ubuntu https://github.com/MatrixAI/Polykey-CLI/pull/26#issuecomment-1760993275 and we aren't able to update to pkg-fetch 3.5, node version 18.15 due to no new release on vercel/pkg
3. Repeated start/stop results in a task manager error: https://github.com/MatrixAI/Polykey-CLI/pull/26#issuecomment-1760765602
4. The esbuild minified output doesn't work.

Anything else...?

addievo:
More Issues

1. Gestalts discovery based on identity not working atm.
2. Gestalt identity search from identity providers is slow as its polling for every matching identity.

@amydevs is currently working on 1. 

For 2, we have a couple solutions, i. is to use Promise.all to fetch all identities at once/2 is to use graphql  HTTP.

addievo:
1. On testnet, we still have a hash password problem https://github.com/MatrixAI/Polykey-CLI/pull/26#issuecomment-1760971608 - so currently the testnet nodes are just failing to run.

Currently we think that this is due to the fact that the VPC does not have enough mem to support hashing algorithms, currently proposed solution is to simply increase the VPC memory to allow the hashing algorithm to run.

addievo:
The PolykeyClient is supposed to encapsulate WebSocketClient, so that needs to be done on Polykey.
is fixed in https://github.com/MatrixAI/Polykey/pull/582

addievo:
> The identities authenticate right has a too low timeout, it should allow for a longer time.

Changed timeout to 1000000 in https://github.com/MatrixAI/Polykey/pull/582

CMCDragonkai:
> > The identities authenticate right has a too low timeout, it should allow for a longer time.
> 
> Changed timeout to 1000000 in [MatrixAI/Polykey#582](https://github.com/MatrixAI/Polykey/pull/582)

Just use `Infinity`.

tegefaulkes:
> The produced ELF doesn't work on Ubuntu atm.
> 
> ```
> vboxuser@Ubuntu-Test:~/Downloads$ ./result 
> /snapshot/polykey-cli/bundle/polykey.js:66268
>       throw new Error(`Failed requiring possible native bindings: ${prebuildTargets.concat(npmTargets)}`);
>       ^
> 
> Error: Failed requiring possible native bindings: /snapshot/prebuild/quic-linux-x64.node,@matrixai/quic-linux-x64
>     at requireBinding (/snapshot/polykey-cli/bundle/polykey.js:66268:13)
>     at node_modules/@matrixai/quic/dist/native/quiche.js (/snapshot/polykey-cli/bundle/polykey.js:66310:29)
>     at __require (/snapshot/polykey-cli/bundle/polykey.js:13:50)
>     at node_modules/@matrixai/quic/dist/native/index.js (/snapshot/polykey-cli/bundle/polykey.js:66437:20)
>     at __require (/snapshot/polykey-cli/bundle/polykey.js:13:50)
>     at node_modules/@matrixai/quic/dist/QUICSocket.js (/snapshot/polykey-cli/bundle/polykey.js:67240:20)
>     at __require (/snapshot/polykey-cli/bundle/polykey.js:13:50)
>     at node_modules/@matrixai/quic/dist/index.js (/snapshot/polykey-cli/bundle/polykey.js:70045:24)
>     at __require (/snapshot/polykey-cli/bundle/polykey.js:13:50)
>     at node_modules/polykey/dist/nodes/utils.js (/snapshot/polykey-cli/bundle/polykey.js:73986:18)
> 
> Node.js v18.5.0
> ```
> 
> I'm getting that.
> 
> Trying to update to the latest pkg release of 3.5... but it's kind of weird why it's not realising it.
> 
> ```
> diff --git a/utils.nix b/utils.nix
> index b3e08f8..044fbb3 100644
> --- a/utils.nix
> +++ b/utils.nix
> @@ -69,29 +69,29 @@ rec {
>      '';
>    });
>    pkgBuilds = {
> -    "3.4" = {
> +    "3.5" = {
>        "linux-x64" = fetchurl {
> -        url = "https://github.com/vercel/pkg-fetch/releases/download/v3.4/node-v18.5.0-linux-x64";
> -        sha256 = "0b7iimvh2gldvbqfjpx0qvzg8d59miv1ca03vwv6rb7c2bi5isi5";
> +        url = "https://github.com/vercel/pkg-fetch/releases/download/v3.5/node-v18.15.0-linux-x64";
> +        sha256 = "0glr88p9higdwsffg3l243kpixqcf1mb7fawq62rj9n7b275lwx4";
>        };
>        "win32-x64" = fetchurl {
> -        url = "https://github.com/vercel/pkg-fetch/releases/download/v3.4/node-v18.5.0-win-x64";
> -        sha256 = "0jxrxgcggpzzx54gaai24zfywhq6fr0nm75iihpn248hv13sdsg0";
> +        url = "https://github.com/vercel/pkg-fetch/releases/download/v3.5/node-v18.15.0-win-x64";
> +        sha256 = "1d51w02m5jv7fgk3brkv3wizn1l75rai1zyq8m9vlm1za1gaha8p";
>        };
>        "macos-x64" = fetchurl {
> -        url = "https://github.com/vercel/pkg-fetch/releases/download/v3.4/node-v18.5.0-macos-x64";
> -        sha256 = "0dg46fw3ik2wxmhymcj3ih0wx5789f2fhfq39m6c1m52kvssgib3";
> +        url = "https://github.com/vercel/pkg-fetch/releases/download/v3.5/node-v18.15.0-macos-x64";
> +        sha256 = "1qcih9l3vncg05glhr45avcz2p5sqk7sp9776q4133xg88s09k0k";
>        };
>        # No build for v18.15.0 macos-arm64 build
>        # "macos-arm64" = fetchurl {
> -      #   url = "https://github.com/vercel/pkg-fetch/releases/download/v3.4/node-v18.5.0-macos-arm64";
> +      #   url = "https://github.com/vercel/pkg-fetch/releases/download/v3.5/node-v18.15.0-macos-arm64";
>        #   sha256 = "1znxssrwcg8nxfr03x1dfz49qq70ik33nj42dxr566vanayifa94";
>        # };
>      };
>    };
>    pkgCachePath =
>      let
> -      pkgBuild = pkgBuilds."3.4";
> +      pkgBuild = pkgBuilds."3.5";
>        fetchedName = n: builtins.replaceStrings ["node"] ["fetched"] n;
>      in
>        linkFarm "pkg-cache"
> ```
> 
> The above is the diff I'm trying and it doesn't work, something about the `pkg` at 5.8.1 still wants to use the 3.4. In fact the `package-lock.json` shows this:
> 
> ```
> *     "node_modules/pkg": {
>         "version": "5.8.1",
> *       "resolved": "https://registry.npmjs.org/pkg/-/pkg-5.8.1.tgz",
>         "integrity": "sha512-CjBWtFStCfIiT4Bde9QpJy0KeH19jCfwZRJqHFDFXfhUklCx8JoFmMj3wgnEYIwGm>
>         "dev": true,
>         "dependencies": {
>           "@babel/generator": "7.18.2",
>           "@babel/parser": "7.18.4",
>           "@babel/types": "7.19.0",
>           "chalk": "^4.1.2",
>           "fs-extra": "^9.1.0",
>           "globby": "^11.1.0",
>           "into-stream": "^6.0.0",
>           "is-core-module": "2.9.0",
>           "minimist": "^1.2.6",
>           "multistream": "^4.1.0",
> *         "pkg-fetch": "3.4.2",
>           "prebuild-install": "7.1.1",
>           "resolve": "^1.22.0",
>           "stream-meter": "^1.0.4"
>         },
>         "bin": {
> *         "pkg": "lib-es5/bin.js"
>         },
>         "peerDependencies": {
>           "node-notifier": ">=9.0.1"
>         },
>         "peerDependenciesMeta": {
>           "node-notifier": {
>             "optional": true
>           }
>         }
>       },
> ```
> 
> The simple reason is that the latest release of vercel/pkg at https://github.com/vercel/pkg/releases/tag/5.8.1 still uses the old pkg-fetch at 3.4. So we cannot use the latest 3.5.
> 
> However I'm not sure if that is the reason for the failure on ubuntu.

`PKG` has trouble resolving dependencies for native modules since they're usually dynamically loaded. In this case you'll need to add all of the quic optional native dependencies as assests to the pkg config.

CMCDragonkai:
> Remaining issues in summary.
> 
> * [ ]  1. On testnet, we still have a hash password problem [Bundling and pkg fixes #26 (comment)](https://github.com/MatrixAI/Polykey-CLI/pull/26#issuecomment-1760971608) - so currently the testnet nodes are just failing to run.
> * [ ]  2. The built pkg executable doesn't work on Ubuntu [Bundling and pkg fixes #26 (comment)](https://github.com/MatrixAI/Polykey-CLI/pull/26#issuecomment-1760993275) and we aren't able to update to pkg-fetch 3.5, node version 18.15 due to no new release on vercel/pkg
> * [ ]  3. Repeated start/stop results in a task manager error: [Bundling and pkg fixes #26 (comment)](https://github.com/MatrixAI/Polykey-CLI/pull/26#issuecomment-1760765602)
> * [ ]  4. The esbuild minified output doesn't work.
> * [ ]  5. I think the RPC's usage of graceful stream closure needs to be revisited and also made more robust in case the other side doesn't gracefully close within a certain timeout.
> * [x]  6. The `PolykeyClient` is supposed to encapsulate `WebSocketClient`, so that needs to be done on `Polykey`.
> * [ ]  7. The identities list or discover is slow, but a query can be crafted to make this faster.
> * [ ]  8. Authenticate uses `github.com:abc`, claim is `github.com abc`, so really we should make this consistent.
> * [ ]  9. Cannot pull/clone a vault from a "gestalt", we have to do it directly from a node ID, that makes it a bit difficult. It would be ideal to "scan" all the nodes of a gestalt to find a particular vault name, however resolve ambiguities and then pull it. Remember different nodes might share different vault ids with the same name, so when pulling/cloning one has to choose which one to take, might require prompting the user.
> * [ ]  10. The remote certs chain is not properly serialised due being a uint8array.
> * [ ]  11. There are still some test failures in Polykey-CLI, that has to be addressed too.
> * [ ]  12. Finding its own node ID should work.
> * [ ]  13. When sharing a vault, it should be possible to share a vault to a gestalt ID, whether node or identity. And it should be possible to do this, without identities discovering the node ID.
> * [ ]  14. Cloning a vault should be able take a gestalt id.
> * [x]  15. The identities authenticate right has a too low timeout, it should allow for a longer time.
> 
> Anything else...?

This is now being tracked in https://github.com/MatrixAI/Polykey/issues/551#issuecomment-1763555795. Got rid of duplicate tasks.

CMCDragonkai:
This PR will focus on just getting the builds happening, which it is... so I'll investigate the pkg build problem by first seeing if we should use another pkg repo now that it has been forked.

And then there's just the tests currently failing in this branch and if these 2 are done, we should merge this, and all other issues should be separate PRs.

CMCDragonkai:
Changing to https://github.com/yao-pkg/pkg-fetch/releases.

Looks like node2nix needs updating, since it no longer understands version 3 of package lock. #21

tegefaulkes:
> Changing to https://github.com/yao-pkg/pkg-fetch/releases.
> 
> Looks like node2nix needs updating, since it no longer understands version 3 of package lock. #21

I don't think it ever supported version 3 of the package lock. It's possible to force the package lock version to 2 if you need to generate it from scratch.

CMCDragonkai:
Going to try `buildNpmPackage`. https://nixos.org/manual/nixpkgs/stable/#javascript-tool-specific

CMCDragonkai:
https://github.com/NixOS/nixpkgs/tree/fc96f6d0fdac6ac8e06af95c6ed416cb7c7d4e2e/pkgs/build-support/node/build-npm-package

So the way this appears to work is quite clean.

It makes a few setup hooks and integrates it into the standard derivation process.

So there are 3 primary phases:

1. Config phase - https://github.com/NixOS/nixpkgs/blob/fc96f6d0fdac6ac8e06af95c6ed416cb7c7d4e2e/pkgs/build-support/node/build-npm-package/hooks/npm-config-hook.sh
2. Build phase - https://github.com/NixOS/nixpkgs/blob/fc96f6d0fdac6ac8e06af95c6ed416cb7c7d4e2e/pkgs/build-support/node/build-npm-package/hooks/npm-build-hook.sh
3. Install phase - https://github.com/NixOS/nixpkgs/blob/fc96f6d0fdac6ac8e06af95c6ed416cb7c7d4e2e/pkgs/build-support/node/build-npm-package/hooks/npm-install-hook.sh

It appears the idea is that config phase will:

* patch stuff
* setup some relevant env variables like `HOME`, `npm_config_nodedir`, `npm_config_node_gyp`
* setup other configuration variables
* installing dependencies using the `npm ci` command, I believe this actually installs development dependencies
* patch shebangs
* npm rebuild
* patch shebangs

THe build phase will run:

* npm run build - this will basically do our prebuild, build, postbuild

The install phase:

* Takes the built stuff, and finally installs the package

So interestingly... if we don't integrate `esbuild` into our `build` script which we haven't this should actually work fine.

CMCDragonkai:
Looks like I need to move ahead with the nixpkgs hash due to this bug for `buildNpmPackage`: https://github.com/NixOS/nixpkgs/issues/223964

CMCDragonkai:
Ok so it turns out, the problem is actually https://github.com/NixOS/nixpkgs/issues/261137.

The solutions are:

1. Delete `package-lock.json` and `rm -r node_modules`. Then regenerate both.
2. Use https://github.com/jeslie0/npm-lockfile-fix to fix the lock file.

This is actually an upstream npm bug, see: https://github.com/jeslie0/npm-lockfile-fix/issues/1.

Very annoying. I'd like nixpkgs `buildNpmPackage` to fix this automatically. Atm `npm-lockfile-fix` is a nix flake, and I haven't studied it yet to how to integrate it into our development environment. And `npm-lockfile-fix` isn't packaged in nixpkgs.

So for now, we just have to fix up our `package-lock.json` prior to running `nix-build`...

CMCDragonkai:
The new `buildNpmPackage` is much better than `node2nix`, it pretty much does the right from the beginning, and gets rid of development dependencies too. I think I'll stick with it.

However more optimal results would be from getting esbuild working in it too. It's just we would still have to keep some of the `node_modules` because of the paths referring to certain external packages atm. Can't use `pkg` because the node executable it uses would not be compatible on NixOS due to the loader.

CMCDragonkai:
So confirming:

* it will unpack sources
* patch sources
* configuring
* building involves installing dependencies including dev dependencies
* shrinking rpaths of elf executables
* patching script interpreter paths
* this creates a `polykey-cli-0.0.1-npm-deps` nix store path which is basically the npm cache
* it now setups up the repo by running the config hook
* then it installs dependencies from the cache
* patches interpreter paths 
* finishes the npm config hook
* then runs the npm build hook
* `shx rm -rf ./dist && tsc -p ./tsconfig.build.json`
* finishes npm build hook
* runs the install now
* the installation produces a `result/` with `bin` and `lib` similar to the original node2nix, but now without dev dependencies, and automatically resolves the `bin` scripts
* the `bin` scripts is using make wrapper, with `node` executing the path to `polykey.js`

CMCDragonkai:
So this works now:

```sh
nix-build ./release.nix -A application --argstr npmDepsHash "$(prefetch-npm-deps ./package-lock.json)"
```

It is necessary to ensure that the lock file is properly generated, that means sometimes deleting the `node_modules` and `package-lock.json` and regenerating from scratch.

That above parameter is necessary.

It will then generate a `result/` that looks like a NPM package.

However it can be further optimised, by applying `esbuild` to it, I think it is matter of customising the build phase with a `postBuild` hook.

CMCDragonkai:
It's all working now. Nix builds is properly building bundled output and pkg as well has been switched over. No more node2nix.

CMCDragonkai:
Remaining thing is the failing test @tegefaulkes please have a look at it, but I think things should be building.

I'd be interested to see if the pkg output still fails on Ubuntu, but that'd be a separate issue to look at.

scripts/brew-install.sh
```diff
@@ -10,5 +10,5 @@ export HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK=1
 export HOMEBREW_NO_AUTO_UPDATE=1
 export HOMEBREW_NO_ANALYTICS=1
 
-brew install node@18
-brew link --overwrite node@18
+brew reinstall node@20
+brew link --overwrite node@20
```
CMCDragonkai, line 14: The only thing that actually requires node 18 is `pkg`. So everything should still work on node 20. And then we will eventually update to node 20. But since I just started using nodejs, maybe we just all swap to using node 18?

CMCDragonkai:
Task 3 will depend on fixing the test case.

If @addievo can get the encapsulation fully working, we could incorporate here, but I think this will probably merge first, then another PR to integrate that change.

felschr:
> **It is necessary to ensure that the lock file is properly generated, that means sometimes deleting the `node_modules` and `package-lock.json` and regenerating from scratch.** - this will need to go into README.md

npm issues that track the problem of missing `integrity` & `resolved` fields:
https://github.com/npm/cli/issues/4263, https://github.com/npm/cli/issues/4460 https://github.com/npm/cli/issues/6301

CMCDragonkai:
Docker build works fine right now.

```
[nix-shell:~/Projects/Polykey-CLI]$ docker run -it -v "tmp:/srv" polykey-cli-0.0.1:lp7f5j5wmcxjn6fvhry63sk4fr4jra06 agent start -np /srv
✔ Enter new password … ******
✔ Confirm new password … ******
pid	1
nodeId	"v7poq4ut3in64l9r81tqe4bicfire2quhhtl7ga6fg218gp7e1c80"
clientHost	"127.0.0.1"
clientPort	34707
agentHost	"::"
agentPort	39376
recoveryCode	"kitchen wool romance mirror buyer old welcome arctic pistol horse shrug hundred general gain wage where milk resist pepper absorb bean avoid foster bicycle"
^C
```

CMCDragonkai:
So the esbuilt bundle seems to behave differently in relation to signal handling - in particular cleanup.

```
[nix-shell:~/Projects/Polykey-CLI]$ node ./build/polykey.js agent start
✔ Please enter the password … ******
pid	1161525
nodeId	"vj4tcqd7fom79fo3q418g1616739260ngdibpteguvaid49kcakjg"
clientHost	"::1"
clientPort	36029
agentHost	"::"
agentPort	47935
```

```
[nix-shell:~/Projects/Polykey-CLI]$ dist/polykey.js agent start
✔ Please enter the password … ******
pid	1161652
nodeId	"vj4tcqd7fom79fo3q418g1616739260ngdibpteguvaid49kcakjg"
clientHost	"::1"
clientPort	37555
agentHost	"::"
agentPort	56440
```

The issue is that `Ctrl+C` is immediate for `dist/` while `Ctrl+C` takes some time for `build/`.

---

This may be related to the occasional exception I get when running polykey agent against an existing node state.

```
ERROR:polykey.PolykeyAgent.scheduler:Failed scheduling loop TypeError: The first argument must be of type string or an instance of Buffer, ArrayBuffer, or Array or an Array-like Object. Received an instance of Object
ErrorTaskManagerScheduler: TaskManager scheduling loop encountered an unrecoverable error
  cause: TypeError: The first argument must be of type string or an instance of Buffer, ArrayBuffer, or Array or an Array-like Object. Received an instance of Object
```

Also I had seen this earlier: https://github.com/MatrixAI/Polykey-CLI/pull/26#issuecomment-1760765602 but without the earlier `Failed scheduling loop` error message.

CMCDragonkai:
Docker seems to fail in VirtualBox, so don't use it there.

tegefaulkes:
Ok, so we have found two small issues with the CLI that I think are inter-related.

1. The esbuild output, when running `pk agent start` and then closing the program with `ctrl c` doesn't do a graceful shutdown of the `PolykeyAgen`. Where as the normal build output does this properly. I've Can see that the clean up is being called, It's just not waiting for it to do anything before the process exits.
2. When running the agent after shutting it down, we sometimes get an error coming from the `TaskManager` scheduler. I'll have a snippet of the error below.

```
pid     4167964
nodeId  "vh75uomlvc1fm710e1nepu71kfj1fs1lag99or0htq9n9f9gcsqpg"
clientHost      "::1"
clientPort      40871
agentHost       "::"
agentPort       40755
TypeError [ERR_INVALID_ARG_TYPE]: The first argument must be of type string or an instance of Buffer, ArrayBuffer, or Array or an Array-like Object. Received an instance of Object
    at new NodeError (node:internal/errors:405:5)
    at Function.from (node:buffer:333:9)
    at decrypt (/home/faulkes/matrixcode/polykey/Polykey-CLI/dist/polykeyWorker.js:34861:59)
    at /home/faulkes/matrixcode/polykey/Polykey-CLI/dist/polykeyWorker.js:478:24
    at Generator.next (<anonymous>)
    at /home/faulkes/matrixcode/polykey/Polykey-CLI/dist/polykeyWorker.js:384:71
    at new Promise (<anonymous>)
    at __awaiter3 (/home/faulkes/matrixcode/polykey/Polykey-CLI/dist/polykeyWorker.js:366:14)
    at runFunction (/home/faulkes/matrixcode/polykey/Polykey-CLI/dist/polykeyWorker.js:475:14)
    at /home/faulkes/matrixcode/polykey/Polykey-CLI/dist/polykeyWorker.js:521:13
ERROR:polykey.PolykeyAgent.scheduler:Failed scheduling loop TypeError: The first argument must be of type string or an instance of Buffer, ArrayBuffer, or Array or an Array-like Object. Received an instance of Object
ErrorTaskManagerScheduler: TaskManager scheduling loop encountered an unrecoverable error
  cause: TypeError: The first argument must be of type string or an instance of Buffer, ArrayBuffer, or Array or an Array-like Object. Received an instance of Object
```

I'm looking into problem 1. right now. I think 2. is caused by 1.

CMCDragonkai:
Ok this is updated with alpha.13 which uses `@matrixai/quic` at 1.0.0. And the docker works.

However still need to fix the reported problems @tegefaulkes is on and the existing test problems here.

Now I can try to get the CI ready.

tegefaulkes:
progress update.

Problem 2 with the taskmanager. The error is coming out of the worker script. The data seems to be formatted badly in the esbuild build.

worker function being called.
```js
decrypt(key, cipherText) {
    console.log('key', key);
    console.log('cipherText', cipherText);
    const plainText = keysUtils.decryptWithKey(Buffer.from(key), Buffer.from(cipherText));
    if (plainText != null) {
        return (0, worker_1.Transfer)(plainText.buffer);
    }
    else {
        return;
    }
}
```

normal build data
```
key ArrayBuffer {
  [Uint8Contents]: <02 50 56 20 16 0f 49 2f ed c2 29 c5 d6 5a 0b 5f 7d e5 0b b5 37 5f 20 47 b9 4e d4 77 52 f1 a2 e0>,
  byteLength: 32
}
cipherText ArrayBuffer {
  [Uint8Contents]: <e5 24 f0 37 b5 37 90 9e 86 3e d6 37 f1 02 1c 01 bb 5d f5 eb bf fe 58 60 04 94 39 88 07 c9 9f b3 be af 48 6c 1b 2e 1b ac 4b 95 2b 37 64 87 2e ae a3 91 f8 ef 1b f5 c9 3c 02 73 b4 f6 73 48 1d 8a 3c bf e4 69 28 dc 78 bb ec 2a 30 e7 dc 2d 40 47 35 00 ac 77 ae ce 7e ec 6e 66 6e c7 41 64 a4 79 b5 17 c7 6f ... 198 more bytes>,
  byteLength: 298
}

```

esbuild data
```
key {
  send: ArrayBuffer {
    [Uint8Contents]: <02 50 56 20 16 0f 49 2f ed c2 29 c5 d6 5a 0b 5f 7d e5 0b b5 37 5f 20 47 b9 4e d4 77 52 f1 a2 e0>,
    byteLength: 32
  },
  transferables: [
    ArrayBuffer {
      [Uint8Contents]: <02 50 56 20 16 0f 49 2f ed c2 29 c5 d6 5a 0b 5f 7d e5 0b b5 37 5f 20 47 b9 4e d4 77 52 f1 a2 e0>,
      byteLength: 32
    }
  ]
}
cipherText {
  send: ArrayBuffer {
    [Uint8Contents]: <63 9b 45 61 5d 83 bc b9 49 e3 16 c6 5b 79 a0 64 be b6 16 ce cb bd 83 a6 26 cc ae 2d 8c 83 c6 63 a0 7b 12 68 bd 8c 16 95 b3 b8 5c 2d bb cf ec e7 dd 6c 8f a8 44 c5 57 62 56 53 90 06 03 59 f6 8e 32 d9 c9 0e 0a 7b 4b fa f6 d3 35 57 16 f5 34 54 fe 45 20 83 82 bc d3 ca 27 a3 97 2f 67 f4 47 a0 37 e5 a8 23 ... 204 more bytes>,
    byteLength: 304
  },
  transferables: [
    ArrayBuffer {
      [Uint8Contents]: <63 9b 45 61 5d 83 bc b9 49 e3 16 c6 5b 79 a0 64 be b6 16 ce cb bd 83 a6 26 cc ae 2d 8c 83 c6 63 a0 7b 12 68 bd 8c 16 95 b3 b8 5c 2d bb cf ec e7 dd 6c 8f a8 44 c5 57 62 56 53 90 06 03 59 f6 8e 32 d9 c9 0e 0a 7b 4b fa f6 d3 35 57 16 f5 34 54 fe 45 20 83 82 bc d3 ca 27 a3 97 2f 67 f4 47 a0 37 e5 a8 23 ... 204 more bytes>,
      byteLength: 304
    }
  ]
}
```

CMCDragonkai:
So the `threads` package doesn't work nicely with esbuild output. We just need to make sure to keep it external for now, however we should be replacing it later during ESM migration and it would be important keep in mind this problem.

CMCDragonkai:
Ok all tests is passing now.

But there's a weird exit gracefully problem:

```
A worker process has failed to exit gracefully and has been force exited. This is likely caused by tests leaking due to improper teardown. Try running with --detectOpenHandles to find leaks. Active timers can also cause this, ensure that .unref() was called on them.
Test Suites: 38 passed, 38 total
Tests:       21 skipped, 1 todo, 120 passed, 142 total
Snapshots:   0 total
Time:        56.186 s
Ran all test suites.
GLOBAL TEARDOWN
Destroying Global Data Dir: /tmp/polykey-test-global-kyq3VE
```

Possibly related to https://github.com/MatrixAI/js-timer/issues/15, not a problem atm.

CMCDragonkai:
@tegefaulkes can you start squashing this so it can be merged. We can merge in the replacement of websocketclient afterwards.

Also make sure you're squashing any of my wip commits.

CMCDragonkai:
Ok this is squashed, merging to staging to confirm the final CI steps is working which should also be auto-building and pushing to the ECR.

CMCDragonkai:
However, still pending major issue is https://github.com/MatrixAI/Polykey/pull/582. That's next priority for PK CLI.

scripts/brew-install.sh
CMCDragonkai: I've set it to node 18 for now. That's specific to PK CLI due to pkg not supporting node 20 any time soon.

PR #29, opened 2023-10-18
Fixing identities claim to be in-line with identities discover and identities trust, by introducing a new parser parseGestaltIdentityId

### LINKED ISSUE #28: Discrepancy in claim, discover and trust commands in CLI
### Specification

Currently the claim command in Polykey-CLI uses the following syntax 

```typescript
npm run polykey -- identities claim github.com GITHUBUSERNAME
```

Whereas the discover and trust commands use 

```typescript
npm run polykey -- identities discover github.com:ANOTHERGITHUBUSERNAME
npm run polykey -- identities trust github.com:ANOTHERGITHUBUSERNAME
``` 

The claim command is out of sync of trust and discover, as they both utilize a COLON to separate provider and identity, whereas the claim command uses a space, this discrepancy is a bug.

### Tasks

1. `src/identities/CommandClaim.ts` to alter the input args to be in-line with trust and discover.
2. ...
3. ...

After this change, all the identity related functions of polykey would be syntactically identical.

addievo:
### Description

The `pk identities claim` currently takes two parameters in order to claim an identity. 

```
pk identities claim github.com cmcdragonkai
```

This is not consistent with other two commands `trust` and `discover`. 

```
pk identities trust github.com:cmcdragonkai
pk identities discover github.com:cmcdragonkai
```

These two commands are not using the `ProviderIdentityId` type. 

```
type ProviderIdentityId = [ProviderId, IdentityId];
```

This is because the encoded form of this type is `["github.com", "cmcdragonkai"]`.

Instead we need to use a parser similar to `binParsers.parseGestaltId`  (which is used by the trust and discover commands), however this parser, parses both node IDs and identities. 

What we need to do is to extract the identity parsing portion of `binParsers.parseGestaltId`.

```typescript
function parseGestaltId(data: any): GestaltId {
  if (typeof data !== 'string') {
    throw new validationErrors.ErrorParse('Gestalt ID must be string');
  }
  const nodeId = decodeNodeId(data);
  if (nodeId != null) {
    return ['node', nodeId];
  }
  const match = (data as string).match(/^(.+):(.+)$/);
  if (match == null) {
    throw new validationErrors.ErrorParse(
      'Gestalt ID must be either a Node ID or `Provider ID:Identity ID`',
    );
  }
  const providerId = parseProviderId(match[1]);
  const identityId = parseIdentityId(match[2]);
  return ['identity', [providerId, identityId]];
}
```

It will be:

```ts
  const match = (data as string).match(/^(.+):(.+)$/);
  if (match == null) {
    throw new validationErrors.ErrorParse(
      'Gestalt ID must be either a Node ID or `Provider ID:Identity ID`',
    );
  }
  const providerId = parseProviderId(match[1]);
  const identityId = parseIdentityId(match[2]);
  return ['identity', [providerId, identityId]];
```

This will be called `parseGestaltIdentityId`.

### Issues Fixed

* Fixes #28 

### Tasks

- [x] 1. Alter handler call.
- [x] 2. Alter args.

### Final checklist

* [x] Domain specific tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/src/identities/CommandClaim.ts
+++ b/src/identities/CommandClaim.ts
@@ -1,4 +1,5 @@
 import type PolykeyClient from 'polykey/dist/PolykeyClient';
+import type { ProviderIdentityId } from 'polykey/dist/ids';
 import CommandPolykey from '../CommandPolykey';
 import * as binUtils from '../utils';
 import * as binOptions from '../utils/options';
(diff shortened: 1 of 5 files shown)


src/identities/CommandClaim.ts
```diff
@@ -1,5 +1,6 @@
 import type PolykeyClient from 'polykey/dist/PolykeyClient';
 import type { WebSocketClient } from '@matrixai/ws';
+import type { GestaltId } from 'polykey/dist/gestalts/types';
```
tegefaulkes, line 3: You can import types from the `polykey` directly now. Us 
```suggestion
import type { GestaltId } from 'polykey';
```

src/identities/CommandClaim.ts
```diff
@@ -12,19 +13,14 @@ class CommandClaim extends CommandPolykey {
...
-      'Name of the digital identity provider',
-      binParsers.parseProviderId,
-    );
-    this.argument(
-      '<identityId>',
-      'Digital identity to claim',
-      binParsers.parseIdentityId,
+      '<gestaltId>',
+      '`Provider ID:Identity ID`',
+      binParsers.parseGestaltId,
```
tegefaulkes, line 18: This command is for claiming an identity right? We Only want a IdentityId to be valid. `parseGestaltId` can parse `NodeId`s as well, so  we don't want that here. 

I'm pretty sure with this the command will take a `NodeId` but break.

src/identities/CommandClaim.ts
```diff
@@ -12,19 +13,14 @@ class CommandClaim extends CommandPolykey {
...
     );
     this.addOption(binOptions.nodeId);
     this.addOption(binOptions.clientHost);
     this.addOption(binOptions.clientPort);
-    this.action(async (providerId, identityId, options) => {
+    this.action(async (gestaltId: GestaltId, options) => {
```
tegefaulkes, line 23: Like the above comment, we're after an `IdentityId` only, not a `GestaltId`.

src/identities/CommandClaim.ts
```diff
@@ -70,25 +66,30 @@ class CommandClaim extends CommandPolykey {
...
+            if (claimMessage.url) {
+              output.push(`Url: ${claimMessage.url}`);
+            }
+            process.stdout.write(
+              binUtils.outputFormatter({
+                type: options.format === 'json' ? 'json' : 'list',
+                data: output,
+              }),
+            );
+          }
```
tegefaulkes, line 91: This is really bad, I understand that you want to enforce the type for an `IdentityId`. BUT a switch statement with only 1 case is really bad way to do this.

1. We shouldn't be using the `GestaltId` here, we should only be working with an `IdentityId` here.
2. You cant use a single case for a switch statement like this. It's basically saying only handle the `identity` type and do nothing in all other cases. If someone provides a `NodeId` this block will do nothing, If for some reason anything else was provided (it can happen, thats a whole rant about typescript to go into) it will do nothing. No explination, just nothing happens. That is not a fun bug to track down.
3. If you ever use a switch statement. ALWAYS include a default case that calls `utils.never()` even if all cases are covered. This is a fail safe.


Ultimately you shouldn't *need* a switch statement here, you shouldn't be using the `parseGestaltId`.

Also, did this not feel wrong at all when writing it? This is a pretty bad code smell.

tegefaulkes (changes requested):
The main issue is that you're using the `GestaltId` parser when you should only ever be parsing a `IdentityId`. If there is no existing parser for `IdentityId` for the format of `provider:user` then you need to make one.

Using the switch statement like that to enforce the type is a pretty weird decision. It's almost as bad as using the `as` keyword to narrow the type. 

Remember, a quick solution that doesn't address the issue is not a good one. You need take some extra time to think through your implementation and weather it's a good one. 

If you're unsure about something of can think of multiple ways to address something the you can @ someone in a the issue or PR with a question.  it also helps to think out loud in the issues or PRs when working.

tegefaulkes:
Remember, don't resolve review comments until you have committed and pushed the change. It's really confusing on my end if its resolved with no changes.

addievo:
### Important

This PR needs https://github.com/MatrixAI/Polykey/pull/595 to be merged in Polykey to work.

src/identities/CommandClaim.ts
```diff
@@ -13,17 +13,17 @@ class CommandClaim extends CommandPolykey {
     this.name('claim');
     this.description('Claim a Digital Identity for this Keynode');
     this.argument(
-      '<gestaltId>',
-      '`Provider ID:Identity ID`',
-      binParsers.parseGestaltId,
+      '<providerIdentityId>',
+      'Name of the digital identity provider',
+      binParsers.parseGestaltIdentityId,
```
tegefaulkes, line 18: You still want to be clear how it's formatted here.

CMCDragonkai: Line 17 use the string that trust/discover uses.

src/utils/parsers.ts
```diff
@@ -86,6 +86,9 @@ const parseNodeId: (data: string) => ids.NodeId = validateParserToArgParser(
 );
 const parseGestaltId: (data: string) => ids.GestaltId =
   validateParserToArgParser(ids.parseGestaltId);
+
```
CMCDragonkai, line 89: No whitespace, but you can have another commit that creates whitespace between all the definitions.

addievo: Resolved

addievo:
@tegefaulkes this PR was finished reviewing by Roger and you. Think you can merge it

tegefaulkes:
This is not techincally finished if it's waiting for a new version of Polykey to be released. All tests need to be passing as well.

tests/identities/claim.test.ts
tegefaulkes, line 1: I don't think you updated all the tests here.

tegefaulkes: the other two tests shouldn't be doing this.

```
({ exitCode } = await testUtils.pkStdio(
        ['identities', 'claim', testToken.providerId, ''],
        {
          env: {
            PK_NODE_PATH: nodePath,
            PK_PASSWORD: password,
          },
          cwd: dataDir,
        },
      ));
      expect(exitCode).toBe(sysexits.USAGE);
```

with `['identities', 'claim', testToken.providerId, '']` as the arguments. It should be `['identities', 'claim', `${testToken.providerId}:`]`. We're not testing if commander can count, we're testing if we can handle invalid input in the parser.

CMCDragonkai:
The `parseGestaltIdentityId` is a "fragment" of `parseGestaltId` so it should be returning `['identity', ProviderId, IdentityId`, that's correct.

tegefaulkes:
I've reverted that change now. I'll finish off this PR at @CMCDragonkai 's request

tests/identities/claim.test.ts
tegefaulkes: I've fixed this now.

tegefaulkes (approved):

tegefaulkes:
This is done now, I'm merging.

PR #39, opened 2023-10-19
Unifies the `polykey.ts` and `polykey-agent.ts` to make forking to background portable to bundling

### LINKED ISSUE #34: Starting agent in background doesn't work with esbuilt bundle
### Specification

The reason for this is quite simple. To run the agent in the background it relies on forking `polykey-agent.js`.

It looks like this:

```ts
        const agentProcess = childProcess.fork(
          path.join(__dirname, '../polykey-agent'),
          [],
          {
            cwd: process.cwd(),
            env: process.env,
            detached: true,
            serialization: 'advanced',
            stdio,
          },
        );
```

That's because the file `CommandStart` exists in `src/agent` directory. So it's trying to refer the file relative to the location of the script file.

However this no longer makes sense if the code is to be bundled together. Because then the location of the source script doesn't matter. What matters is the location of the bundled file.

This is a similar problem to the location of node native addons https://github.com/MatrixAI/js-quic/issues/71. And it's also similar to the problem of loading `polykeyWorker.js` which is now re-exported as an additional entrypoint under `dist/`.

There 2 ways we could solve this.

1. Combine `polykey.ts` and `polykey-agent.ts` together and use conditional logic to separate the different behaviour.
2. Maintain the location of the primary/main script and pass that down to `CommandStart`, so it always uses the path relative to the main script. This allows us to assume that `polykey-agent` will always relative to the `polykey` script (right next to it). And then this could work.

Solution 2 would be the quickest, however I don't know if this will work with a `pkg` bundled executable. Does the VFS it creates also intercepts forking calls to the local filesystem? It seems like it does... because, that's what happened with the workers.

```ts
  return await WorkerManager.createWorkerManager<PolykeyWorkerModule>({
    workerFactory: () => spawn(new Worker('./polykeyWorker')),
    cores,
    logger,
  });
```

If that's always relative to the script executing, then that's why we needed `dist/polykeyWorker.js`, and thus setting it to the `scripts` property in the pkg configuration.

### Additional context

* https://github.com/MatrixAI/js-quic/issues/71
* https://github.com/MatrixAI/js-workers/issues/16 - threads can be made more robust by always embedding the script

### Tasks

1. ...
4. ...
5. ...

CMCDragonkai:
### Description

This unifies the `polykey.ts` and `polykey-agent.ts` so that way forking to the background is portable.

### Issues Fixed

* Fixes #34

### Tasks

- [x] 1. Switching between normal polykey and agent mode with `--agent-mode` internal flag.
- [x] 2. The `globalThis.PK_MAIN_EXECUTABLE` is a special property set to always the file path of the main executable. This should also work for pkg, but needs to be tested on the CI with Ubuntu, right now set to `__filename`, but it can be changed to `import.meta.url` in ESM.

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased


CMCDragonkai:
One of the things I observed is that error reporting from `agent start --background` isn't good atm, it does not tell you why it failed. Like that there was an error or something. We need to improve the error reporting.

CMCDragonkai:
This works for the esbuilt bundle:

```
[nix-shell:~/Projects/Polykey-CLI]$ ./dist/polykey.js agent start --background
✔ Please enter the password … ******
pid	269435
nodeId	"vj4tcqd7fom79fo3q418g1616739260ngdibpteguvaid49kcakjg"
clientHost	"::1"
clientPort	46311
agentHost	"::"
agentPort	42836
```

Need to test on the CI for pkg and ubuntu.

Issue #22, opened 2022-02-17
Standardise STDOUT output for CLI commands


emmacasolin:
### Specification
Our new commandments for a better, brighter, command line:

* Send output to stdout. The primary output for your command should go to stdout. Anything that is machine readable should also go to stdout—this is where piping sends things by default.
* Send messaging to stderr. Log messages, errors, and so on should all be sent to stderr. This means that when commands are piped together, these messages are displayed to the user and not fed into the next command.
* When we are displaying output text to indicate success, we need to be brief. Feedback is always preferred, even though deviating from UNIX conventions.
* The output from stderr can be redirected to `/dev/null`, but we're going to provide more verbosity options that suppresses all non-essential text (feedback, warnings, etc.).
* We should ALWAYS be using `camelCase` for anything going to `stdout`. As we prioritize parsability, this needs to be the default. No more manually outputting strings like `Name: amy`. 
* Nested structures should always be avoided in `dicts` formats wherever possible. Instead, the nested structure should be spread into the `data`.
* Streaming with the JSON format should output the objects in JSONL format. To parse it with `jq`, one needs to use `--slurp`

#### STDOUT

Traditionally, we have constructed human readable dictionaries as such:

```ts
process.stdout.write(
  binUtils.outputFormatter({
    type: options.format === 'json' ? 'json' : 'list',
    data: [`Root certificate:\t\t${response.getCert()}`],
  }),
);
```

This is clunky, inconsistent, worse for parsability and more...

In all cases where the output is "semantically" a list, the output should use either `json` or `list`. In all other cases, the `list` option should be replaced with `dict`. This should be done like this:

```ts
process.stdout.write(
  binUtils.outputFormatter({
    type: options.format === 'json' ? 'json' : 'dict',
    data: { rootCertificate: response.getCert() },
  }),
);
```

The `data` parameter must be kept as close to the return value of the RPC call possible. This ensures that our output is predictable. Furthermore

#### STDERR

Log messages, errors, and so on should all be sent to stderr. This means that when commands are piped together, these messages are displayed to the user and not fed into the next command.

Most STDERR messages should either be using the `raw` or `list` format. This is because they are intended to be human readable. Under `json` format, the data should always be `{ message: "..." }`.

```ts
process.stderr.write(
  binUtils.outputFormatter({
    type: 'list',
    data: ['Message 1'],
  }),
);
```

```ts
process.stderr.write(
  binUtils.outputFormatter({
    type: 'raw',
[...]

PR #45, opened 2023-10-25
Feature cli polish

### LINKED ISSUE #44: CLI polish and fixes for deployment
### Specification

While testing hole punching with the testnet, we found a few issues with the CLI.

1. Nodes Get All is throwing an error
2. We are getting a weird output format for nodes connections what is the 0 and what is this NA for?
3. I think nodes connections is better suited for dictionary format - list format or table format (can you check which ones done robustly)
4. The output changes sometimes, to just join the NA after the ip address, so no port is available
5. Summary - the nodes connections command is just very non-deterministic
6. Secret leak is occurring on agent service, there should be no authorisation metadata see --format=json
8. What's usageCount

### Additional context
 
* Related: https://github.com/MatrixAI/Polykey/issues/551

### Tasks

- [x] 1. Fix up `nodes getall` command, it's throwing an error relating to `host` not being defined.
- [x] 2. Review and fix up the `nodes connections` command output formatting. The output is very unclear.
- [x] 3. client call metadata is being included in the `nodes connections --format json` output. This should not be the case.
- [x] 4. `agent status` command outputs too much information. We should strip out the key and certificate information.

addievo:
### Description

This PR aims to fix a lot of UI/UX issues of Polykey CLI to improve clarity and security of command outputs.

The main issues to be fixed include: 

1. Fixes `nodes getall`:

This previous threw an error related to undefined host, this has now been resolved in the corresonding Polykey Handler by getting the valid address and port from the `info` field which is of type `NodeData`
```ts
          host: info.address.host,
          port: info.address.port,
```

2. Improve `nodes connections` Output: 

Previously, the output was rather unclear and lacked corresponding labels to relevant labels for `Node Host`, `UsageCount` and `Timeout`, now labels are added for the same.

Previously we used a hardcoded amount of tabbed spaces between columns, this has now been changed to be padded based on length of row with the max length.

```ts
// Old Code
const outputLine = `${hostString}\t${usageCount}\t${timeout}`;

// New Code
const outputLine = hostString.padEnd(maxHostStringLength) + "\t" + usageCount.padEnd(maxUsageCountLength) + "\t" + timeout.padEnd(maxTimeoutLength);
```

3. Sanitizing `JSON` formatted output for `nodes connections`:

This was a critical security issue where in the sensitive information of authentication data (although this is the user's authentication data itself) was being included in the JSON output, this has been sanitized out.

```ts
// Implementation Snippet
const sanitizedConnections = JSON.parse(JSON.stringify(connections));
delete connection.metadata;
```

4. Refactor `agent status` output:

JWK and PEM have very little usage in status, and correspondingly are being removed from agent status, this is a very simple fix of just removing the following keys from the output

```ts
// Old Code
data: { ... , publicKeyJWK: response.publicKeyJwk, certChainPEM: response.certChainPEM }

// New Code
data: { ... , /* publicKeyJWK and certChainPEM removed */ }
```

5. Additional scope added to this PR - Standardisation of table layout for `outputFormatter` : 

Table output in `outputFormatter` was intiially very static and did not take into account variety in terms of padding, this is altered now to allow field for headers, and a boolean to toggle numbering of rows. 

`Table` format now includes a standardised TSV padding.

`Table` format handles `\t` and `\n` by enclosing them in quotation marks to prevent disruption of the layout.

Additional supported is added for stream based padding to improve efficiency to `O(2n)`.

### Issues Fixed
* Related #44 

### Tasks
- [x] 1. `nodes connection` human format label.
- [x] 2. `nodes connection` json format remove metadata (sanitize).
- [x] 3. Remove `JWK` and `PEM` from `agent status`.
- [x] 4. Add a jest for `nodes connection`.
- [x] 5. Update jests which rely on `agent status`.

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/src/agent/CommandStatus.ts
+++ b/src/agent/CommandStatus.ts
@@ -79,8 +79,6 @@ class CommandStatus extends CommandPolykey {
               clientPort: response.clientPort,
               agentHost: response.agentHost,
               agentPort: response.agentPort,
-              publicKeyJWK: response.publicKeyJwk,
-              certChainPEM: response.certChainPEM,
             },
           }),
         );
--- a/src/nodes/CommandConnections.ts
+++ b/src/nodes/CommandConnections.ts
@@ -48,35 +48,41 @@ class CommandAdd extends CommandPolykey {
             });
           const connectionEntries: Array<NodeConnectionMessage> = [];
           for await (const connection of connections) {
-            connectionEntries.push(connection);
+            connectionEntries.push({
+              host: connection.host,
+              hostname: connection.hostname,
+              nodeIdEncoded: connection.nodeIdEncoded,
+              port: connection.port,
+              timeout: connection.timeout,
+              usageCount: connection.usageCount,
+            });
           }
           return connectionEntries;
         }, auth);
         if (options.format === 'human') {
-          const output: Array<string> = [];
-          for (const connection of connections) {
-            const hostnameString =
-              connection.hostname === '' ? '' : `(${connection.hostname})`;
-            const hostString = `${connection.nodeIdEncoded}@${connection.host}${hostnameString}:${connection.port}`;
-            const usageCount = connection.usageCount;
-            const timeout =
-              connection.timeout === -1 ? 'NA' : `${connection.timeout}`;
-            const outputLine = `${hostString}\t${usageCount}\t${timeout}`;
-            output.push(outputLine);
-          }
-          process.stdout.write(
-            binUtils.outputFormatter({
-              type: 'list',
-              data: output,
-            }),
-          );
+          // Wait for outputFormatter to complete and then write to stdout
+          const formattedOutput = await binUtils.outputFormatter({
+            type: 'table',
+            data: connections,
+            options: {
+              headers: [
+                'host',
+                'hostname',
+                'nodeIdEncoded',
+                'port',
+                'timeout',
+                'usageCount',
+              ],
+            },
+          });
+          process.stdout.write(formattedOutput);
         } else {
-          process.stdout.write(
-            binUtils.outputFormatter({
-              type: 'json',
-              data: connections,
-            }),
-          );
+          // Wait for outputFormatter to complete and then write to stdout
+          const formattedOutput = await binUtils.outputFormatter({
+            type: 'json',
+            data: connections,
+          });
+          process.stdout.write(formattedOutput);
         }
       } finally {
         if (pkClient! != null) await pkClient.stop();
--- a/src/utils/utils.ts
+++ b/src/utils/utils.ts
@@ -1,4 +1,5 @@
 import type { POJO } from 'polykey/dist/types';
+import type { TableRow, TableOptions } from '../types';
 import process from 'process';
 import { LogLevel } from '@matrixai/logger';
 import ErrorPolykey from 'polykey/dist/ErrorPolykey';
@@ -34,6 +35,7 @@ type OutputObject =
   | {
       type: 'table';
       data: Array<POJO>;
+      options?: TableOptions;
     }
   | {
       type: 'dict';
@@ -62,7 +64,93 @@ function standardErrorReplacer(key: string, value: any) {
   return value;
 }
 
-function outputFormatter(msg: OutputObject): string | Uint8Array {
+async function* arrayToAsyncIterable<T>(array: T[]): AsyncIterable<T> {
+  for (const item of array) {
+    yield item;
+  }
+}
+
+// Function to handle 'table' type output
+const outputTableFormatter = async (
+  rowStream: TableRow[] | AsyncIterable<TableRow>,
+  options?: TableOptions,
+): Promise<string> => {
+  let output = '';
+  const maxColumnLengths: Record<string, number> = {};
+  let rowCount = 0;
+
+  // Initialize maxColumnLengths with header lengths if headers are provided
+  if (options?.headers) {
+    for (const header of options.headers) {
+      maxColumnLengths[header] = header.length;
+    }
+  }
+
+  let iterableStream = Array.isArray(rowStream)
+    ? arrayToAsyncIterable(rowStream)
+    : rowStream;
+
+  const updateMaxColumnLengths = (row: TableRow) => {
+    for (const [key, value] of Object.entries(row)) {
+      const cellValue =
+        value === null || value === '' || value === undefined ? 'N/A' : value;
+      maxColumnLengths[key] = Math.max(
+        maxColumnLengths[key] || 0,
+        cellValue.toString().length,
+      );
+    }
+  };
+
+  // Precompute max column lengths by iterating over the rows first
+  for await (const row of iterableStream) {
+    updateMaxColumnLengths(row);
+  }
+
+  // Reset the iterableStream if it's an array so we can iterate over it again
+  if (Array.isArray(rowStream)) {
+    iterableStream = arrayToAsyncIterable(rowStream);
+  }
+
+  // If headers are provided, add them to your output first
+  if (options?.headers) {
+    const headerRow = options.headers
+      .map((header) => header.padEnd(maxColumnLengths[header]))
+      .join('\t');
+    output += headerRow + '\n';
+  }
+
+  // Function to format a single row
+  const formatRow = (row: TableRow) => {
+    let formattedRow = '';
+
+    if (options?.includeRowCount) {
+      formattedRow += `${++rowCount}\t`;
+    }
+
+    const keysToUse = options?.headers ?? Object.keys(maxColumnLengths);
+
+    for (const key of keysToUse) {
+      const cellValue = Object.prototype.hasOwnProperty.call(row, key)
+        ? row[key]
+        : 'N/A';
+      formattedRow += `${cellValue
+        ?.toString()
+        .padEnd(maxColumnLengths[key] || 0)}\t`;
+    }
+
+    return formattedRow.trimEnd();
+  };
+
+  for await (const row of iterableStream) {
+    output += formatRow(row) + '\n';
+  }
+
+  return output;
+};
+
+function outputFormatter(
+  msg: OutputObject,
+): string | Uint8Array | Promise<string> {
   let output = '';
   if (msg.type === 'raw') {
     return msg.data;
@@ -75,40 +163,34 @@ function outputFormatter(msg: OutputObject): string | Uint8Array {
       output += `${msg.data[elem]}\n`;
     }
   } else if (msg.type === 'table') {
-    for (const key in msg.data[0]) {
-      output += `${key}\t`;
-    }
-    output = output.substring(0, output.length - 1) + `\n`;
-    for (const row of msg.data) {
-      for (const key in row) {
-        let value = row[key];
-        // Empty string for null or undefined values
-        if (value == null) {
-          value = '';
-        }
-        value = value.toString();
-        // Remove the last line terminator if it exists
-        // This may exist if the value is multi-line string
-        value = value.replace(/(?:\r\n|\n)$/, '');
-        output += `${value}\t`;
-      }
-      output = output.substring(0, output.length - 1) + `\n`;
-    }
+    return outputTableFormatter(msg.data, msg.options);
   } else if (msg.type === 'dict') {
+    let maxKeyLength = 0;
+    for (const key in msg.data) {
+      if (key.length > maxKeyLength) {
+        maxKeyLength = key.length;
+      }
+    }
+
     for (const key in msg.data) {
       let value = msg.data[key];
-      // Empty string for null or undefined values
       if (value == null) {
         value = '';
       }
-      value = JSON.stringify(value);
-      // Remove the last line terminator if it exists
-      // This may exist if the value is multi-line string
+
+      // Only trim starting and ending quotes if value is a string
+      if (typeof value === 'string') {
+        value = JSON.stringify(value).replace(/^"|"$/g, '');
+      } else {
+        value = JSON.stringify(value);
+      }
+
+      // Re-introduce value.replace logic from old code
       value = value.replace(/(?:\r\n|\n)$/, '');
-      // If the string has line terminators internally
-      // Then we must append `\t` separator after each line terminator
       value = value.replace(/(\r\n|\n)/g, '$1\t');
-      output += `${key}\t${value}\n`;
+
+      const padding = ' '.repeat(maxKeyLength - key.length);
+      output += `${key}${padding}\t${value}\n`;
     }
   } else if (msg.type === 'json') {
     output = JSON.stringify(msg.data, standardErrorReplacer);
--- a/src/nodes/CommandClaim.ts
+++ b/src/nodes/CommandClaim.ts
@@ -65,27 +65,25 @@ class CommandClaim extends CommandPolykey {
         );
         const claimed = response.success;
         if (claimed) {
-          process.stdout.write(
-            binUtils.outputFormatter({
-              type: options.format === 'json' ? 'json' : 'list',
-              data: [
-                `Successfully generated a cryptolink claim on Keynode with ID ${nodesUtils.encodeNodeId(
-                  nodeId,
-                )}`,
-              ],
-            }),
-          );
+          const formattedOutput = await binUtils.outputFormatter({
+            type: options.format === 'json' ? 'json' : 'list',
+            data: [
+              `Successfully generated a cryptolink claim on Keynode with ID ${nodesUtils.encodeNodeId(
+                nodeId,
+              )}`,
+            ],
+          });
+          process.stdout.write(formattedOutput);
         } else {
-          process.stdout.write(
-            binUtils.outputFormatter({
-              type: options.format === 'json' ? 'json' : 'list',
-              data: [
-                `Successfully sent Gestalt Invite notification to Keynode with ID ${nodesUtils.encodeNodeId(
-                  nodeId,
-                )}`,
-              ],
-            }),
-          );
+          const formattedOutput = await binUtils.outputFormatter({
+            type: options.format === 'json' ? 'json' : 'list',
+            data: [
+              `Successfully sent Gestalt Invite notification to Keynode with ID ${nodesUtils.encodeNodeId(
+                nodeId,
+              )}`,
+            ],
+          });
+          process.stdout.write(formattedOutput);
         }
       } finally {
         if (pkClient! != null) await pkClient.stop();
--- a/src/nodes/CommandFind.ts
+++ b/src/nodes/CommandFind.ts
@@ -90,12 +90,11 @@ class CommandFind extends CommandPolykey {
         }
         let output: any = result;
         if (options.format === 'human') output = [result.message];
-        process.stdout.write(
-          binUtils.outputFormatter({
-            type: options.format === 'json' ? 'json' : 'list',
-            data: output,
-          }),
-        );
+        const formattedOutput = await binUtils.outputFormatter({
+          type: options.format === 'json' ? 'json' : 'list',
+          data: output,
+        });
+        process.stdout.write(formattedOutput);
         // Like ping it should error when failing to find node for automation reasons.
         if (!result.success) {
           throw new errors.ErrorPolykeyCLINodeFindFailed(result.message);
--- a/src/nodes/CommandGetAll.ts
+++ b/src/nodes/CommandGetAll.ts
@@ -60,12 +60,11 @@ class CommandGetAll extends CommandPolykey {
               `NodeId ${value.nodeIdEncoded}, Address ${value.host}:${value.port}, bucketIndex ${value.bucketIndex}`,
           );
         }
-        process.stdout.write(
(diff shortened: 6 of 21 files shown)


addievo:
Sanitization isn't a good solution for num 2. 

This issue is related to this issue:

https://github.com/MatrixAI/Polykey/issues/572

addievo:
Console output from handler for number 2:

```ts
{
  nodeId: IdInternal(32) [Uint8Array] [
    211, 206, 216, 221, 137,  66, 215, 114,
      8, 228,  98, 250, 193, 250, 150,  56,
    210,  34,  94, 186,  71, 180, 100, 130,
    229, 191, 254,  19,  45,  18, 227, 171
  ],
  address: { host: '127.0.0.1', port: 55551, hostname: undefined },
  usageCount: 0,
  timeout: undefined
}
```

So it is more than obvious that this JSON:

```JSON
[{"host":"127.0.0.1","hostname":"","nodeIdEncoded":"vqf7dhnc98bbn4274cbtc3ukm73924nlq8uq690n5nvv16b8iselg","port":55551,"timeout":-1,"usageCount":0,"metadata":{"timeout":15000,"authorization":"Bearer {\"payload\":\"eyJpYXQiOjE2OTgyMTY4NTguNzA1LCJpc3MiOiJ2cTQ4MG9mbGQ3dnJvbDB1cmExdmlpZjk0NzRwYjdwdHJxbTlzOGxkMjN2bjV0OTJvaWY0ZyIsInN1YiI6InZxNDgwb2ZsZDd2cm9sMHVyYTF2aWlmOTQ3NHBiN3B0cnFtOXM4bGQyM3ZuNXQ5Mm9pZjRnIn0\",\"signatures\":[{\"protected\":\"eyJhbGciOiJCTEFLRTJiIn0\",\"signature\":\"UbIRKl5a8AWuNIAK2aCRmgatv0e940ZNZJ2Zvzi_WF4\"}]}"}}]
```

The metadata field isn't from the backend, consequently the issue lies in the frontend.

addievo:
Considering the issue arises in the frontend itself, I think a sensitization should suffice for number 2.

CMCDragonkai:
> Console output from handler for number 2:
> 
> ```ts
> {
>   nodeId: IdInternal(32) [Uint8Array] [
>     211, 206, 216, 221, 137,  66, 215, 114,
>       8, 228,  98, 250, 193, 250, 150,  56,
>     210,  34,  94, 186,  71, 180, 100, 130,
>     229, 191, 254,  19,  45,  18, 227, 171
>   ],
>   address: { host: '127.0.0.1', port: 55551, hostname: undefined },
>   usageCount: 0,
>   timeout: undefined
> }
> ```
> 
> So it is more than obvious that this JSON:
> 
> ```json
> [{"host":"127.0.0.1","hostname":"","nodeIdEncoded":"vqf7dhnc98bbn4274cbtc3ukm73924nlq8uq690n5nvv16b8iselg","port":55551,"timeout":-1,"usageCount":0,"metadata":{"timeout":15000,"authorization":"Bearer {\"payload\":\"eyJpYXQiOjE2OTgyMTY4NTguNzA1LCJpc3MiOiJ2cTQ4MG9mbGQ3dnJvbDB1cmExdmlpZjk0NzRwYjdwdHJxbTlzOGxkMjN2bjV0OTJvaWY0ZyIsInN1YiI6InZxNDgwb2ZsZDd2cm9sMHVyYTF2aWlmOTQ3NHBiN3B0cnFtOXM4bGQyM3ZuNXQ5Mm9pZjRnIn0\",\"signatures\":[{\"protected\":\"eyJhbGciOiJCTEFLRTJiIn0\",\"signature\":\"UbIRKl5a8AWuNIAK2aCRmgatv0e940ZNZJ2Zvzi_WF4\"}]}"}}]
> ```
> 
> The metadata field isn't from the backend, consequently the issue lies in the frontend.

Where did you do this?

CMCDragonkai:
Tasks should be not just fix 1, 2, 3. Summarise the titles.

src/nodes/CommandConnections.ts
```diff
@@ -71,10 +113,18 @@ class CommandAdd extends CommandPolykey {
...
           );
         } else {
+          // Deep copy the connections array to avoid modifying the original object
+          const sanitizedConnections = JSON.parse(JSON.stringify(connections));
+          // Loop through each connection and remove the sensitive 'authorization' field
+          for (const connection of sanitizedConnections) {
+            if (connection.metadata && connection.metadata.authorization) {
+              delete connection.metadata;
+            }
+          }
```
tegefaulkes, line 123: This is a bandaid, not a fix. Find out why metadata is included in the first place. That's the real problem.

addievo:
> > Console output from handler for number 2:
> > ```ts
> > {
> >   nodeId: IdInternal(32) [Uint8Array] [
> >     211, 206, 216, 221, 137,  66, 215, 114,
> >       8, 228,  98, 250, 193, 250, 150,  56,
> >     210,  34,  94, 186,  71, 180, 100, 130,
> >     229, 191, 254,  19,  45,  18, 227, 171
> >   ],
> >   address: { host: '127.0.0.1', port: 55551, hostname: undefined },
> >   usageCount: 0,
> >   timeout: undefined
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > So it is more than obvious that this JSON:
> > ```json
> > [{"host":"127.0.0.1","hostname":"","nodeIdEncoded":"vqf7dhnc98bbn4274cbtc3ukm73924nlq8uq690n5nvv16b8iselg","port":55551,"timeout":-1,"usageCount":0,"metadata":{"timeout":15000,"authorization":"Bearer {\"payload\":\"eyJpYXQiOjE2OTgyMTY4NTguNzA1LCJpc3MiOiJ2cTQ4MG9mbGQ3dnJvbDB1cmExdmlpZjk0NzRwYjdwdHJxbTlzOGxkMjN2bjV0OTJvaWY0ZyIsInN1YiI6InZxNDgwb2ZsZDd2cm9sMHVyYTF2aWlmOTQ3NHBiN3B0cnFtOXM4bGQyM3ZuNXQ5Mm9pZjRnIn0\",\"signatures\":[{\"protected\":\"eyJhbGciOiJCTEFLRTJiIn0\",\"signature\":\"UbIRKl5a8AWuNIAK2aCRmgatv0e940ZNZJ2Zvzi_WF4\"}]}"}}]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The metadata field isn't from the backend, consequently the issue lies in the frontend.
> 
> Where did you do this?

@CMCDragonkai, console.log'ed the handler and ran PK with --verbose to see the output

src/nodes/CommandConnections.ts
addievo: > This is a bandaid, not a fix. Find out why metadata is included in the first place. That's the real problem.

@tegefaulkes I think the core issue is again of rpc, which is this

    return {
      readable: outputMessageTransformStream.readable,
      writable: inputMessageTransformStream.writable,
      cancel: rpcStream.cancel,
      meta: metadata,
    };
    
The client returns the metadata input. 

How do I go about this?

tegefaulkes: Let's reason through this. We're getting metadata in the output and we don't know why. Where is it coming from? It is a server stream call, so we're getting a stream of the output. 

OK, so in the RPC the first message of the response, or any message can include metadata. Ok so we *do* need to strip out the metadata when reading the output.

But how do we do it? doing a `JSON.parse(JSON.stringify())` is a code smell and should almost never be done since it's a very expensive operation done twice. There are better ways to deep copy.

Finally, why not just strip out the metadata when we're collecting the stream on line 50? You don't need a conditional, you just construct a new object with the information you want.

addievo: Resolved

CMCDragonkai:
What's the status on this PR?

addievo:
> What's the status on this PR?

Now it is ready to merge

tegefaulkes:
Is `nodesConnection is is sometimes lising the IP as as a ipv6 mapped ipv4 address. This needs to be looked into.` addressed from the issue?

addievo:
> Is `nodesConnection is is sometimes lising the IP as as a ipv6 mapped ipv4 address. This needs to be looked into.` addressed from the issue?

No, this issue does not address that.

CMCDragonkai:
That should be looked into the ipv4-mapped-ipv6 addresses.

src/utils/utils.ts
```diff
@@ -95,20 +95,30 @@ function outputFormatter(msg: OutputObject): string | Uint8Array {
...
+
+    // Step 2: Use the maxKeyLength to align the keys and values
     for (const key in msg.data) {
       let value = msg.data[key];
       // Empty string for null or undefined values
       if (value == null) {
         value = '';
       }
-      value = JSON.stringify(value);
+      value = JSON.stringify(value).replace(/"/g, '');
```
CMCDragonkai, line 113: This isn't safe. There may be actual `"` inside the values of the JSON. It's better to left and right trim the `"` if the value is of type string.

src/utils/utils.ts
```diff
@@ -95,20 +95,30 @@ function outputFormatter(msg: OutputObject): string | Uint8Array {
...
       // Remove the last line terminator if it exists
-      // This may exist if the value is multi-line string
       value = value.replace(/(?:\r\n|\n)$/, '');
-      // If the string has line terminators internally
-      // Then we must append `\t` separator after each line terminator
-      value = value.replace(/(\r\n|\n)/g, '$1\t');
-      output += `${key}\t${value}\n`;
```
CMCDragonkai, line 111: Why is the `value.replace` being removed?

CMCDragonkai:
Can you rewrite the spec, I think some of the spec here is just plain incorrect.

CMCDragonkai:
Not ready to be merged. Spec does not match implementation details.

src/nodes/CommandConnections.ts
```diff
@@ -48,20 +48,69 @@ class CommandAdd extends CommandPolykey {
...
+
+            const outputLine =
+              hostString.padEnd(maxHostStringLength) +
+              '\t' +
+              usageCount.padEnd(maxUsageCountLength) +
+              '\t' +
+              timeout.padEnd(maxTimeoutLength);
             output.push(outputLine);
           }
           process.stdout.write(
```
CMCDragonkai, line 116: I believe `node connections` should be using a table format. Not a list format.

So this is not correct. A proper refactoring is to go to the output formatter and update the table formatting system first, and then send an array of dictionaries to the output formatter.

Focus on making the output formatter more generically useful will therefore benefit ALL commands, not just this command.

src/agent/CommandStatus.ts
```diff
@@ -79,8 +79,6 @@ class CommandStatus extends CommandPolykey {
               clientPort: response.clientPort,
               agentHost: response.agentHost,
               agentPort: response.agentPort,
-              publicKeyJWK: response.publicKeyJwk,
-              certChainPEM: response.certChainPEM,
```
CMCDragonkai, line 83: Didn't you say you were working on something to expand the `agent status` command to add additional details like number of connections, number of nodes in the graph, number of gestalt vertices, and sigchain claims? Where is that PR?

addievo:
As 
https://github.com/MatrixAI/Polykey-CLI/issues/44

has been bifurcated into a new issue in Polykey, this PR is ready to be reviewed and merged.

I am doing that now.

src/agent/CommandStatus.ts
addievo: That is in a new branch called feature-cli-polish-2, since the expansion is still in discussion, I didn't want to muddy this PR.

I will be creating a PR for that soon.

src/nodes/CommandConnections.ts
addievo: Alright, I'll get on that then.

src/utils/utils.ts
addievo: I modified the code to only remove the starting and ending quotes if the value is of type string.

src/utils/utils.ts
addievo: Added back the logic to handle /t terminators.

CMCDragonkai:
Go to the output formatter issue and solve things there first then review this in accordance.

addievo:
My proposed changes to the `table` type output formatter:

1. Optional header row

This would be an optional parameter which would allow adding a header row to signify which key a column represents in a particular table.

2. Adding padding based on length of longest key in a column.

3. Adding a row count indicator.

addievo:
> Go to the output formatter issue and solve things there first then review this in accordance.

Oh okay

addievo:
@CMCDragonkai,

So this is what I got from https://github.com/MatrixAI/Polykey-CLI/issues/22

1. TSV standardization : Output should be tab seperated values.
2. Handling special characters: Tabs and newlines should be wrapped in quotes.
3. Padding: Make a single pass thru data, and for streams, use a max len and then alter according to data being streamed.

CMCDragonkai:
> @CMCDragonkai the issue with stream based padding, i.e. updating counter for every item in stream is that, that would have to be implemented on a case-to-case basis, since the data is fed into output formatter after it has been all *collected* and added to an array.
> 

No it can be done in the output formatter. Do a scan.

src/utils/utils.ts
```diff
@@ -62,7 +64,93 @@ function standardErrorReplacer(key: string, value: any) {
...
+  // Initialize maxColumnLengths with header lengths if headers are provided
+  if (options?.headers) {
+    for (const header of options.headers) {
+      maxColumnLengths[header] = header.length;
+    }
+  }
+
+  let iterableStream = Array.isArray(rowStream)
+    ? arrayToAsyncIterable(rowStream)
+    : rowStream;
```
tegefaulkes, line 91: Why is this needed? It's perfectly valid to do 

```ts
const array = [1, 2, 3];

for await (const value of array) {
  //do things
}

```

Remember, you can `await` anything, not just promises.

You don't need to convert an array to an async iterator. Unless you want to use iterator calls like `thing.next()`?

src/utils/utils.ts
```diff
@@ -62,7 +64,93 @@ function standardErrorReplacer(key: string, value: any) {
...
+
+  const updateMaxColumnLengths = (row: TableRow) => {
+    for (const [key, value] of Object.entries(row)) {
+      const cellValue =
+        value === null || value === '' || value === undefined ? 'N/A' : value;
```
tegefaulkes, line 96: You can simplify this. If a value is `null` or `undefined` you can use `??` to use a default.

`value ?? 'N/A`, so only `value == ''` needs to be checked. The whole thing can be compressed to `value == '' ? 'N/A' : value ?? 'N/A'`

src/utils/utils.ts
```diff
@@ -62,7 +64,93 @@ function standardErrorReplacer(key: string, value: any) {
   return value;
 }
 
-function outputFormatter(msg: OutputObject): string | Uint8Array {
+async function* arrayToAsyncIterable<T>(array: T[]): AsyncIterable<T> {
+  for (const item of array) {
+    yield item;
+  }
+}
```
tegefaulkes, line 71: This is a very weird thing to need.

src/utils/utils.ts
```diff
@@ -62,7 +64,93 @@ function standardErrorReplacer(key: string, value: any) {
...
+    if (options?.includeRowCount) {
+      formattedRow += `${++rowCount}\t`;
+    }
+
+    const keysToUse = options?.headers ?? Object.keys(maxColumnLengths);
+
+    for (const key of keysToUse) {
+      const cellValue = Object.prototype.hasOwnProperty.call(row, key)
+        ? row[key]
+        : 'N/A';
```
tegefaulkes, line 135: This is very weird. Are you just trying to convert undefined values to `'N/A'`? Just use `cellValue = row[key] ?? 'N/A';`

tegefaulkes:
I see we strip out the `""` on strings when formatting the output. Is that something we really want? How would I tell  if `1235` is different from `"12345"` or if the whitespace was part of the value?

CMCDragonkai:
> I see we strip out the `""` on strings when formatting the output. Is that something we really want? How would I tell if `1235` is different from `"12345"` or if the whitespace was part of the value?

You cannot. We are making a UI decision here. In fact we do space padding anyway in tables to make this work.

Actually now thinking about it, however for a "password      " to work. I imagine you'd want to show the spaces. That's a good question... and perhaps in that case we would want to show the whitespace different from padded.

One issue is then some passwords are `password   \n\n`. Right? So in that case, we cannot encode it with `\n` at all.

In fact in such cases, it's not really useful at all to simply print it with padded whitespace or otherwise.

Maybe then we should keep the double quote....

Yes let's keep the double quotes, they are the only way to know. But I can imagine an issue with things like `\t` or `\v` and other whitespace characters.

I think generally though if we are printing out a secret data, we would print it out literally. For example: singular outputs, list outputs. I believe singular outputs and list outputs is currently the same. I now realise they should be considered separate output formats.

The double quotations only come into play in the more complex outputs... That would be for list, dict, table.

addievo:
> > I see we strip out the `""` on strings when formatting the output. Is that something we really want? How would I tell if `1235` is different from `"12345"` or if the whitespace was part of the value?
> 
> You cannot. We are making a UI decision here. In fact we do space padding anyway in tables to make this work.
> 
> Actually now thinking about it, however for a "password " to work. I imagine you'd want to show the spaces. That's a good question... and perhaps in that case we would want to show the whitespace different from padded.
> 
> One issue is then some passwords are `password \n\n`. Right? So in that case, we cannot encode it with `\n` at all.
> 
> In fact in such cases, it's not really useful at all to simply print it with padded whitespace or otherwise.
> 
> Maybe then we should keep the double quote....
> 
> Yes let's keep the double quotes, they are the only way to know. But I can imagine an issue with things like `\t` or `\v` and other whitespace characters.
> 
> I think generally though if we are printing out a secret data, we would print it out literally. For example: singular outputs, list outputs. I believe singular outputs and list outputs is currently the same. I now realise they should be considered separate output formats.
> 
> The double quotations only come into play in the more complex outputs... That would be for list, dict, table.

I have never seen a password with any whitespace character in between.t

CMCDragonkai:
I have passwords with whitepsaces.

addievo:
Apparently Polykey includes spaces typed while setting a password, which I think is a big UI/UX flaw.

https://security.stackexchange.com/questions/17192/why-disallow-special-characters-in-a-password

addievo:
Also, why would we ever want to print out the password? Isn't that also a big security flaw? Printing our plaintext passwords

src/utils/utils.ts
addievo: Hey @tegefaulkes, so using `for await (const value of array)` is actually worse in terms of space complexity as you have to store the array instead of just reading it, both your approach and mine use the same time complexity of O(2n), which is what Roger wanted, so I don't see whats wrong with my implementation here. Please do elaborate.

src/utils/utils.ts
addievo: Ah right, I was specifically checking for no property, vs. checking for undefined and null as well. Changed it as such

src/utils/utils.ts
addievo: Alright, thanks, that looks much cleaner.

CMCDragonkai:
It's not just passwords. There are other kinds of text too. I'm thinking secrets in the vaults. I think you need to add the quotes back. And then special encode any non printable characters except spaces.

addievo:
#50

src/utils/utils.ts
tegefaulkes: I think you misunderstand, I'm saying you can consume an array and an `AsyncIterable` the same way. You don't need to convert an array into an `AsyncIterable` to consume it with a async for of loop.

amydevs: > Hey @tegefaulkes, so using `for await (const value of array)` is actually worse in terms of space complexity as you have to store the array instead of just reading it, both your approach and mine use the same time complexity of O(2n), which is what Roger wanted, so I don't see whats wrong with my implementation here. Please do elaborate.

i don't think there is any performance difference between

```
const array = [1, 2, 3];

for await (const value of array) {
  //do things
}
```

and

```
const array = [1, 2, 3];

for (const value of array) {
  //do things
}
```

Also, `arrayToAsyncIterable` would create a AsyncIterable instance, where as  `for (const value of array)` wouldn't need to

src/nodes/CommandConnections.ts
amydevs: how was this resolved @addievo ?

src/utils/utils.ts
amydevs: it's been removed

src/utils/utils.ts
amydevs: I changed this to
```ts
const cellValue = row[key] == null || row[key].length === 0 ? 'N/A' : row[key];
```
so it's a bit more readable

PR #50, opened 2023-11-02
Fixing outputFormatter and standardising STDOUT

addievo:
### Description

This PR builds upon https://github.com/MatrixAI/Polykey-CLI/pull/45

And aims to fix some issues which were not resolved in their entirety in #45. 

#### Automatic Encoding

For these two formats, encoding is applied by default so that special characters will not mess up the padding for any given output.
- dict format
  - will automatically encase any string in a `\"\"` and apply encoding to the input
- table format
  - will automatically encase any string in a `\"\"` and apply encoding to the input

#### Manual Encoding with Utility Functions

For any other format passed into `outputFormatter`, there will be no encoding of escapable characters by default.

It is up to the user to use the following functions in `src/utils/utils.ts` to encode parts of their input:

```ts
/**
 * This function:
 *
 * 1. Keeps regular spaces, only ' ', as they are.
 * 2. Converts \\n \\r \\t to escaped versions, \\\\n \\\\r and \\\\t.
 * 3. Converts other control characters to their Unicode escape sequences.\
 * 4. Converts ' \` " to escaped versions, \\\\' \\\\\` and \\\\"
 *
 * Unless you're using this in a `JSON.stringify` replacer, you probably want to use {@link encodeEscapedWrapped} instead.
 */
function encodeEscaped(str: string): string
```

```ts
/**
 * This function:
 * 1. Keeps regular spaces, only ' ', as they are.
 * 2. Converts \\n \\r \\t to escaped versions, \\\\n \\\\r and \\\\t.
 * 3. Converts other control characters to their Unicode escape sequences.
 * 4. Converts ' \` " to escaped versions, \\\\' \\\\\` and \\\\"
 * 5. Wraps the whole thing in `""` if any characters have been encoded.
 */
function encodeEscapedWrapped(str: string): string
```

To use this:

```ts
const v = `test ${encodeEscapedWrapped("\u0000\"\n")}` // returns '"\\u0000\\"\\n"'
```

Note that any encoded strings must be encased in `""` anyways. This is so that when the user copy-pastes the output as the input of another command, or pipes the output to another command, those escaped characters will be collapsed correctly

These mainly include:

1. Simplification of outputTableFormatter.
2. Adding quotes back on stdout.
3. Encode non-printable characters except spaces.

### Tasks

- [x] 1. Tidy up outputTableFormatter.
- [x] 2. Adding double quotes back on STDOUT.
- [x] 3. Encoding non-printable chars except spaces on STDOUT.
- [x] 4. Define and implement new `encodeEscaped` and `encodeEscapedWrapped` API for output character escape encoding

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/src/nodes/CommandConnections.ts
+++ b/src/nodes/CommandConnections.ts
@@ -59,7 +59,7 @@ class CommandAdd extends CommandPolykey {
         }, auth);
         if (options.format === 'human') {
           // Wait for outputFormatter to complete and then write to stdout
-          const formattedOutput = await binUtils.outputFormatter({
+          const formattedOutput = binUtils.outputFormatter({
             type: 'table',
             data: connections,
             options: {
@@ -76,7 +76,7 @@ class CommandAdd extends CommandPolykey {
           process.stdout.write(formattedOutput);
         } else {
           // Wait for outputFormatter to complete and then write to stdout
-          const formattedOutput = await binUtils.outputFormatter({
+          const formattedOutput = binUtils.outputFormatter({
             type: 'json',
             data: connections,
           });
--- a/src/nodes/CommandConnections.ts
+++ b/src/nodes/CommandConnections.ts
@@ -63,7 +63,7 @@ class CommandAdd extends CommandPolykey {
             type: 'table',
             data: connections,
             options: {
-              headers: [
+              columns: [
                 'host',
                 'hostname',
                 'nodeIdEncoded',
@@ -71,6 +71,7 @@ class CommandAdd extends CommandPolykey {
                 'timeout',
                 'usageCount',
               ],
+              includeHeaders: true,
             },
           });
           process.stdout.write(formattedOutput);
--- a/src/nodes/CommandConnections.ts
+++ b/src/nodes/CommandConnections.ts
@@ -59,7 +59,7 @@ class CommandAdd extends CommandPolykey {
         }, auth);
         if (options.format === 'human') {
           // Wait for outputFormatter to complete and then write to stdout
-          const formattedOutput = binUtils.outputFormatter({
+          const outputFormatted = binUtils.outputFormatter({
             type: 'table',
             data: connections,
             options: {
@@ -74,14 +74,14 @@ class CommandAdd extends CommandPolykey {
               includeHeaders: true,
             },
           });
-          process.stdout.write(formattedOutput);
+          process.stdout.write(outputFormatted);
         } else {
           // Wait for outputFormatter to complete and then write to stdout
-          const formattedOutput = binUtils.outputFormatter({
+          const outputFormatted = binUtils.outputFormatter({
             type: 'json',
             data: connections,
           });
-          process.stdout.write(formattedOutput);
+          process.stdout.write(outputFormatted);
         }
       } finally {
         if (pkClient! != null) await pkClient.stop();
--- a/src/types.ts
+++ b/src/types.ts
@@ -43,7 +43,7 @@ type AgentChildProcessOutput =
 type TableRow = Record<string, any>;
 
 interface TableOptions {
-  headers?: string[];
+  headers?: Array<string>;
   includeRowCount?: boolean;
 }
 
--- a/src/types.ts
+++ b/src/types.ts
@@ -5,6 +5,15 @@ import type { RecoveryCode } from 'polykey/dist/keys/types';
 import type { StatusLive } from 'polykey/dist/status/types';
 import type { NodeIdEncoded } from 'polykey/dist/ids/types';
 
+
+type TableRow = Record<string, any>;
+
+interface TableOptions {
+  columns?: Array<string> | Record<string, number>;
+  includeHeaders?: boolean;
+  includeRowCount?: boolean;
+}
+
 type AgentStatusLiveData = Omit<StatusLive['data'], 'nodeId'> & {
   nodeId: NodeIdEncoded;
 };
@@ -40,17 +49,10 @@ type AgentChildProcessOutput =
       error: POJO;
     };
 
-type TableRow = Record<string, any>;
-
-interface TableOptions {
-  headers?: Array<string>;
-  includeRowCount?: boolean;
-}
-
 export type {
+  TableRow,
+  TableOptions,
   AgentStatusLiveData,
   AgentChildProcessInput,
   AgentChildProcessOutput,
-  TableRow,
-  TableOptions,
 };
--- a/src/types.ts
+++ b/src/types.ts
@@ -5,7 +5,6 @@ import type { RecoveryCode } from 'polykey/dist/keys/types';
 import type { StatusLive } from 'polykey/dist/status/types';
 import type { NodeIdEncoded } from 'polykey/dist/ids/types';
 
-
 type TableRow = Record<string, any>;
 
 interface TableOptions {
--- a/src/utils/utils.ts
+++ b/src/utils/utils.ts
@@ -95,35 +95,61 @@ function encodeNonPrintable(str: string) {
   });
 }
 
-// Function to handle 'table' type output
+/**
+ * Function to handle the `table` output format.
+ * @param rows
+ * @param options
+ * @param options.columns - Can either be an `Array<string>` or `Record<string, number>`.
+ * If it is `Record<string, number>`, the `number` values will be used as the initial padding lengths.
+ * The object is also mutated if any cells exceed the inital padding lengths.
+ * This paramater can also be supplied to filter the columns that will be displayed.
+ * @param options.includeHeaders - Defaults to `True`
+ * @param options.includeRowCount - Defaults to `False`.
+ * @returns
+ */
 function outputTableFormatter(
-  rowStream: Array<TableRow>,
-  options?: TableOptions,
+  rows: Array<TableRow>,
+  options: TableOptions = {
+    includeHeaders: true,
+    includeRowCount: false,
+  },
 ): string {
   let output = '';
   let rowCount = 0;
+  // Default includeHeaders to true
+  const includeHeaders = options.includeHeaders ?? true;
   const maxColumnLengths: Record<string, number> = {};
 
+  const optionColumns =
+    options?.columns != null
+      ? Array.isArray(options.columns)
+        ? options.columns
+        : Object.keys(options.columns)
+      : undefined;
+
   // Initialize maxColumnLengths with header lengths if headers are provided
-  if (options?.headers) {
-    for (const header of options.headers) {
-      maxColumnLengths[header] = header.length;
+  if (optionColumns != null) {
+    for (const column of optionColumns) {
+      maxColumnLengths[column] = Math.max(
+        options?.columns?.[column] ?? 0,
+        column.length,
+      );
     }
   }
 
   // Precompute max column lengths by iterating over the rows first
-  for (const row of rowStream) {
-    for (const key in options?.headers ?? row) {
-      if (row[key] != null) {
-        row[key] = encodeNonPrintable(row[key].toString());
+  for (const row of rows) {
+    for (const column in options?.columns ?? row) {
+      if (row[column] != null) {
+        row[column] = encodeNonPrintable(row[column].toString());
       }
       // Row[key] is definitely a string or null after this point due to encodeNonPrintable
-      const cellValue: string | null = row[key];
+      const cellValue: string | null = row[column];
       // Null or '' will both cause cellLength to be 3
       const cellLength =
         cellValue == null || cellValue.length === 0 ? 3 : cellValue.length; // 3 is length of 'N/A'
-      maxColumnLengths[key] = Math.max(
-        maxColumnLengths[key] || 0,
+      maxColumnLengths[column] = Math.max(
+        maxColumnLengths[column] || 0,
         cellLength, // Use the length of the encoded value
       );
     }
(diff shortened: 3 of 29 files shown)


src/utils/utils.ts
```diff
@@ -70,6 +70,39 @@ async function* arrayToAsyncIterable<T>(array: T[]): AsyncIterable<T> {
...
+This function:
+1. Keeps regular spaces, only ' ', as they are.
+2. Converts \n \r \t to escaped versions, \\n \\r and \\t.
+3. Converts other control characters to their Unicode escape sequences.
+ */
+
+function encodeNonPrintable(input) {
+  // Ensure the input is a string
+  const str = String(input);
+  // eslint-disable-next-line
```
tegefaulkes, line 83: 1. Can't you specify `input: string` if you want it to be a string? Why cast it here with `String()`?
2. never use `eslint-disable-next-line` unless you absolutely have to, and if you do you *must* include the reasoning with a comment. And you must specify the exact rule you are ignoring line ` eslint-disable-next-line no-control-regex`. Generally we want a very good reason for doing this.

src/utils/utils.ts
```diff
@@ -90,13 +123,22 @@ const outputTableFormatter = async (
...
 
+  const encodeRowValues = (row: TableRow) => {
+    for (const key in row) {
+      if (row[key] != null) {
+        // Convert the value to a string and encode non-printable characters
+        row[key] = encodeNonPrintable(row[key].toString());
+      }
+    }
+  };
+
```
tegefaulkes, line 134: Why is this an arrow function? It doesn't need to be. Unless you have to use it as a closure then do define it like this. Make it a separate function.

It's a small thing, but each new closure allocates new memory and holds reference to the scope it was created in. It's best to avoid doing it.

tests/utils.test.ts
```diff
@@ -3,9 +3,18 @@ import ErrorPolykey from 'polykey/dist/ErrorPolykey';
...
 import * as nodesUtils from 'polykey/dist/nodes/utils';
 import * as polykeyErrors from 'polykey/dist/errors';
+import * as fc from 'fast-check';
 import * as binUtils from '@/utils/utils';
 import * as testUtils from './utils';
 
+const nonPrintableCharArbitrary = fc.oneof(
+  fc.integer({ min: 0, max: 0x1f }).map((code) => String.fromCharCode(code)),
+  fc.integer({ min: 0x7f, max: 0x9f }).map((code) => String.fromCharCode(code)),
+);
```
tegefaulkes, line 13:
```suggestion
const nonPrintableCharArbitrary = fc.oneof(
  fc.integer({ min: 0, max: 0x1f }),
  fc.integer({ min: 0x7f, max: 0x9f }),
).map((code) => String.fromCharCode(code));
```

tests/utils.test.ts
```diff
@@ -41,12 +47,11 @@ describe('bin/utils', () => {
         ],
       });
       expect(tableOutput).toBe(
-        'value1\tvalue2\ndata1 \tdata2\nundefined\tundefined\n',
+        'value1\tvalue2\n' + 'data1 \tdata2\n' + 'N/A   \tN/A\n',
```
tegefaulkes, line 50: Why concat the string here, can't it be all one string?

tegefaulkes (changes requested):
The code doesn't build. That *needs* to be fixed before merging.

It was broken in #45 too so I'm not sure why that was merged. You need to follow the merge checklist. It's there to catch this problem. But it's fully ticked right now even though the build fails. Why is that?

CMCDragonkai:
There was prior review commentary located here: https://github.com/MatrixAI/Polykey-CLI/pull/45#discussion_r1381138471.

@amydevs best to refer to the prior comments that @tegefaulkes made too.

src/utils/utils.ts
amydevs: why are there two loops as well?

tests/utils.test.ts
amydevs: it's been made into just one string

tests/utils.test.ts
amydevs: this has been condensed

src/utils/utils.ts
amydevs: It's been made into just one loop, and now the `cellLength` rather than the `cellValue` is determined so that extra strings don't need to be instantiated:

```ts
    for (const key in row) {
      if (row[key] != null) {
        row[key] = encodeNonPrintable(row[key].toString());
      }
      // Row[key] is definitely a string or null after this point due to encodeNonPrintable
      const cellValue: string | null = row[key];
      // Null or '' will both cause cellLength to be 3
      const cellLength =
        cellValue == null || cellValue.length === 0 ? 3 : cellValue.length; // 3 is length of 'N/A'
      maxColumnLengths[key] = Math.max(
        maxColumnLengths[key] || 0,
        cellLength, // Use the length of the encoded value
      );
    }
```

src/utils/utils.ts
amydevs: the Docbloc has been cleaned up, the type signature now takes a string, and a comment has been added for reasoning as to why `eslint-disable-next-line no-control-regex` is used
```ts
/**
 * This function:
 * 1. Keeps regular spaces, only ' ', as they are.
 * 2. Converts \n \r \t to escaped versions, \\n \\r and \\t.
 * 3. Converts other control characters to their Unicode escape sequences.
 */
function encodeNonPrintable(str: string) {
  // We want to actually match control codes here!
  // eslint-disable-next-line no-control-regex
  return str.replace(/[\x00-\x1F\x7F-\x9F]/g, (char) => {
    switch (char) {
      case ' ':
        return char; // Preserve regular space
      case '\n':
        return '\\n'; // Encode newline
      case '\r':
        return '\\r'; // Encode carriage return
      case '\t':
        return '\\t'; // Encode tab
      case '\v':
        return '\\v'; // Encode tab
      case '\f':
        return '\\f'; // Encode tab
      // Add cases for other whitespace characters if needed
      default:
        // Return the Unicode escape sequence for control characters
        return `\\u${char.charCodeAt(0).toString(16).padStart(4, '0')}`;
    }
  });
}
```

src/agent/CommandStatus.ts
```diff
@@ -79,8 +79,6 @@ class CommandStatus extends CommandPolykey {
               clientPort: response.clientPort,
               agentHost: response.agentHost,
               agentPort: response.agentPort,
-              publicKeyJWK: response.publicKeyJwk,
-              certChainPEM: response.certChainPEM,
```
CMCDragonkai, line 83: Does this fix any particular issue?

src/nodes/CommandConnections.ts
```diff
@@ -48,35 +48,42 @@ class CommandAdd extends CommandPolykey {
...
-            output.push(outputLine);
-          }
-          process.stdout.write(
-            binUtils.outputFormatter({
-              type: 'list',
-              data: output,
-            }),
-          );
+          // Wait for outputFormatter to complete and then write to stdout
+          const formattedOutput = binUtils.outputFormatter({
```
CMCDragonkai, line 64: Prefer `outputFormatted = ...`. SOV... and I usually put adjectives at the end.

src/types.ts
```diff
@@ -40,8 +40,18 @@ type AgentChildProcessOutput =
...
+  includeRowCount?: boolean;
+}
+
 export type {
   AgentStatusLiveData,
   AgentChildProcessInput,
   AgentChildProcessOutput,
+  TableRow,
+  TableOptions,
 };
```
CMCDragonkai, line 57: Since these are more generic types, put them at the top of the exports list.

src/utils/utils.ts
```diff
@@ -62,53 +64,177 @@ function standardErrorReplacer(key: string, value: any) {
...
+}
+
+/**
+ * Function to handle the `table` output format.
+ * @param rows
+ * @param options
+ * @param options.columns - Can either be an `Array<string>` or `Record<string, number>`.
+ * If it is `Record<string, number>`, the `number` values will be used as the initial padding lengths.
+ * The object is also mutated if any cells exceed the inital padding lengths.
+ * This paramater can also be supplied to filter the columns that will be displayed.
```
CMCDragonkai, line 105: It is `parameter`.

src/utils/utils.ts
```diff
@@ -62,53 +64,177 @@ function standardErrorReplacer(key: string, value: any) {
...
+      default:
+        // Return the Unicode escape sequence for control characters
+        return `\\u${char.charCodeAt(0).toString(16).padStart(4, '0')}`;
+    }
+  });
+}
+
+/**
+ * Function to handle the `table` output format.
+ * @param rows
```
CMCDragonkai, line 100: Always newline separation.

src/utils/utils.ts
```diff
@@ -62,53 +64,177 @@ function standardErrorReplacer(key: string, value: any) {
...
-        // Empty string for null or undefined values
-        if (value == null) {
-          value = '';
-        }
-        value = value.toString();
-        // Remove the last line terminator if it exists
-        // This may exist if the value is multi-line string
-        value = value.replace(/(?:\r\n|\n)$/, '');
-        output += `${value}\t`;
+    return outputTableFormatter(msg.data, msg.options);
```
CMCDragonkai, line 214: If this is going to be a separate function, then factor out all the other ones too.

And I suggest:

* `outputFormatterTable`
* `outputFormatterList`
* `outputFormatterDict` ...

src/utils/utils.ts
```diff
@@ -62,53 +64,177 @@ function standardErrorReplacer(key: string, value: any) {
...
+  return output;
+}
+
+/**
+ * Formats a message suitable for output.
+ * @param msg - The msg that needs to be formatted.
+ * @see {@link outputTableFormatter} for information regarding usage where `msg.type === 'table'`.
+ * @returns
+ */
 function outputFormatter(msg: OutputObject): string | Uint8Array {
```
CMCDragonkai, line 204: Question, is `Uint8Array` used for `raw`?

tests/keys/cert.test.ts
```diff
@@ -44,7 +44,7 @@ describe('cert', () => {
       },
     ));
     expect(exitCode).toBe(0);
-    const certStatus = JSON.parse(stdout).certChainPEM;
+    const certStatus = JSON.parse(stdout).cert;
```
CMCDragonkai, line 47: Is this the same thing as the cert chain PEM? As in we should have something that tests for the cert chain! Not just the root cert pem.

tests/nodes/connections.test.ts
```diff
@@ -0,0 +1,129 @@
...
+    nodePath = path.join(dataDir, 'keynode');
+    pkAgent = await PolykeyAgent.createPolykeyAgent({
+      password,
+      options: {
+        seedNodes: {}, // Explicitly no seed nodes on startup
+        nodePath,
+        agentServiceHost: '127.0.0.1',
+        clientServiceHost: '127.0.0.1',
+        agentServicePort: 55555,
+        clientServicePort: 55554,
```
CMCDragonkai, line 35: Can these ports be `0` and just dynamically refer to them? I avoid hardcoding ports even in tests.

CMCDragonkai:
This is conflicting with staging? You may need a rebase?

And also can you provide previews of what all these commands look like now?

src/utils/utils.ts
amydevs: yes, for `CommandGet` of `secrets` domain

src/utils/utils.ts
amydevs: resolved by adding new line

src/utils/utils.ts
amydevs: resolved by fixing typo

src/types.ts
amydevs: resolved by moving types up

src/agent/CommandStatus.ts
amydevs: i don't think so, just @addievo mentioned that it had little usage in status, so it was removed:
https://github.com/MatrixAI/Polykey-CLI/pull/45

src/nodes/CommandConnections.ts
amydevs: renamed all instances of it

src/utils/utils.ts
amydevs: resolved by 4fd8ce70aad2d93348645a602d19d795595e45eb

tests/keys/cert.test.ts
amydevs: i'm pretty sure `tests/keys/certchain.test.ts` applies certchain tests

tests/nodes/connections.test.ts
amydevs: this has been changed, no need to dynamically refer to them, as the nodepaths are used in the tests already

tests/keys/cert.test.ts
amydevs: speaking of, the second part of the test in `tests/keys/certchain.test.ts` is no longer needed and wasn't correct in the first place anyways, as `const certChainStatus = JSON.parse(stdout).rootCertChainPem;` never existed on the `agent status` command

amydevs: the second part of the similar test in `cert.test.ts` is also wrong, but doesn't matter anyway now that that certificates have been removed from `agent status` command

CMCDragonkai: What do you suggest?

amydevs: just removing the second part of those tests, they were meant to test `agent status` printing the certificates, but that is no longer the intended behaviour so i've removed it

CMCDragonkai: Ok so is this resolved?

amydevs: yes

amydevs:
> This is conflicting with staging? You may need a rebase?
> 
> And also can you provide previews of what all these commands look like now?

Dict:

image

List:

image

CMCDragonkai:
Vault IDs should not have space characters nor non-printable characters, can we eliminate it from the generated IDs? Or is due to `IdRandom` or something?

CMCDragonkai:
Surely we can choose a particular base encoding that prevents such uses.

CMCDragonkai:
@amydevs are you going to change the vault ID generation?

It doesn't look right to me, the original code never encoded the tabs.

amydevs:
> @amydevs are you going to change the vault ID generation?
> 
> It doesn't look right to me, the original code never encoded the tabs.

no, what i'm saying is that a lot of the commands were using `\t` before for spacing.

like this
```ts
data.push(
              `${vaultListMessage.vaultName}:\t${
                vaultListMessage.vaultIdEncoded
              }`,
            );
```

I've replaced it with spaces instead

```ts
data.push(
              `${vaultListMessage.vaultName}:${' '.repeat(4)}${
                vaultListMessage.vaultIdEncoded
              }`,
            );
```

amydevs:
https://stackoverflow.com/questions/20585729/how-to-detect-if-nodes-process-stdout-is-being-piped

We might want to never escape when the output is being redirected btw

amydevs:
should be done, just wrote up new api for encoding escape characters. Please refer to this PR if it comes up

src/utils/utils.ts
```diff
@@ -50,7 +50,7 @@ type OutputObject =
       data: Error;
     };
 
-function standardErrorReplacer(key: string, value: any) {
```
CMCDragonkai, line 53: How come the `_key` is not used here?

src/utils/utils.ts
CMCDragonkai, line 1: The escaped names is sort of generic.

Would have preferred if you called them "quoted".

CMCDragonkai: Also did you bench these somewhat to be aware of speed?

CMCDragonkai:
What if there's other kinds non-printable characters? Usually when quoted in JS, they appear as slash unicodes. Also I was wondering if we could have just used the standard console.log and captured it's output somehow. JS seems to have native quoting when it comes to strings. Did you have a look?

amydevs:
> What if there's other kinds non-printable characters? Usually when quoted in JS, they appear as slash unicodes. Also I was wondering if we could have just used the standard console.log and captured it's output somehow. JS seems to have native quoting when it comes to strings. Did you have a look?

native quoting only applies to the browser console, not the STDOUT in Node.JS

image

src/utils/utils.ts
amydevs: resolved in 84e7c348f84138787b90ead0a47528b9af789b7d

src/utils/utils.ts
amydevs: not yet, but i think there's not really anyway to make it faster. I've merged three nested regex queries into one, and the query itself isn't too complicated.

CMCDragonkai: Ok.

Issue #40, opened 2023-10-19
CLI Beta Launch


CMCDragonkai:
### Specification

This epic focuses on the CLI beta launch targeting November 10 2023.

This follows from the 6th testnet deployment https://github.com/MatrixAI/Polykey/issues/551 and focuses on any UX issues that should be resolved before we launch. Plus any documentation related things and metrics for tracking how the launch went, as well as content we need to write to prepare for it.

We should also get our demo video fully polished as well.

One of the things we need to do is:

1. Add a `audit` or `logs` domain to PK at first to track information about the network, we will need this as part of our metrics.
2. Create a simple dashboard on `testnet.polykey.com` and `mainnet.polykey.com` so we can see what's going on and show everybody how the network is building up.

We are currently working through this list of issues.
https://github.com/MatrixAI/Polykey-CLI/issues/40#issuecomment-1853222990

### Additional context

* https://github.com/MatrixAI/Polykey/issues/551

### Tasks

1. Spec out the experience for the launch
4. Schedule content releases
5. Plan out all UX-related issues

CMCDragonkai:
https://github.com/MatrixAI/Polykey/pull/600 was merged incorrectly, https://github.com/MatrixAI/Polykey/pull/601 is being done by @addievo to fix up the additional options of `privateKey` and `privateKeyPath` and `recoveryCode`. Also they would need to be exposed on `PolykeyAgent.start` too. See review comments in https://github.com/MatrixAI/Polykey/pull/600

tegefaulkes:
I finished up and merged https://github.com/MatrixAI/Polykey/pull/601. Having a look over the CLI now.

CMCDragonkai:
The `PolykeyClient` is going to be create/destroy only, I think it's pretty clear that can work.

CMCDragonkai:
The change to the `GitHubProvider` to using graphql for getting connected identities is now causing occassional failures in the `IdentitiesManager.test.ts`. I've seen it on CI and also locally.

```
 FAIL  tests/identities/IdentitiesManager.test.ts (10.559 s)
  IdentitiesManager
    ✓ IdentitiesManager readiness (154 ms)
    ✓ get, set and unset tokens (with seed=1486144507) (1350 ms)
    ✓ start and stop preserves state (with seed=1486144507) (966 ms)
    ✓ fresh start deletes all state (with seed=1486144507) (922 ms)
    ✓ register and unregister providers (103 ms)
[...]

PR #56, opened 2023-11-10
Stdout standardization with output formatter

### LINKED ISSUE #22: Standardise STDOUT output for CLI commands

amydevs:
### Description
This PR fixes all the CLI stdout inconsistencies, this is kind of tedious and not critical at the moment though.

#### Changelog

- agent
  - CommandStatus - versionMetadata is no longer spread into parent object with JSON format. 
  - CommandBootstrap - Now uses dict format to output the recoveryCode.
- identities
  - CommandAuthenticate - Now uses dict format
  - CommandClaim - Dict Output, JSON format is now of type `{ claimId: string; url?: string }`.
  - CommandInvite - Feedback is now moved to `logger.info` so that it goes to stderr
  - CommandList output should be reconsidered due to nested structure. *
  - CommandSearch and CommandGet need streamlining in terms of their output *
- keys
  - CommandCert - Uses dict format
  - CommandCertchain - The message at the start is no longer included.
  - CommandEncrypt - Raw format
  - CommandDecrypt - Raw format
  - CommandSign - now uses camelcase for non-json formats
  - CommandVerify - now uses camelcase for non-json formats
  - CommandPublic, Private, and Pair will need to have further consideration due to nested structure. *
- nodes
  - CommandClaim - Stderr feedback and stdout of success *

#### Dict Format

This PR also implements changes to the dict format that allows for nesting.

Changes made to commands with this:

`agent/CommandStatus`
```
status                  LIVE
pid                     90556
nodeId                  vq3pk7bip927v2ivk7457p5g7b4q3ad0jksqrkv9rfkhl31l1ukng
clientHost              ::1
clientPort              33179
agentHost               ::
agentPort               35454
upTime                  9
startTime               1711068204
connectionsActive       2
nodesTotal              8
version                 1.2.1-alpha.50-1-1
sourceVersion           1.2.1-alpha.50
stateVersion            1
networkVersion          1
versionMetadata  
  version               0.2.3
  commitHash            9275aa5d0d5470e9e586a6fbbdc8ca150c37e5aa
  libVersion            1.2.1-alpha.50-1-1
  libSourceVersion      1.2.1-alpha.50
  libStateVersion       1
  libNetworkVersion     1
```

`identities/CommandList`
```
gestalt
  actionsList       notify
  identities 
    github.com:amydevs
  nodeIds    
    vq3pk7bip927v2ivk7457p5g7b4q3ad0jksqrkv9rfkhl31l1ukng
gestalt
  actionsList       scan
  identities 
    github.com:amydevs
  nodeIds    
    vq3pk7bip927v2ivk7457p5g7b4q3ad0jksqrkv9rfkhl31l1ukng
```

`notificiations/CommandRead`
```
notificiation
  data  
    message test message 1
    type    General
  iat       1711063520.438
  isRead    true
  iss       vmqmqq9h4jgduaeifep1nko1j2uebgg0uh48sc4fl9upqdsijjmf0
  sub       voeo52qencdh720o084mlfpgb1st09vjca0ti9qvm6e0p19ma316g
  typ       notification
notificiation
  data  
    message test message 2
    type    General
  iat       1711063520.438
  isRead    true
  iss       vmqmqq9h4jgduaeifep1nko1j2uebgg0uh48sc4fl9upqdsijjmf0
  sub       voeo52qencdh720o084mlfpgb1st09vjca0ti9qvm6e0p19ma316g
  typ       notification
```

### Issues Fixed

* Related #22 (doesn't implement verbosity options) 
* https://github.com/MatrixAI/Polykey-CLI/issues/158

### Tasks

- [x] 1. Fix all necessary `list` formatted outputs to just be strings.
- [x] 2. Fix all necessary `list` formatted outputs to just be `dicts`.
- [x] 3. Fix `json` formatted outputs to return structured POJO rather than just an array with the message itself.
- [x] 4. Implement nesting in `dict` format
- [x] 5. Fix related tests.

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/src/secrets/CommandList.ts
+++ b/src/secrets/CommandList.ts
@@ -46,23 +46,36 @@ class CommandList extends CommandPolykey {
           logger: this.logger.getChild(PolykeyClient.name),
         });
         const data = await binUtils.retryAuthentication(async (auth) => {
-          const data: Array<string> = [];
+          const data: Array<{ secretName: string }> = [];
           const stream = await pkClient.rpcClient.methods.vaultsSecretsList({
             metadata: auth,
             nameOrId: vaultName,
           });
           for await (const secret of stream) {
-            data.push(secret.secretName);
+            data.push({
+              secretName: secret.secretName,
+            });
           }
           return data;
         }, auth);
 
-        const outputFormatted = binUtils.outputFormatter({
-          type: options.format === 'json' ? 'json' : 'list',
-          data: data,
-        });
-
-        process.stdout.write(outputFormatted);
+        if (options.format === 'json') {
+          process.stdout.write(
+            binUtils.outputFormatter({
+              type: 'json',
+              data: data,
+            }),
+          );
+        } else {
+          process.stdout.write(
+            binUtils.outputFormatter({
+              type: 'list',
+              data: data.map(
+                (secretsListMessage) => secretsListMessage.secretName,
+              ),
+            }),
+          );
+        }
       } finally {
         if (pkClient! != null) await pkClient.stop();
       }
--- a/src/agent/CommandStatus.ts
+++ b/src/agent/CommandStatus.ts
@@ -81,15 +81,14 @@ class CommandStatus extends CommandPolykey {
               agentHost: response.agentHost,
               agentPort: response.agentPort,
               upTime: getUnixtime() - response.startTime,
+              startTime: response.startTime,
               connectionsActive: response.connectionsActive,
               nodesTotal: response.nodesTotal,
               version: response.version,
               sourceVersion: response.sourceVersion,
               stateVersion: response.stateVersion,
               networkVersion: response.networkVersion,
-              ...(options.format === 'json'
-                ? { versionMetadata: response.versionMetadata }
-                : response.versionMetadata),
+              versionMetadata: response.versionMetadata,
             },
           }),
         );
--- a/src/agent/CommandStatus.ts
+++ b/src/agent/CommandStatus.ts
@@ -87,7 +87,9 @@ class CommandStatus extends CommandPolykey {
               sourceVersion: response.sourceVersion,
               stateVersion: response.stateVersion,
               networkVersion: response.networkVersion,
-              ...response.versionMetadata,
+              ...(options.format === 'json'
+                ? { versionMetadata: response.versionMetadata }
+                : response.versionMetadata),
             },
           }),
         );
--- a/src/agent/CommandStop.ts
+++ b/src/agent/CommandStop.ts
@@ -27,13 +27,13 @@ class CommandStop extends CommandPolykey {
       );
       const statusInfo = clientStatus.statusInfo;
       if (statusInfo?.status === 'DEAD') {
-        this.logger.info('Agent is already dead');
+        process.stderr.write('Agent is already dead\n');
         return;
       } else if (statusInfo?.status === 'STOPPING') {
-        this.logger.info('Agent is already stopping');
+        process.stderr.write('Agent is already stopping\n');
         return;
       } else if (statusInfo?.status === 'STARTING') {
-        throw new errors.ErrorPolykeyCLIAgentStatus('agent is starting');
+        throw new errors.ErrorPolykeyCLIAgentStatus('Agent is starting');
       }
       const auth = await binProcessors.processAuthentication(
         options.passwordFile,
@@ -62,7 +62,7 @@ class CommandStop extends CommandPolykey {
             }),
           auth,
         );
-        this.logger.info('Stopping Agent');
+        process.stderr.write('Stopping Agent\n');
       } finally {
         if (pkClient! != null) await pkClient.stop();
       }
--- a/src/bootstrap/CommandBootstrap.ts
+++ b/src/bootstrap/CommandBootstrap.ts
@@ -1,5 +1,6 @@
 import process from 'process';
 import CommandPolykey from '../CommandPolykey';
+import * as binUtils from '../utils';
 import * as binOptions from '../utils/options';
 import * as binProcessors from '../utils/processors';
 
@@ -36,7 +37,14 @@ class CommandBootstrap extends CommandPolykey {
         logger: this.logger,
       });
       this.logger.info(`Bootstrapped ${options.nodePath}`);
-      if (recoveryCodeOut != null) process.stdout.write(recoveryCodeOut + '\n');
+      process.stdout.write(
+        binUtils.outputFormatter({
+          type: options.format === 'json' ? 'json' : 'dict',
+          data: {
+            recoveryCode: recoveryCodeOut,
+          },
+        }),
+      );
     });
   }
 }
--- a/src/identities/CommandAuthenticate.ts
+++ b/src/identities/CommandAuthenticate.ts
@@ -80,17 +80,27 @@ class CommandAuthenticate extends CommandPolykey {
                 if (e.code !== 'ENOENT') throw e;
                 // Otherwise we ignore `ENOENT` as a failure to spawn a browser
               }
-              process.stdout.write(
-                binUtils.outputFormatter({
-                  type: options.format === 'json' ? 'json' : 'dict',
-                  data: {
-                    url: message.request.url,
-                    ...(options.format === 'json'
-                      ? { dataMap: message.request.dataMap }
-                      : message.request.dataMap),
-                  },
-                }),
-              );
+              if (options.format === 'json') {
+                process.stdout.write(
+                  binUtils.outputFormatter({
+                    type: 'json',
+                    data: {
+                      url: message.request.url,
+                      dataMap: message.request.dataMap,
+                    },
+                  }),
+                );
+              } else {
+                process.stdout.write(
+                  binUtils.outputFormatter({
+                    type: 'dict',
+                    data: {
+                      url: message.request.url,
+                      ...message.request.dataMap,
+                    },
+                  }),
+                );
+              }
             } else if (message.response != null) {
               process.stderr.write(
                 `Authenticated digital identity provider ${providerId}\n`,
--- a/src/identities/CommandAuthenticate.ts
+++ b/src/identities/CommandAuthenticate.ts
@@ -83,7 +83,9 @@ class CommandAuthenticate extends CommandPolykey {
                   type: options.format === 'json' ? 'json' : 'dict',
                   data: {
                     url: message.request.url,
-                    ...message.request.dataMap,
+                    ...(options.format === 'json'
+                      ? { dataMap: message.request.dataMap }
+                      : message.request.dataMap),
                   },
                 }),
               );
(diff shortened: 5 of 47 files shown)


CMCDragonkai:
This can be incrementally cherry-picked? Or will you do it one go?

CMCDragonkai:
Cherry pick relevant commits back to staging @amydevs

amydevs:
the first two commits have been cherrypicked into staging, remember to rebase after

CMCDragonkai:
You should rebase this PR if you've cherrypicked, since then this will be left with just 1 commit! Otherwise it will be confusing.

CMCDragonkai:
There's still a problem here: https://github.com/MatrixAI/Polykey-CLI/pull/124#issuecomment-1950969158

CMCDragonkai:
This needs rebase to be ready to go. I think this week @amydevs you're back on PK #40.

CMCDragonkai:
Can you expand the task list so we know how close we are finishing this?

CMCDragonkai:
@amydevs where's the counter?

amydevs:
> @amydevs where's the counter?

on the tagged issue

amydevs:
@tegefaulkes should be ready to review, each commit description will name the changes to each command.

src/secrets/CommandList.ts
```diff
@@ -46,13 +46,19 @@ class CommandList extends CommandPolykey {
           logger: this.logger.getChild(PolykeyClient.name),
         });
         const data = await binUtils.retryAuthentication(async (auth) => {
-          const data: Array<string> = [];
+          const data: Array<any> = [];
           const stream = await pkClient.rpcClient.methods.vaultsSecretsList({
             metadata: auth,
```
tegefaulkes: I *really* don't like using the `any` type, can we properly type this?

tegefaulkes (commented):
I skimmed through it, generally seems good.

src/secrets/CommandList.ts
amydevs: sure

amydevs: i've rebased a commit to change this

amydevs: i think `src/identities/CommandList.ts` might need the same change. Although, I'm a bit caught up on whether the `json` format should be returning the type `{ permissions: Array<...>, nodes: Array<...>, identities: Array<...> }` rather than `GestaltMessage` type. If so, this is a command that is aggregating data from js-rpc, and hence can decide to not follow the js-rpc response type. This would make sense if `identities list` is meant to list identities, rather than gestalts themselves. If not, I'll keep the non-json format output the same, and change the json format to output a similar structure to the rpc response.

tegefaulkes: Not sure, It should only keep the data it needs for for the command. The main thing I want is explicit typing. So no use of `any` type, implicitly or otherwise.

amydevs:
i think we should have a tree format for displaying certain nested structures, but then isn't that just yaml? Regardless, it should be discussed in another PR

CMCDragonkai:
The tree format is the dictionary format?

amydevs:
> The tree format is the dictionary format?

i was thinking that we should have a tree format that is dictionary but will add indentation for nested values. Currently i'm using `dict` and `list` formats to make up for it.

src/secrets/CommandList.ts
amydevs: I've made it so that that command uses explicit typing, and that the output of the command more closely follows the RPC response.

CMCDragonkai:
Yes the dict is supposed to be capable of nesting indentation! I just didn't get around to implementing it. Can you try?

CMCDragonkai:
Typically a tab is used for terminal indentation.

tegefaulkes:
@amy is this waiting on anything?

amydevs:
> @amy is this waiting on anything?

the `dict` format indentation now basically

CMCDragonkai:
> > @amy is this waiting on anything?
> 
> the `dict` format indentation now basically

Are you working on that?

amydevs:
> > > @amy is this waiting on anything?
> > 
> > 
> > the `dict` format indentation now basically
> 
> Are you working on that?

i haven't started working on it, but i'll get to it next. It kind of expands the scope of the PR a bit, but i think it's better to get it done and review it once than review the same sections of code multiple times.

CMCDragonkai:
Ok ETA on it? And then review later.

amydevs:
> Ok ETA on it? And then review later.

probably should be able to get it done by the end of the week, need to spec out API mainly

CMCDragonkai:
If so, we might separate that into a different PR then.

tegefaulkes (approved):
Looks fine to me.

CMCDragonkai:
What I meant by prefix per message is more like this:

```
Message: ...
Message: ...
```

Something like that.

But I'm talking in abstract. A better design is necessary.

```
v-sdf32434: ...
v-345345: ...
```

At the very least listing messages should tell us things like:

1. The notification ID
2. The sender ID
3. The time sent/date in UTC time - ISO8601 format
4. The message contents

Things like that. This is the sort of thing that I want @CryptoTotalWar and @faithveloro09 to be aware of. The product design side of PK - from CLI and GUI must be consistent.

Now I'm just spitballing but that could look like:

```
Notification: v234829759845
From: GestaltId-NodeId
Message: ...
Notification: v348395745
```

The `Notification: ...` is the "separator" here. Indicating the beginning of a new notification.

It's important to maintain only a single NL separation for simplified machine parsing.

Human readability on the CLI/terminal must also be amenable to simple text manipulation - Unix principles.

So that as a CLI - it is tasteful.

amydevs:
@CMCDragonkai
If we were to do something like

```
Notification: v234829759845
From: GestaltId-NodeId
Message: ...
Notification: v348395745
```

The value of the notification key would not be obvious as to what it actually is.

I've updated the PR with some changes that i've made regarding this. 

I have made it so that the dict format will have line separated array elements when dealing with an array nested within an object according to this [spec](https://github.com/MatrixAI/Polykey-CLI/issues/158) (this was so i did not have to mix the `list` and `dict` with `identities/CommandList`)

Taking inspiration from your proposal, this is the new output:

```
notificiation
  data  
    message test message 1
    type    General
  iat       1711063520.438
  isRead    true
  iss       vmqmqq9h4jgduaeifep1nko1j2uebgg0uh48sc4fl9upqdsijjmf0
  sub       voeo52qencdh720o084mlfpgb1st09vjca0ti9qvm6e0p19ma316g
  typ       notification
notificiation
  data  
    message test message 2
    type    General
  iat       1711063520.438
  isRead    true
  iss       vmqmqq9h4jgduaeifep1nko1j2uebgg0uh48sc4fl9upqdsijjmf0
  sub       voeo52qencdh720o084mlfpgb1st09vjca0ti9qvm6e0p19ma316g
  typ       notification
```

In this case, the `notification` line has two separate purposes. It acts as the separator for each notification. But also according to the [nested `dict` spec](https://github.com/MatrixAI/Polykey-CLI/issues/158), they represent an array of objects that have the single key `notification` in them. This makes it both more readable and be more similar to the RPC response for the purpose of machine parsability.

CMCDragonkai:
Can we have the notification ID?

amydevs:
> Can we have the notification ID?

as in the id they are indexed by in the db? cos i dont think they are exposed through the js-rpc response, it's currently purely internal @CMCDragonkai

CMCDragonkai:
They maybe in the future. I reckon it breaks up the monotony of just `Notification`.

CMCDragonkai:
```
notification id abc123
  ...
```

PR #79, opened 2023-12-13
Fixes the vault cloning issue when trying to clone a vault with the same name as one already present

### LINKED ISSUE #73: Vault Clone throws "undefined" if trying to clone a vault with the same name as one already existing
### Describe the bug

When trying to clone a vault from another node which has a name which is the name of a vault that already exists within the nodeID, Polykey throws an "undefined".

### To Reproduce

1. Have a vault named x in node A.
2. Have a vault named x in node B.
3. Share x from node B to node A.
4. Try to clone x from node A.

### Expected behavior

- It should still clone the vault but add a (1) at the end of the vault name.

addievo:
### Description

In CLI, an option has been added to override the default cloning behaviour. 

### Issues Fixed

* Fixes #73 

### Tasks

- [x] 1. Add override for cloning
- [x] 2. Pass the override in CloneVault

### Final checklist
* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/src/utils/options.ts
+++ b/src/utils/options.ts
@@ -153,6 +153,11 @@ const forceNodeAdd = new commander.Option(
   'Force adding node to nodeGraph',
 ).default(false);
 
+const overrideClone = new commander.Option(
+  '--override',
+  'Force clone node and append with a _1',
+).default(false);
+
 const noPing = new commander.Option('--no-ping', 'Skip ping step').default(
   true,
 );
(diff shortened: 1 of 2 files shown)


addievo:
Related https://github.com/MatrixAI/Polykey/issues/652

CMCDragonkai:
This is not intended, the vault name is supposed to be auto incremented if there's a conflicting name.

CMCDragonkai:
It should be a bug fix in PK.

PR #82, opened 2023-12-18
Integration testing refactor stage 1

### LINKED ISSUE #71: Integration Tests for `testnet.polykey.com`
### Specification
We need a suite of tests to cover interacting with a deployed agent, which we can do using `testnet.polykey.io`. These tests need to cover various different connection scenarios:

- Connecting to a deployed agent as a seed node during startup
- Pinging a deployed agent
- Using the deployed agent as a signaller (and eventually a relay once this is implemented)
- Tests for when the deployed agent is contacting an agent behind a NAT (https://github.com/MatrixAI/js-polykey/issues/159)
- Any bugs that are discovered during the above tasks

These tests should go into their own subdirectory `tests/testnet` and should not be run with the other tests. They should be disabled in our jest config and should only run when explicitly called (which will happen during the integration stage of our pipelines).

Required tests:
- `tests/testnet/testnetConnection.test.ts`
  - Can connect to the testnet
    - Within a reasonable amount of time
    - Without errors/shutting down the local agent
    - Without errors/shutting down the testnet
  - Can disconnect from the testnet
    - Within a reasonable amount of time
    - Without errors/shutting down the local agent
    - Without errors/shutting down the testnet
  - Can reconnect to the testnet
    - Able to handle different node ids (testnet is a cluster of nodes)
- `tests/testnet/testnetPing.test.ts`
  - Can ping the testnet
    - Able to handle different node ids (testnet is a cluster of nodes)
  - Can ping another node via the testnet (signaling)
  - Can ping another node via the testnet (relay)
  - Can attempt to ping another node that doesn't exist
    - Without shutting down the testnet
- `tests/testnet/testnetNAT.test.ts`
  - Can ping a node that is behind endpoint-independent NAT via the testnet
    - From a node that is not behind a NAT (DMZ)
    - From a node that is behind endpoint-independent NAT
    - From a node that is behind endpoint-dependent NAT
  - Can ping a node that is behind endpoint-dependent NAT via the testnet
    - From a node that is not behind a NAT (DMZ)
    - From a node that is behind endpoint-independent NAT
    - From a node that is behind endpoint-dependent NAT
- Should also incorporate tests from https://github.com/MatrixAI/Polykey/pull/326#issuecomment-1041190329

### Additional context
- https://github.com/MatrixAI/js-polykey/issues/159 - NAT traversal tests against `testnet.polykey.io`
- https://github.com/MatrixAI/js-polykey/issues/414 - Issue for failed connections to/from a deployed agent
- https://github.com/MatrixAI/js-polykey/issues/415 - Issue for continuous re-connection bug when connecting to deployed seed node
- https://github.com/MatrixAI/js-polykey/issues/413 - Issue for potential infinite loop when connecting to deployed seed node

### Tasks

- [x] 1. Attempt connections to the deployed seed node and create issues for all bugs discovered (and resolve them)
- [x] 2. Create tests for simple connections to `testnet.polykey.io`
  * 1 node connected to testnet.polykey.io and maintains connection
  * 2 nodes connected to testnet.polykey.io and can ping each other (they will have the same IP but different ports)
- ~[ ] 3. Create tests for edge cases and previous bugs~ - most edge cases will go to the simulation suite
- ~[ ] 4. Create tests for connecting to a deployed seed node from behind a NAT~ - this can only be done as part of a simulation suite, since we don't control host firewalls
- [x] 5. Finish off all diagrams as part of NAT testing MatrixAI/Polykey#388 
- [x] 6. Add new INFO level logs for situations where connections are going into stopping. This is next to the debug logs.
- ~[ ] 7. Add a logging debug filter to command line arguments to take a regular expression. Should be global option like `--log='/regex/'`.~ - not relevant to integration testing, the `js-logger` does support REGEX filtering, but the PK CLI currently doesn't have this option.
- ~[ ] 8. Use https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_ContainerOverride.html to be able to easily try different debugging levels and filters.~ - cannot use container overrides in services, can be done as part of tasks though, will have to just redeploy service with different task definition each time
- ~[ ] 9. `agent status` command needs to display useful information like the polykey version and other useful statistics like active connections, number of node graph entries etc etc.~
- ~[ ] 10. `--client-host` needs to support host names.~ - this is pending a change to being able to use `PolykeyClient` to connect to a host name - which would require using the DNS-SD SRV records. This still needs to be specced out how this would work because in some cases you want to connect to a SINGLE Node, in other cases you are "discovering" a node to connect to, but it's not relevant to this epic. 
- [x] 11. EC2 setup with idempotency
- [x] 12. Multi Node Setup on AWS
- [x] 13. Recovery Code Pool on AWS
- [x] 14. Multi-Host DNS resolution
- [x] 15. Multi Node Resolver
- [x] 16. NodeGraph KeyPath to lift Host and Port to the key path
- [x] 17. Put the trusted testnet seed nodes into the `src/config.ts` - from MatrixAI/Polykey#488 

### Emergent bugs

- [x] 1. Seed node is failing to establish a reverse connection back to the connecting node. The problem is somewhere in `ConnectionReverse.start()` between `Starting Connection Reverse` and `Started Connection Reverse`. Start by testing locally.
- [x] 2. Fixing the Random Process Termination
  * https://github.com/MatrixAI/Polykey-CLI/issues/71
  * https://github.com/MatrixAI/Polykey-CLI/issues/71
- [x] 3. Verify that new timeouts are being respected
  * https://github.com/MatrixAI/Polykey-CLI/issues/71

### LINKED ISSUE #81: Refactor testnet integration testing
### Specification

We need to refactor and streamline testnet integration testing. There are a few core goals of this.

1. Extract and simplify integration specific tests and move them to a separate test directory.
2. simplify the normal tests by removing the testif test switching and removing connections to the testnet.
3. update ci jobs to do integration testing after the seed nodes have updated with the new deployment.

For All of the `testif` usage in the standard tests need to be removed. The integration tests should not just be repeats of certain standard tests. The integration tests may still use the testif utility to prevent some tests on certain platforms. This would need to be looked into.

Integration tests need to be moved into a separate test directory to keep it isolated from the standard tests. This will allow us to easily run the standard and integration tests separately. The Integration tests should focus explicitly on two aspects.

1.  Running built and package executable to verify that they are working
2. Doing some basic checks while connecting to the `testnet`.

The ci jobs need to be updated as well. Currently we have a job that triggers the seed node tasks to restart with the latest image we deployed. This will be removed and handled by the infrastructure repo. This step should be replaced with a job that waits for the seed nodes to update before proceeding. We'll need an API endpoint on the polykey dashboard that exposes the seed node versions in some fashion to help with this. @tegefaulkes will need to work with @amydevs to spec this out and address it.

Waiting for the seed nodes to deploy and update is not *essential* here. Any breaking changes on the agent-agent RPC should be pretty rare.  For now I'm going to address it last as things should function without it for now. 

### Additional context

* Related epic #71 

### Tasks

* [x] 1. All the standard tests will no attempt connections to any network. They should explicitly be started with no seed nodes.
* [x] 2. The integration tests will be separated from the standard tests.
    * [x] 1. remove usage of the `testif` utility from the standard tests.
    * [x] 2. integration tests will be in a separate folder structure from the standard tests.
    * [x] 3. integration tests will focus on connecting to the `testnet`.
* [x] 3. integration tests need to wait for the `testnet` to be updated. to this end the following changes are to be made.
    * [ ] 1. CLI ci job for for triggering the `testnet` seed nodes using the new image will be removed. This will be handled by the `testnet` infrastructure.
    * [x] 2. A job will be created to wait for the seed nodes to switch to the new version. This will be done by polling a `polykey` dashboard endpoint that will either return the when all seednodes have been updated, or list the versions of all the seed nodes. This will need to be speced out.

tegefaulkes:
### Description

This PR addresses refactoring integration testing.

### Issues Fixed

* Related #81 
* Related #71 

### Tasks

* [x] 1. All the standard tests will no attempt connections to any network. They should explicitly be started with no seed nodes.
* [x] 2. The integration tests will be separated from the standard tests.
    * [x] 1. remove usage of the `testif` utility from the standard tests.
    * [x] 2. integration tests will be in a separate folder structure from the standard tests.
    * [x] 3. integration tests will focus on connecting to the `testnet`.
* ~3. integration tests need to wait for the `testnet` to be updated. to this end the following changes are to be made.~
    * ~1. CLI ci job for for triggering the `testnet` seed nodes using the new image will be removed. This will be handled by the `testnet` infrastructure.~
    * ~2. A job will be created to wait for the seed nodes to switch to the new version. This will be done by polling a `polykey` dashboard endpoint that will either return the when all seednodes have been updated, or list the versions of all the seed nodes. This will need to be speced out.~

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased


tegefaulkes:
I'm going to merge this with the changes for parts 1 and 2. Once merged I can fix any problems with the CI and move forward with fixing that. 

Part 3 will be addressed in a follow up PR.

PR #116, opened 2024-02-02
Fix overlay compatibility

brynblack:
### Description
This pull request fixes a problem to do with how the commitHash is calculated. Within the default.nix, GIT_DIR is evaluated at compile time with the path './.git', however in a build environment like in a nixpkgs overlay, this isn't taken into account and the path cannot be evaluated to the correct .git directory. This will always happen regardless of if commitHash was set or not. Therefore, the only way to fix this is to actually remove the GIT_DIR line from default.nix, and instead rely purely on hard coded file paths within build.js. This successfuly allows Polykey-CLI to be integrated with a nixpkgs overlay, and can now be deployed as such.

### Tasks
- [x] 1. Remove dependency on GIT_DIR variable
- [x] 2. Deploy to downstream overlay
- [x] 3. Test to see if Polykey-CLI is able to be built through an overlay on Nix
- [x] 4. Update CI to use the commit hash parameter

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/release.nix
+++ b/release.nix
@@ -1,5 +1,6 @@
 { npmDepsHash ? ""
 , commitHash ? null
+, gitDir ? null
 , pkgs ? import ./pkgs.nix {}
 }:
 
@@ -18,7 +19,7 @@ let
         src = utils.src;
         PKG_CACHE_PATH = utils.pkgCachePath;
         PKG_IGNORE_TAG = 1;
-        GIT_DIR = utils.dotGit;
+        GIT_DIR = gitDir;
         postBuild = ''
           npm run pkg -- \
             --output=out \
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -17,7 +17,7 @@ async function main(argv = process.argv) {
   const projectRoot = path.join(__dirname, '..');
   const buildPath = path.join(projectRoot, 'build');
   const distPath = path.join(projectRoot, 'dist');
-  const gitPath = process.env.GIT_DIR ?? path.join(projectRoot, '.git');
+
   await fs.promises.rm(distPath, {
     recursive: true,
     force: true,
@@ -31,9 +31,11 @@ async function main(argv = process.argv) {
     encoding: 'utf-8',
     shell: platform === 'win32' ? true : false,
   });
+
   // This collects the build metadata and adds it to the build folder so that dynamic imports to it will resolve correctly.
   let gitHead = process.env.COMMIT_HASH;
   if (gitHead == null) {
+    const gitPath = process.env.GIT_DIR ?? path.join(projectRoot, '.git');
     gitHead = await fs.promises.readFile(path.join(gitPath, 'HEAD'), 'utf-8');
     if (gitHead.startsWith('ref: ')) {
       const refPath = gitHead.slice(5).trim();
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -69,7 +69,7 @@ check:nix-dry:
     - >
       nix-shell --arg ci true --run $'
       npmDepsHash="$(prefetch-npm-deps ./package-lock.json)";
-      nix-build -v -v --dry-run ./release.nix --argstr npmDepsHash "$npmDepsHash";
+      nix-build -v -v --dry-run ./release.nix --argstr npmDepsHash "$npmDepsHash" --argstr commitHash "$CI_COMMIT_SHA";
       '
   rules:
     # Runs on feature and staging commits and ignores version commits
--- a/default.nix
+++ b/default.nix
@@ -1,5 +1,6 @@
 { npmDepsHash ? ""
 , commitHash ? null
+, gitDir ? null
 , callPackage
 , buildNpmPackage
 }:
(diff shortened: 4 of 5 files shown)


scripts/build.js
```diff
@@ -17,7 +17,12 @@ async function main(argv = process.argv) {
   const projectRoot = path.join(__dirname, '..');
   const buildPath = path.join(projectRoot, 'build');
   const distPath = path.join(projectRoot, 'dist');
-  const gitPath = process.env.GIT_DIR ?? path.join(projectRoot, '.git');
+
+  var gitPath = null;
+  if (process.env.COMMIT_HASH == null) {
+    gitPath = path.join(projectRoot, '.git');
+  }
+
```
tegefaulkes: 1. Try to keep this close to where it's used.
2. We generally void using `var`, Always use `const` if you don't intend for the value to change since it will protect you from mistakes generally. use let only if you need to re-assign. `var` works like `let` but hoists the variable to a level above the current scope. 
3. I think the logic flow is a little weird now. Try to simplify it. The main problem is we're being a little implicit. Where the logic is `if a set b; if b set c; if c....` If it's spread out like that it becomes hard to follow.

tegefaulkes (changes requested):

scripts/build.js
tegefaulkes: I'd say still have the ability to specify the gitPath like `const gitPath = process.env.GIT_DIR ?? path.join(projectRoot, '.git');` but otherwise seems fine.

Also, reply to review comments if you've addressed them and just resolve them if You feel they've been suitably addressed.

brynblack: That seems like a good idea, given `GIT_DIR` could be used as a manual override. We can just add it as a parameter to `default.nix` and `release.nix`, and set it to some default empty value. That would allow for the user to override the `.git` directory, although I'm unsure whether it will resolve to an appropriate path in the Nix store.

brynblack: I think its good to go now, its got a manual override for GIT_DIR that can be set through changing the `nix-build` parameter.

tegefaulkes (approved):

release.nix
```diff
@@ -18,7 +19,7 @@ let
         src = utils.src;
         PKG_CACHE_PATH = utils.pkgCachePath;
         PKG_IGNORE_TAG = 1;
-        GIT_DIR = utils.dotGit;
+        GIT_DIR = gitDir;
```
CMCDragonkai, line 22: I'm confused about this. Originally @amydevs needed to be able to auto resolve the commit hash.

This was the solution she had.

Now what's the solution? We still need to auto resolve the commit hash if we are doing a nix-build on this repo.

The paramaterisation only applies during the flake state.

tegefaulkes: Running npm run build works, but running nix-build is failing with the following.

```

> polykey-cli@0.1.4 build
> node ./scripts/build.js

Running tsc:
tsc -p ./tsconfig.build.json
node:internal/process/promises:288
            triggerUncaughtException(err, true /* fromPromise */);
            ^

[Error: ENOENT: no such file or directory, open '/build/Polykey-CLI/.git/HEAD'] {
  errno: -2,
  code: 'ENOENT',
  syscall: 'open',
  path: '/build/Polykey-CLI/.git/HEAD'
}

Node.js v18.17.1

ERROR: `npm build` failed

Here are a few things you can try, depending on the error:
1. Make sure your build script (build) exists
  If there is none, set `dontNpmBuild = true`.
2. If the error being thrown is something similar to "error:0308010C:digital envelope routines::unsupported", add `NODE_OPTIONS = "--openssl-legacy-provider"` to your derivation
  See https://github.com/webpack/webpack/issues/14532 for more information.

error: builder for '/nix/store/ml0rvfm81va2r7388marjvxkwkik75p4-polykey-cli-0.1.4-linux-x64.drv' failed with exit code 1;
       last 10 log lines:
       > Node.js v18.17.1
       >
       > ERROR: `npm build` failed
       >
       > Here are a few things you can try, depending on the error:
       > 1. Make sure your build script (build) exists
       >   If there is none, set `dontNpmBuild = true`.
       > 2. If the error being thrown is something similar to "error:0308010C:digital envelope routines::unsupported", add `NODE_OPTIONS = "--openssl-legacy-provider"` to your derivation
       >   See https://github.com/webpack/webpack/issues/14532 for more information.
       >
       For full logs, run 'nix log /nix/store/ml0rvfm81va2r7388marjvxkwkik75p4-polykey-cli-0.1.4-linux-x64.drv'.

```

I'm going to revert the merge until this is resolved.

tegefaulkes: Actually, it's not breaking breaking things right now, so we can leave it until it's fixed. CI still passes AFAIK.

CMCDragonkai: Generally speaking you want to minimise any env variables that is needed. And all new env variables must be documented in the `.env.example`.

But also I don't think this makes sense atm. There should be no need for it.

Also the thing so rolled back.

PR #120, opened 2024-02-05
feat: added flake.nix and overlay fix

brynblack:
### Description

This pull request adds flake support to Polykey-CLI, alongside the traditional nix-build system that is used. It also adds in a fix to allow overlays that use the fetchGit method to correctly operate given the appropriate parameters.

### Tasks

- [x] 1. Configure flake
- [x] 2. Fix overlay deployment
- [x] 3. Test through nix-build, overlay, flake build, flake overlay, and CI
- [x] 4. Figure out how to get the npmDepsHash into the flake
- [x] 5. Discuss whether we should keep or remove the old deployment system

### Final checklist

* [ ] Domain specific tests
* [ ] Full tests
* [ ] Lint fixed
* [ ] Squash and rebased


brynblack:
I've created an excalidraw explaining the current problem at hand. This is purely due to the way flakes are sandboxed.

[https://excalidraw.com/#room=f3c40e8b68fda3031df7,xgOz17bgFl1l--AUgT4zaA](https://excalidraw.com/#room=f3c40e8b68fda3031df7,xgOz17bgFl1l--AUgT4zaA)

brynblack:
Also some discussion needs to be had on keeping the original deployment methods available. If we are switching to flakes, and they replace most of the files we were originally using, why bother keeping them? I want to figure out an answer.

I haven't pushed up this code yet from my laptop, but I have found a way to combine both the release and default deployments for Polykey-CLI into one flake.nix, and you can do some really cool stuff with the new nix commands related to building and development. When I can figure out the npmDepsHash problem, I can get this code pushed up.

CMCDragonkai:
Always reference the new PR when closing in favour of, also applies to issues.

Issue #111, opened 2024-01-30
Update to Node20 and switch `pkg`


tegefaulkes:
### Specification

There are two parts to this issue. 
1. Once again updating to node 20 as it's supported by the packager now.
2. Switch the packager from `vercel/pkg` to https://github.com/yao-pkg/pkg

### Additional context

Related https://github.com/MatrixAI/Polykey-CLI/issues/40#issuecomment-1913946054

### Tasks

1. Update to node 20
2. switch to using `yao-pkg/pkg` for packaging. This includes removing the use of `vercel/pkg`.

CMCDragonkai:
When switching to node 20 and the new yao pkg, enable the arm64 builds. It's now available. See: https://github.com/yao-pkg/pkg-fetch/releases

Will need to update `utils.nix` and `.gitlab-ci.yml` to produce those builds. Also this infrastructure should be moved to `nixpkgs-matrix` so we can use it across multiple projects.

CMCDragonkai:
PK is already on Node 20. This should be part of the #40.

CMCDragonkai:
Node 20 is LTS since october last year.

amydevs:
This seems to have already been done.

> enable the arm64 builds

We will need to make arm64 builds for the dependencies before this can happen (js-db, js-mdns, etc.)

CMCDragonkai:
This is done? Where's the PR.

amydevs:
pretty sure the switch to `yao-pkg` was done in 2ef4aa966fd87b9e2902712da14d0bffa6c79bd8

amydevs:
ah wait, the environment scripts and the flake still refer to node 18, will switch it over

amydevs:
https://github.com/yao-pkg/pkg#targets

It seems that it is not possible to generate arm64 macos binaries on any systems other than an arm64 macos machine.

This is the error when i try running it on my nixos x64 linux laptop:

```bash
pkg dist/polykey.js --config=/build/3x0x90lzbmrynhqyxs0fj5msdqiidd93-source/pkg.json --targets=node20-macos-arm64 --no-bytecode --no-native-build --public --public-packages='*>
> pkg@5.11.4
> Fetching base Node.js binaries to PKG_CACHE_PATH
> Fetching base Node.js binaries to PKG_CACHE_PATH
  fetched-v20.11.1-macos-arm64        [] 0%> Not found in remote cache:
[...]

PR #124, opened 2024-02-07
Flakes Implementation

brynblack:
### Description

This pull request adds flake support to Polykey-CLI, replacing the traditional "legacy" build system of Nix. It simplifies a lot of the Nix build codebase, and increases reproducability and reliability.

### Tasks

- [x] 1. Configure flake
- [x] 2. Fix overlay deployment
- [x] 3. Test through nix-build, overlay, flake build, flake overlay, and CI
- [x] 4. Figure out how to get the npmDepsHash into the flake
- [x] 5. Discuss whether we should keep or remove the old deployment system
- [x] 6. Implement remaining minor changes
- [x] 7. Fix CI dry run
- [x] 8. Re-add release.nix style package output (vercel/pkg, binary output format should be the same)
- [x] 9. Minor revisions

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased


.gitlab-ci.yml
```diff
@@ -49,7 +49,7 @@ check:lint:
   needs: []
   script:
     - >
-      nix-shell --arg ci true --run $'
+      nix develop .#ci --command bash -c $'
```
CMCDragonkai, line 52: How does this work? Does it use the `bash` from nixpkgs upstream? Can you specify `zsh` too? And is there no "default shell"?

.gitlab-ci.yml
```diff
@@ -233,26 +232,20 @@ integration:builds:
...
-        --attr docker \
-        --attr package.linux.x64.elf \
-        --attr package.windows.x64.exe \
-        --attr package.macos.x64.macho \
-        --argstr npmDepsHash "$npmDepsHash" \
-        --argstr commitHash "$CI_COMMIT_SHA")";
+        .#docker \
+        .#packages.x86_64-linux \
+        .#packages.x86_64-windows \
+        .#packages.x86_64-darwin;
```
CMCDragonkai: I want to redo the naming here... `docker`, `linux.x64.elf`. Why change the names here?

README.md
```diff
@@ -46,48 +46,72 @@ Have a security issue you want to let us know? You can contact us on our website
...
+nix build .#packages.x86_64-linux.polykey-cli
+nix build .#packages.x86_64-windows.polykey-cli
+nix build .#packages.x86_64-darwin.polykey-cli
 ```
 
 Install into Nix user profile:
 
 ```sh
-nix-env -f ./release.nix --install --attr application --argstr npmDepsHash "$(prefetch-npm-deps ./package-lock.json)"
+nix profile install github:MatrixAI/Polykey-CLI
```
CMCDragonkai, line 99: How does this work? It understands `github:`?

default.nix
CMCDragonkai, line 1: I believe you still need this for compat with `nixpkgs`. I feel like you cannot just not have this? Can your flakes import this?

npmDepsHash.nix
CMCDragonkai, line 1: I think a raw file is fine: `"sha256-bfDJ4AVsfSlEKpCNNTVBa0YrMQkafVe2/uZw37wbNXA="` is a valid nix expression.

flake.nix
```diff
@@ -0,0 +1,86 @@
...
+        };
+
+        packages = {
+          polykey-cli = polykey_cli;
+          docker = dockerImage;
+        };
+
+        devShells = {
+          default = import ./shell.nix { inherit pkgs; };
+          ci = import ./shell.nix { inherit pkgs; ci = true; };
```
CMCDragonkai: What's the point of having a `shell.nix` anymore then?

default.nix
CMCDragonkai: I'd argue that `default.nix` is sort of more necessary than `shell.nix`.

flake.nix
brynblack: `shell.nix` can be kept separate from the `flake.nix` to modularise it more. We are not just limited to keeping everything in one file, I felt the file was too large and just kept the `shell.nix` outside of the file.

Also makes more sense when specifying additional shells with different parameters, such as the `ci` shell, instead of redeclaring the entire shell again you can just import that `shell.nix`.

npmDepsHash.nix
brynblack: You're right, this can probably be changed. Its usually easier to work with `import`, but `builtins.readFile` should work fine here too.

default.nix
brynblack: Yes, flakes can import this. I will be updating the flakes PR in the public overlay with the way that this would be put into practice.

.gitlab-ci.yml
brynblack: The names are changed here, because when the flake builds for each of those platforms, it will output the relevant binary with the appropriate file extension for the platform.

.gitlab-ci.yml
brynblack: When running the `--command` argument, I was getting errors related to the syntax of the $'...'.

I'm not too sure why, but I read in the manpage for `nix develop` that they would use `bash -c` or a given shell to execute the command. I still don't know why, I have a feeling that the default `--command` argument has a bundled shell that does not have the same syntax as the shell code that was already written in the CI file.

I will look more into this, maybe we can make this more generic.

README.md
brynblack: In flakes, `github` and `gitlab` or whatever else are resolved to `https://INSERT_SITE`. This is purely a convenience feature of flakes. Read up on it here: [https://nixos.org/manual/nix/stable/command-ref/new-cli/nix3-flake.html#url-like-syntax](https://nixos.org/manual/nix/stable/command-ref/new-cli/nix3-flake.html#url-like-syntax)

CMCDragonkai:
There's some conflicts with `staging`.

But some notes:

1. So we are getting rid of `default.nix` and `shell.nix` and fully moving to `nix build` and `nix develop`.
2. The `nixpkgs-matrix` can refer to our flakes using just `builtins.getFlake` this is a new primitive that supports protocol URIs like `github:...` or `flakes:...`
3. If `flakes:...` is more generic, change to using this.

The nixpkgs upstream may move to accommodate more out-of-tree packaging for scalability reasons - the future is decentralized.

CMCDragonkai:
Also everything in our `utils.nix` should be factored out into `nixpkgs-matrix`! So it can be shared between all of our projects!

.gitlab-ci.yml
CMCDragonkai: We don't want the file extension, none of the binaries should have a file extension.

npmDepsHash.nix
CMCDragonkai: Actually `import` should work on `"..."` as that is a valid Nix expression.

However `builtins.readFile` is actually more appropriate. So then you just use `...` without the double quotes.

flake.nix
CMCDragonkai: I think you can just use a function and have it all in one `flakes.nix`. It would be strange to not have a `default.nix` and just one `shell.nix`, as it would imply the usage of `nix-shell`, whereas we want to use `nix develop` now.

CMCDragonkai: One file should be fine for all of it! Just use `let` declarations to factor and abstract.

brynblack:
This CI dry-run issue is very annoying. I don't understand why the runner doesn't have access to the `tmp` directory, given its being manually set to the correct permissions within the `flake.nix`. I saw some discussion about it, but there doesn't seem to be one clear solution.

CMCDragonkai:
> This CI dry-run issue is very annoying. I don't understand why the runner doesn't have access to the `tmp` directory, given its being manually set to the correct permissions within the `flake.nix`. I saw some discussion about it, but there doesn't seem to be one clear solution.

Check if tmp actually exists. Run `file /tmp` or `ls /`. Run them both beforehand or stat.

.gitlab-ci.yml
```diff
@@ -67,9 +67,8 @@ check:nix-dry:
   needs: []
   script:
     - >
-      nix-shell --arg ci true --run $'
-      npmDepsHash="$(prefetch-npm-deps ./package-lock.json)";
-      nix-build -v -v --dry-run ./release.nix --argstr npmDepsHash "$npmDepsHash";
+      nix develop .#ci --command bash -c $'
+      ls -l /tmp
       '
```
CMCDragonkai: Oh you are not using `nix-shell`.

Ok well like I said, you can do these 2:

```
check:nix-dry:
  stage: check
  needs: []
  script:
    - stat /tmp
    - ls -l /tmp
```

And

```
check:nix-dry:
  stage: check
  needs: []
  script:
    - >
      nix develop .#ci --command bash -c $'
      stat /tmp;
      ls -l /tmp;
      '
```

Also can you try not using `--command bash` I think this could be adding something weird into the environment...

CMCDragonkai: image

CMCDragonkai: Use `nix run` instead?

CMCDragonkai: Oh and:

> --impure: Allows accessing non-Nix-provided dependencies and resources from the environment, bypassing the typical purity guarantees of Nix. Useful when the development environment needs access to system libraries or tools not specified in the flake.

You can use this for `nix develop`.

CMCDragonkai: And apparently you can do `nix develop --command 'echo Hello';`. It just runs that within a script... Try this too.

brynblack:
<details><summary>Details</summary>
<p>

docker run -it --entrypoint /bin/bash c4d61a3b707b
bash-5.2# git clone https://github.com/MatrixAI/Polykey-CLI
Cloning into 'Polykey-CLI'...
remote: Enumerating objects: 2440, done.
remote: Counting objects: 100% (1066/1066), done.
remote: Compressing objects: 100% (316/316), done.
remote: Total 2440 (delta 797), reused 930 (delta 724), pack-reused 1374
Receiving objects: 100% (2440/2440), 2.34 MiB | 6.09 MiB/s, done.
Resolving deltas: 100% (1815/1815), done.
bash-5.2# cd /Polykey-CLI
bash-5.2# git checkout feature-flakes
branch 'feature-flakes' set up to track 'origin/feature-flakes'.
Switched to a new branch 'feature-flakes'
bash-5.2# nix build .#docker
error:
       … while calling the 'derivationStrict' builtin

         at /builtin/derivation.nix:9:12: (source not available)

       … while evaluating derivation 'docker-image-polykey-cli-0.1.4.tar.gz'
         whose name attribute is located at /nix/store/f6zwsqxlr744fg8mmgdlq3qi3r1nlk3y-source/pkgs/stdenv/generic/make-derivation.nix:303:7

       … while evaluating attribute 'baseJson' of derivation 'docker-image-polykey-cli-0.1.4.tar.gz'

         at /nix/store/f6zwsqxlr744fg8mmgdlq3qi3r1nlk3y-source/pkgs/build-support/docker/default.nix:598:18:

          597|           imageTag = if tag == null then "" else tag;
          598|           inherit fromImage baseJson;
             |                  ^
          599|           layerClosure = writeReferencesToFile layer;

       (stack trace truncated; use '--show-trace' to show the full trace)

       error: program '/etc/nix/upload-to-cache.sh' failed with exit code 1
bash-5.2# echo "" > /etc/nix/upload-to-cache.sh
bash-5.2# nix build .#docker
bash-5.2# nix develop 
Entering "polykey-cli"
bash: ./.env: No such file or directory

mkdir --parents "$(pwd)/tmp"

export PATH="$(pwd)/dist/bin:$(npm root)/.bin:$PATH"

npm install --ignore-scripts
npm WARN EBADENGINE Unsupported engine {
npm WARN EBADENGINE   package: '@matrixai/db@5.3.0',
npm WARN EBADENGINE   required: { msvs: '2019', node: '^20.5.1' },
npm WARN EBADENGINE   current: { node: 'v18.17.1', npm: '9.6.7' }
npm WARN EBADENGINE }
npm WARN EBADENGINE Unsupported engine {
npm WARN EBADENGINE   package: '@matrixai/mdns@1.2.5',
npm WARN EBADENGINE   required: { msvs: '2019', node: '^20.5.1' },
npm WARN EBADENGINE   current: { node: 'v18.17.1', npm: '9.6.7' }
npm WARN EBADENGINE }

up to date, audited 741 packages in 680ms

168 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities

set +v
bash-5.2# nix build .#docker
bash-5.2#

</p>
</details>

.gitlab-ci.yml
brynblack: `nix run` is different, it will run an `app` defined inside of the flake. So in this scenario, you could do something like `nix run .#polykey-cli` and it would build and run Polykey-CLI, without you needing to go to the build outputs and run it yourself. You cannot just do `nix run 'nix develop ...'` or whatnot, its only for app outputs explicitly defined in the `flake.nix`.

The definition in code:
```
        apps = {
          polykey-cli = {
            type = "app";
            program = "${self.packages.${system}.polykey-cli}/bin/polykey";
          };
        };
```

CMCDragonkai:
Check the internals of upload to cache. It is supposed to upload to S3, nothing to do with nix I think.

brynblack:
> Check the internals of upload to cache. It is supposed to upload to S3, nothing to do with nix I think.

You're right, it had nothing to do with the s3 cache. Funnily enough the problem was right in front of my eyes the whole time. It seems wrapping a `nix build ...` inside of a `nix develop ...` does not seem to maintain the appropriate file permissions, and as such it cannot access directories such as `/tmp` and so on. Moving forwards we should keep this in mind when exercising CI usage.

CMCDragonkai:
> > Check the internals of upload to cache. It is supposed to upload to S3, nothing to do with nix I think.
> 
> You're right, it had nothing to do with the s3 cache. Funnily enough the problem was right in front of my eyes the whole time. It seems wrapping a `nix build ...` inside of a `nix develop ...` does not seem to maintain the appropriate file permissions, and as such it cannot access directories such as `/tmp` and so on. Moving forwards we should keep this in mind when exercising CI usage.

Does it work fine when running `nix-build` inside `nix-shell`?

Like what exactly is `nix develop` doing? You should explain a bit more clearly, since I can always run `nix-build` inside `nix-shell`, and I often do.

brynblack:
> Does it work fine when running `nix-build` inside `nix-shell`?
> 
> Like what exactly is `nix develop` doing? You should explain a bit more clearly, since I can always run `nix-build` inside `nix-shell`, and I often do.

Okay so to clarify, when `nix develop` is ran it starts a bash shell that provides an interactive build environment that is nearly identical to what Nix would typically use to build. Inside this shell, environment variables and shell functions are set up so that package building can be done interactively and incrementally. This is according to the [Nix wiki](https://nixos.org/manual/nix/stable/command-ref/new-cli/nix3-develop.html) itself.

Now in the case of Polykey-CLI, the default CI shell that runs installs a few dependencies such as Node.js and Skopeo. What is happening here when `nix build` is ran inside of `nix develop`, its creating a bash shell derivation that has access to the given inputs/environment variables. My theory here is that inside of the nix flake shell, its got restricted access to the certain system paths, and this makes sense given that the `/tmp` directory was inaccessible inside of the CI. I don't have a clear picture though, if you want a specific answer I will have to do more research into this topic.

Even if this is not the case, it is completely unnecessary to run `nix build` inside of a dev shell given that the flake build system in itself already handles the process of sandboxing everything. If you have the `nix` command available, you do not need it to use the nix command again to build. It will just work as intended. The flake build system can't even receive environment variables or other external mutability, so its redundant and this sort of wrapping should be avoided if the intention was to provide extra environment modifications.

CMCDragonkai:
What about during interactive development on local computers. Are you saying using `nix develop` means I cannot `ls /tmp` or create files under `/tmp`...?

brynblack:
> What about during interactive development on local computers. Are you saying using `nix develop` means I cannot `ls /tmp` or create files under `/tmp`...?

I don't think that's the case now that I think about it, my theory only makes sense in the domain of the custom shell that was used in the flake itself, but even then running `nix develop` inside Polykey-CLI locally and doing an `ls /tmp` works.

When I was looking through the source code for Polykey-CLI, I noticed that `pkg.js` was using a `find` function to locate the project files in the directory, and given that the `tmp` directory is sourced from the root that `npm ls ...` is run at, then possibly the shell environment is redirecting the true `npm` directory, leading to a failure to find the files (this is the exact part that was failing in the CI before I just applied this new fix). Maybe its the `pkg.js` itself. As I said, I don't have a clear picture and I need to look into this more.

CMCDragonkai:
Can you also address all remaining comments?

CMCDragonkai:
They are currently unresolved.

.npmrc
tegefaulkes, line 1: What did this fix again? the comment here says `# Enables npm link` which is something we definitely use for debugging in other repos. So if possible can we keep this?

README.md
```diff
@@ -46,48 +46,68 @@ Have a security issue you want to let us know? You can contact us on our website
...
-nix-build ./release.nix --attr docker --argstr npmDepsHash "$(prefetch-npm-deps ./package-lock.json)"
-nix-build ./release.nix --attr package.linux.x64.elf --argstr npmDepsHash "$(prefetch-npm-deps ./package-lock.json)"
-nix-build ./release.nix --attr package.windows.x64.exe --argstr npmDepsHash "$(prefetch-npm-deps ./package-lock.json)"
-nix-build ./release.nix --attr package.macos.x64.macho --argstr npmDepsHash "$(prefetch-npm-deps ./package-lock.json)"
+nix build .#polykey-cli
+nix build .#docker
+
+nix build .#packages.x86_64-linux.polykey-cli
+nix build .#packages.x86_64-windows.polykey-cli
+nix build .#packages.x86_64-darwin.polykey-cli
```
tegefaulkes: Why is there an empty line here? If you're groping them in some fashion then use a comment `# some comment` to explain it in the example. Otherwise I don't see a point to the line.

README.md
```diff
@@ -46,48 +46,68 @@ Have a security issue you want to let us know? You can contact us on our website
...
 ### Docker
 
 Install into Docker:
 
 ```sh
-loaded="$(docker load --input "$(nix-build ./release.nix --attr docker --argstr npmDepsHash "$(prefetch-npm-deps ./package-lock.json)")")"
-image="$(cut -d' ' -f3 <<< "$loaded")"
-docker run -it "$image"
+nix build .#docker
+docker load < result | cut -d' ' -f3 | xargs docker run -i
```
tegefaulkes: Is this just an example of loading it now? Or does this run it as well? It's a little confusing so keep the extra line were you run it with `docker run -it` in the example.

README.md
```diff
@@ -46,48 +46,68 @@ Have a security issue you want to let us know? You can contact us on our website
...
 ```
 
 ## Development
 
-Run `nix-shell`, and once you're inside, you can use:
+Run `nix develop`, and once you're inside, you can use:
```
tegefaulkes, line 120: Does this enter a dev shell? or just install polykey for you? I though you'd have to run `nix develop .#dev` for the dev shell?

flake.nix
tegefaulkes, line 1: I've brought before that the output for the native builds needs to be packaged using `vercel/pkg`. To my untrained eyes I can't tell how that is being done here. Can you explain how that step was integrated into the flakes building process?

scripts/npmDepsHash.sh
tegefaulkes, line 1: From what I recall of our discussions, this is meant to be triggered by the `dependencies` hook in the `package.json` scripts? I don't see any changes to `package.json`.

tegefaulkes (commented):
I think you forgot about one or two things we discussed before.

Before this is ready we need to sanity check that the outputs with the flake build methods produces the same thing as the old `release.nix` method. Mainly the following needs to be confirmed and sanity checked.

1. [x] The win, linux and mac output are a single executable file that doesn't depend on node being present to run. So make sure the results aren't just directories but a single binary file.
2. [x] The win, linux and mac outputs run on their relative platforms without issues. Check we get help output as a minimum. Ideally check that an agent runs and you can stat it with `pk agent start ...` and `pk agent status ...`.
3. [x] Make sure you can run the application in nix. I think this is already shown to work using the flake system but might as well sanity check it before merging.
4. [x] Make sure the docker image works for the same reasons.

README.md
brynblack: `nix develop` does not install Polykey-CLI, it enters the development environment that contains `npm` and related commands which are used for development purposes.

scripts/npmDepsHash.sh
brynblack: This is now done.

README.md
brynblack: I have made it more verbose.

README.md
brynblack: You're right, it should've been made more explicit in the usage. I have added inline comments.

.npmrc
brynblack: I re-added this now, I realised that it was not the cause of the CI issues.

.gitlab-ci.yml
brynblack: It seems that `nix develop` as of my earlier research by default uses bash anyway, so it does not really change the outcome here. It just means its possible to specify a multiline command.

flake.nix
brynblack: This needs to be done. Will do this ASAP.

.npmrc
CMCDragonkai: Npm linking is important for debugging.

flake.nix
CMCDragonkai: Make sure vercel/pkg builds are working, infact there's an issue to update to node 20 too #111 ... can that be combined right now since we are doing an infra update? Ignore if there are code app updates that are too extensive.

CMCDragonkai:
Tick off all the things in @tegefaulkes comment if they are addressed, and always explain how it was addressed.

flake.nix
brynblack: Okay, this has been resolved as of commit a9ba3b2 and a7a1351.

This was done by mapping the target build system to the vercel/pkg format. So if the target system was `x86_64-windows`, it would map to `win32-x64` and so on.

brynblack:
There is one thing I need to discuss, as of the two commits I have just published which mapped the target build system to the `vercel/pkg` format, there now exists a dependency on running the Nix build process on a x86_64-linux system. There exists a way to make this generic using `builtins.currentSystem`, however by default flakes restrict this due to it introducing mutability and reducing reproducibility. See [https://github.com/NixOS/nix/issues/3843](https://github.com/NixOS/nix/issues/3843).

.gitlab-ci.yml
brynblack: The binaries do not have a file extension as of commit a9ba3b2 and a7a1351.

brynblack:
Another problem is that the derivation that is outputted from the build process is not runnable as an ELF by Nix. I don't know the best way to deal with this, but before I migrated to using `vercel/pkg` for the build process in the flake it was able to run (it was installed as source and then ran with Node).

We should find out a better way of doing this then, instead of using something like `steam-run`.

CMCDragonkai:
No it's fine to use x86-64 to build all the binaries. It's not a cross-compiling system, it's just doing a simple bundle of a precompiled binary. Dealing with that is a separate issue.

Keep expectations of x86-64 to build all binaries, and don't worry about the `steam-run`, that's a NixOS specific issue, again not relevant.

Other than that, what's keeping this from being merged?

CMCDragonkai:
> The win, linux and mac outputs run on their relative platforms without issues. Check we get help output as a minimum. Ideally check that an agent runs and you can stat it with pk agent start ... and pk agent status ....

The CI should prove that this is working. Monitor the help outputs at the very least, since Epsilon DC doesn't have windows and mac turned on at this point.

CMCDragonkai:
The nix-runnable binary does not exist, it's supposed to just use the `application` build, which I think you've renamed as just the `default` build... right?

README.md
```diff
@@ -46,48 +46,71 @@ Have a security issue you want to let us know? You can contact us on our website
...
-nix-build ./release.nix --attr package.windows.x64.exe --argstr npmDepsHash "$(prefetch-npm-deps ./package-lock.json)"
-nix-build ./release.nix --attr package.macos.x64.macho --argstr npmDepsHash "$(prefetch-npm-deps ./package-lock.json)"
+# Build the packages for your current platform
+nix build .#polykey-cli
+nix build .#docker
+
+# You can also specify the platform to build for
+nix build .#packages.x86_64-linux.polykey-cli
+nix build .#packages.x86_64-windows.polykey-cli
+nix build .#packages.x86_64-darwin.polykey-cli
```
CMCDragonkai, line 93: Why change the paths?

```
package.linux.x64.elf
```

Can we just maintain this path for now.

CMCDragonkai: Maintain these paths if there's no reason to change the path naming.

flake.nix
```diff
@@ -0,0 +1,149 @@
...
+      let
+        # This needs to be made dynamic using something like builtins.currentSystem
+        buildSystem = "x86_64-linux";
+
+        systems = {
+          "x86_64-linux" = [ "linux" "x64" ];
+          "x86_64-windows" = [ "win32" "x64" ];
+          "x86_64-darwin" = [ "darwin" "x64" ];
+          "aarch64-darwin" = [ "darwin" "arm64" ];
+        };
```
CMCDragonkai: For now it does not make sense to have separate build systems, we are not doing any kind of cross compilation, can you simplify this for now.

Cross compilation is entirely separate problem and should not be solved here.

PK CLI is currently a JS codebase, which means it is expected to run on Nodejs. Cross compilation of node is an entirely separate problem to be put into another project.

flake.nix
```diff
@@ -0,0 +1,149 @@
...
+              "$packageOut"/scripts \
+              "$packageOut"/tsconfig.build.json \
+              "$packageOut"/tsconfig.json \
+              "$packageOut"/LICENSE \
+              "$packageOut"/ADDITIONAL_TERMS \
+              "$packageOut"/README.md;
+          '';
+        };
+
+        polykey_cli-release = pkgs.buildNpmPackage {
```
CMCDragonkai: Do not use `x_a-b` style naming, too complicated and too prone to error. We only ever use `a-b` naming for final attributes in nix.

package.json
```diff
@@ -91,7 +91,8 @@
     "pkg": "node ./scripts/pkg.js",
     "polykey": "ts-node src/polykey.ts",
     "start": "ts-node src/polykey.ts -- agent start --verbose",
-    "dev": "nodemon src/polykey.ts -- agent start --verbose"
+    "dev": "nodemon src/polykey.ts -- agent start --verbose",
+    "postshrinkwrap": "./scripts/npmDepsHash.sh"
```
CMCDragonkai: I thought there was a different script for this? I don't remember `postshrinkwrap` anywhere? Are you sure this is correct.

scripts/npmDepsHash.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+newHash=$(nix-shell -p prefetch-npm-deps --run "prefetch-npm-deps ./package-lock.json")
+
+echo "$newHash" > ./npmDepsHash
```
CMCDragonkai: To avoid issues with execution do not use `/bin/sh`. Use a regular JS script, and also follow the other JS scripts in `scripts/*`. Check other repos we are doing this for inspiration, do not use `/bin/sh` for anything as it is not suitable for these repos due to context switching.

flake.nix
```diff
@@ -0,0 +1,149 @@
...
+            cp ${if platform == "win32" then "out.exe" else "out"} $out
+          '';
+          dontFixup = true;
+        };
+
+        buildJSON = builtins.fromJSON (builtins.readFile "${polykey_cli-src}/build.json");
+
+        dockerImage = pkgs.dockerTools.buildImage {
+          name = polykey_cli-src.name;
+          copyToRoot = [ polykey_cli-src ];
```
CMCDragonkai: Hmm I'm not sure if I like this. `src` does not make sense here.

CMCDragonkai:
Let's have a meeting to through this since I think this has been stuck in limbo due to some unknown information. @tegefaulkes @brynblack

README.md
brynblack: Okay, there is a reason its like this. When making a flake, you need to specify the system package architecture that will be used to build the flake. Usually, this would be something like `x86_64-linux` or `aarch64-darwin` and so on. These are then used when calling something like nixpkgs.system and related architecture dependent attributes.

At the moment, I have stuck to the recommended naming scheme, as its standardised within the flake ecosystem, but using Nix expressions it is possible for me to map custom system targets to the standardised target naming. So if this is what you would prefer, this can be done.

flake.nix
brynblack: What I have done here is replicating the original `release.nix`. Before you could choose what platform to build a binary for, and in this case I implemented it as such; the only requirement being running the build process on a `x86_64-linux` system. There is no "cross-compilation" when it comes to making builds on any device, its strictly limited to Linux devices, but those devices are capable of leveraging `vercel/pkg` to build for other specific platforms.

But if this is not desired right now, then it is fine for it to be simplified down. Will get this done.

flake.nix
brynblack: To explain, I put two derivations of Polykey-CLI inside the flake, `polykey_cli-src` and `polykey_cli-release`. One of them contains the entire source tree, and is executed with Node automatically when running the output, which itself is a symlink to a script that will run `polykey.js` with Node.

I kept this like this because this is what the original `release.nix` was doing, it was importing the `default.nix` which used an, almost identical to what I have now, build process using `buildNpmPackage`.

I don't know if this should be changed or not, it seems to work perfectly fine as is right now.

brynblack:
> Other than that, what's keeping this from being merged?

Minor things really, I've left it open for review so I can make sure everything is perfect, I don't want a repeat of what happened last time with it getting reverted because I missed something.

brynblack:
> The nix-runnable binary does not exist, it's supposed to just use the `application` build, which I think you've renamed as just the `default` build... right?

Correct. When you build Polykey-CLI using the flake, it will output the `vercel/pkg` based build, which contains bundled Node and everything else. But as you said, this isn't really something to worry about right now, we have `steam-run` to solve this.

README.md
CMCDragonkai: Right so in the case of PK CLI, the only architecture we are using to "build" PK CLI is the `x86_64-linux`.

The target output though may only work on non-NixOS systems, Windows, and Mac.

Like I stated in the other comments, we are not at the point of cross-compiling nodejs itself. It's possible we may never do so if we use Deno and stuff.

I like to think of the JS code as running on a generic JS runtime - that itself should already be precompiled. Don't think of PK CLI as a C project - it is not, one should assume the underlying JS runtime is already precompiled and we just draw it from our dependencies.

That's why the pkg related nix code should actually be moved into our `nixpkgs-matrix` so we can re-use it across different JS-based projects.

CMCDragonkai: And if we have deno, that would then a whole another story too. But we're not at the point of thinking about that due to our native binaries atm being non-compatible.

flake.nix
CMCDragonkai: > Right so in the case of PK CLI, the only architecture we are using to "build" PK CLI is the x86_64-linux.

Add commentary about how to expand this if this changes in the future.

flake.nix
CMCDragonkai: Is it standard practice to name it the name of the application, or can flakes also just have a "default" attribute?

package.json
CMCDragonkai: Wasn't it supposed to be the `dependencies` script?

Also you want to group the npm native scripts together.

flake.nix
CMCDragonkai: So the idea of `src` isn't quite right. I consider that to be an "application" build output or the default build output.

The pkg built ones, should then be suffixed with the relevant output.

So let's make up a spec.

`polykey-cli` is the default.

`polykey-cli-pkg` is the pkg related stuff.

So then `polykey-cli-pkg-win32` should be...?

Think about it that way. Look at how we are naming things across the board, prefix is always the "bigger picture".

BTW maybe this wasn't documented anywhere, but we follow SOV naming style. Subject Object Verb. Or in this case, Subject Object.

CMCDragonkai:
Technically steam-run isn't necessary. It's 2 things: 1. the loader path 2. any c/c++ system library (in fact the js-db demands a specific relevant version). So it is possible to run on NixOS with a easier runner...

The `steam-run` is probably a bit overkill since it's used for games.

There's also this: https://github.com/thiagokokada/nix-alien to try out.

CMCDragonkai:
Oh even this (nix-ld) https://unix.stackexchange.com/questions/522822/different-methods-to-run-a-non-nixos-executable-on-nixos.

However don't apply that specifically to this problem. That's a platform problem, and you should be solving that in our `nixpkgs-matrix-private` and/or orchestrator.

CMCDragonkai:
Monday is it then! Make this merged.

Then you should squash your commits together.

CMCDragonkai:
@brynblack please make sure to test on windows and mac. Can you get those 2 systems online in epsilon dc?

flake.nix
brynblack: Okay, as of commit 1d7af25 this should be made a lot more clear through the code refactoring and comments. I have not removed the functionality to build for other systems, and it is still forced to use the `x86_64-linux` package set for `nixpkgs` when building. So this should be identical in functionality to the original build system in staging now.

package.json
brynblack: Fixed in 1d7af25.

flake.nix
brynblack: I renamed them to what you suggested in 1d7af25. Its a lot clearer now.

scripts/npmDepsHash.sh
brynblack: I have converted it to a js script now in 1d7af25.

flake.nix
brynblack: What you see there, `polykey_cli-release` is just a variable assigned to the derivation produced from `buildNpmPackage`. That name does not change anything. What does change things typically in a flake, is in the `packages` section in the flake outputs, the name of the package will change the binary name. In the case of the `vercel/pkg` system, since that uses a custom output system it will always output under the binary `result`. Same for the docker image.

brynblack: Also the naming is fixed in 1d7af25.

README.md
brynblack: Okay, for now I am leaving this as it is not possible to change the paths as easily in a flake output. Flake outputs adhere to a standard for the naming of systems and their derivations, so changing this would deviate from the way flakes are defined. It would probably be easier to leave it as I have done it for now, with the naming of an output such as `packages.x86_64-linux.polykey-cli`, which is automatically translated to the appropriate naming for it to be recognised by `vercel/pkg`. Look at 1d7af25 and the way I have clearly distinguished the build system and then the target systems.

brynblack:
I think that there is nothing left to really do here, everything is resolved and it should be ready for merge tomorrow. I have already tested the build process for every platform and derivation on my own device, to which it has all succeeded. 

Edit: The CI failed, need to investigate.

brynblack:
For some reason, the `check:test index` job seems to be failing. Not sure why.

CMCDragonkai:
Fast check found an error there.

```
bin/utils › encodeEscaped should encode all escapable characters
    Property failed after 50 tests
    { seed: 1318872640, path: "49:2:0:15:12", endOnFailure: true }
    Counterexample: ["\\r"]
    Shrunk 4 time(s)
    Got error: expect(received).toBe(expected) // Object.is equality
    Expected: "\\r"
    "
      261 |     fc.assert(
      262 |       fc.property(stringWithNonPrintableCharsArb, (value) => {
    > 263 |         expect(binUtils.decodeEscaped(binUtils.encodeEscaped(value))).toBe(
          |                                                                       ^
      264 |           value,
      265 |         );
      266 |       }),
      at toBe (tests/utils.test.ts:263:71)
      at Property.predicate (node_modules/fast-check/lib/check/property/Property.js:17:86)
      at Property.run (node_modules/fast-check/lib/check/property/Property.generic.js:49:33)
      at runIt (node_modules/fast-check/lib/check/runner/Runner.js:21:30)
      at check (node_modules/fast-check/lib/check/runner/Runner.js:73:11)
      at Object.assert (node_modules/fast-check/lib/check/runner/Runner.js:77:17)
      at Object.assert (tests/utils.test.ts:261:8)
      Hint: Enable verbose mode in order to have the list of all failing values encountered during the run
      at buildError (node_modules/fast-check/lib/check/runner/utils/RunDetailsFormatter.js:131:15)
      at throwIfFailed (node_modules/fast-check/lib/check/runner/utils/RunDetailsFormatter.js:143:11)
      at reportRunDetails (node_modules/fast-check/lib/check/runner/utils/RunDetailsFormatter.js:156:16)
      at Object.assert (node_modules/fast-check/lib/check/runner/Runner.js:81:52)
      at Object.assert (tests/utils.test.ts:261:8)
```

@amydevs @tegefaulkes

tegefaulkes:
There's been temperamental failures of some formatting tests since the main release work a while back. Not part of the scope here so it can be ignored. But needs to be addresses eventuallly.

tegefaulkes:
Seems like I'm taking this over and getting it merged today. Just need to sanity check all the jobs work and the build outputs.

CMCDragonkai:
Yes for staging CI jobs too, only can be tested once merged.

tegefaulkes:
I've found some minor things with just getting my bearings.

1. It's possible to run the flake without installing it with `nix run .#polykey-cli` But in this case it's broken in nixos. It's running the linux packaged output. I think it should be running the application build in this case. OR we fix the linux build to work with the nix system. I think we're missing a patchelf to make it work?
2. While we have an application version defined. its not actually exposed so we can run or build the application version.
3. After adding the application version I can see that it builds fine, but when running it with `nix run` it's trying to run `nixstorepath/bin/polykey-cli` but the executables are only aliased as `pk` or `polykey`. Adding `polykey-cli` to this does get it to work.

I'm going to test the other build outputs.

CMCDragonkai:
My thinking is that there are 2 kinds of builds:

1. A nix-like app build which is just a JS app that depends on nodejs.
2. A packaged ELF.

When building `polykey-cli` by itself it should be 1. this keeps it the way it used to be - and thus we consider that to be `polykey-cli` closure. The 2. is the thing that goes into the GitHub releases.

The `polykey-cli` should not be the "alias" here. How does that work? You should be producing just the standard package.json specified binaries so you can use just `pk` or `polykey`. The name of the flake output should have no bearing on the binaries.

CMCDragonkai:
See https://nixos.org/manual/nix/stable/command-ref/new-cli/nix3-run - I think we're missing a attribute to be used to make `3.` work with just `pk` or the more official `polykey`. It explains how to make `nix run` run a specific binary.

CMCDragonkai:
image

CMCDragonkai:
We wouldn't want to switch `pname` because it needs to be used by the `buildNpmPackage` to avoid org-scoping. This was the case before the flakes and still will be due to us continuing to use the same derivation abstraction.

Instead to override it, use `meta.mainProgram`.

It might work within the `buildNpmPackage`, but if it doesn't then you just need to apply it on the outside with an override.

CMCDragonkai:
Something like:

```
{ buildNpmPackage, ... }:

buildNpmPackage rec {
  meta = {
    # Other meta attributes like description, homepage, license, etc.
    mainProgram = "polykey";
  };
}
```

CMCDragonkai:
Naming of all release artifacts should be similar?

image

CMCDragonkai:
PR will need to be squashed and ticked off above.

tegefaulkes:
`nix build` can build multiple outputs at once like before. But there is a problem, it doesn't output the paths it built like it used too.

This is going to break the build jobs.

tegefaulkes:
Just need to add `--print-out-paths` flag to get the previous behaviour.

scripts/npmDepsHash.js
```diff
@@ -0,0 +1,27 @@
...
+
+const fs = require('fs');
+const path = require('path');
+const childProcess = require('child_process');
+
+/* eslint-disable no-console */
+async function main(_argv = process.argv) {
+  const projectRoot = path.join(__dirname, '..');
+
+  childProcess.exec(
```
CMCDragonkai: This should be wrapped as a promise or made synchronous. Why is this not `execSync`?

CMCDragonkai: In fact the piping into `tail` is not the best way to do this.

You should have just processed the output in JS and used `execFileSync` instead as it is most secure and fastest performance.

Remember the reason to use a JS script was to be portable.

CMCDragonkai: Look https://github.com/MatrixAI/Polykey-CLI/blob/5a31aba123d855d585264caef7bba9119e535786/scripts/pkg.js#L101. This is better.

shell.nix
```diff
@@ -1,44 +0,0 @@
...
-
-with pkgs;
-let
-  utils = callPackage ./utils.nix {};
-in
-  mkShell {
-    nativeBuildInputs = [
-      nodejs
-      prefetch-npm-deps
-      shellcheck
```
CMCDragonkai, line 11: Where are these tools specified in the flake?

We still need most of these in the case of `nix develop`.

scripts/npmDepsHash.js
```diff
@@ -0,0 +1,27 @@
...
+    'nix-shell -p prefetch-npm-deps --run "prefetch-npm-deps ./package-lock.json" | tail -n 1',
+    async (error, stdout, _) => {
+      if (error) {
+        console.error(`error: ${error.message}`);
+        return;
+      }
+
+      const hash = stdout.trim();
```
CMCDragonkai: Why is this needed? Just to eliminate the NL at the end? Generally NL is ok, but readers should trim not writers.

flake.nix
```diff
@@ -0,0 +1,151 @@
...
+          extraCommands = ''
+            mkdir -m 1777 tmp
+          '';
+          config = {
+            Entrypoint = [ "polykey" ];
+            Labels = buildJSON.versionMetadata;
+          };
+        };
+
+        shell = { ci ? false }: with pkgs; pkgs.mkShell {
```
CMCDragonkai, line 99: Here... Why not just put npm-deps-hash here?

CMCDragonkai: Like `prefetch-npm-deps`.

flake.nix
```diff
@@ -0,0 +1,151 @@
...
+        arch = builtins.elemAt systems.${targetSystem} 1;
+
+        pkgs = import nixpkgs {
+          system = buildSystem;
+        };
+
+        utils = pkgs.callPackage ./utils.nix {};
+
+        commitHash = toString (self.rev or self.dirtyRev);
+        npmDepsHash = builtins.readFile ./npmDepsHash;
```
CMCDragonkai, line 32: Do your NL trimming here?

CMCDragonkai: For a simpler approach to read file contents and trim a trailing newline in Nix, you can use a combination of `builtins.readFile` and a small Nix expression to directly address the trailing newline without needing to replace all occurrences or manipulate the string extensively. Here's a concise way to do it:

```nix
let
  rawContent = builtins.readFile "/path/to/your/file";
  trimmedContent = builtins.head (builtins.split "\n" rawContent);
in
  # Use trimmedContent as needed
```

This method splits the file content by newline characters into a list of strings, then takes the first element of this list with `builtins.head`. This effectively trims a single trailing newline if it exists, but note that if your file's content includes multiple lines, this will only capture the first line. If the intention is to remove only the final newline from a single-line file or the content is known to be a single line, this approach is significantly simpler and more straightforward. 

For files with multiple lines where you want to preserve all content except the trailing newline, the initial suggestion might still be necessary, albeit more complex. There isn't a built-in function in Nix that specifically targets only the trailing newline in a multi-line string, reflecting Nix's primary focus on package management and configuration rather than general-purpose text processing.

scripts/npmDepsHash.js
CMCDragonkai: NLs are legitimate file endings. Readers should trim. Not writers to a file.

tegefaulkes:
I've fixed up the `npmDepsHash.js` script. I'm resolving all the above comments since they all relate to the same thing.

I've done the following.
1. Using `execSync` now
2. Same as before, error is caught and warned but ultimately ignored so non nix systems still work.
3. The exec'ed command is just `prefetch-npm-deps ./package-lock.json > npmDepsHash` This cuts out all of the processing in the script. Though I need to fix that really quick.

scripts/npmDepsHash.js
tegefaulkes: Using execSync now.

shell.nix
tegefaulkes: It's defined in the flake. Only `prefetch-npm-deps` was missing and that has been fixed now.

scripts/npmDepsHash.js
tegefaulkes: I just tested without it and things seem to work fine. I've removed the trim.

tegefaulkes:
Everything has been addressed and the CI has passed for the feature branch jobs. I'll move on to merging this and then testing the rest of the CI.

If we run into any more CI job problems I'll fix them in the staging branch.

tegefaulkes:
I've squashed everything. I'm going to do a quick once over and merge.

CMCDragonkai:
I'd suggest changing to `execFileSync` when executing single programs.

brynblack:
Okay I think I'm happy with everything that's changed in the flakes implementation, I like how its now got a clear separation between the executable package and the Nix specific package that can be ran. Although I do wish we didn't have to have such a separation, if and/or when we do get Nix execution for builds working using the bundled package I think we can remove some complexity here. Otherwise, its good.

PR #129, opened 2024-02-13
Creating `secrets env` command

### LINKED ISSUE #31: The `pk secrets env` command for meeting Development Environment Usecase
### Specification

1. The `pk secrets env` command should operate similar to the unix `env` command
2. We can use `pk secrets env` to do double duty: inject environment variables into a new subcommand, or allow sourcing of environment variables into an existing subshell
3. The task for the development environment usecase is specified in https://github.com/MatrixAI/Polykey-Desktop/issues/77
4. The `pk secrets env someprogram arg1 arg2` should do a proper process replacement when possible

### Additional context

* https://github.com/MatrixAI/js-polykey/issues/55#issuecomment-658668800 - old comment about the env interface and the "chording" idea

Copied excerpt:

> Now when we want to start development on the project, we have 2 options:
> 
> 1. Open up a shell with Polykey sourcing the environment variables
> 2. Use Polykey to inject environment variables into a sub-shell
> 
> Either result is the same, because you'll be running a shell with environment variables set so that your test runs of your software can inherit this environment configuration. Let's demonstrate both.
> 
> In the first case, we will reuse the `source` command and ask `pk` to constuct the equivalent of the `.env` file but entirely in-memory and only for this specific usage. We must also use process substitution `<(...)` so that the `pk` command output can be redirected into a temporary file descriptor that is read by the `source` command.
> 
> ```sh
> # one secret
> source <(pk secrets env my-software-project:AWS_ACCESS_KEY_ID)
> 
> # multiple secrets
> source <(pk secrets env my-software-project:AWS_ACCESS_KEY_ID my-software-project:GOOGLE_MAPS_API_KEY)
> 
> # globbing style (you can use globstar as well, this will only export immediate files)
> source <(pk secrets env my-software-project:*)
> 
> # use the -e flag to export all variables
> source <(pk secrets env -e my-software-project:*)
> # use the -- to separate so you can export just one out of many
> source <(pk secrets env -- my-software-project:AWS_ACCESS_KEY_ID -e my-software-project:GOOGLE_MAPS_API_KEY)
> source <(pk secrets env -e -- my-software-project:AWS_ACCESS_KEY_ID my-software-project:GOOGLE_MAPS_API_KEY)
> ```
> 
> Now your shell has the relevant environment variables set by Polykey, and they will exist for as long as this current shell is alive.
> 
> In the second case, we ill use `pk secrets env` command to run a subshell, it can run any subprogram, but in the context of a development environment, you usually want a shell.
> 
> ```sh
> # one secret
> pk secrets env my-software-project:AWS_ACCESS_KEY_ID bash
> 
> # multiple secrets
> pk secrets env my-software-project:AWS_ACCESS_KEY_ID my-software-project:GOOGLE_MAPS_API_KEY bash
> 
> # globbing style
> pk secrets env my-software-project:* bash
> 
> # the -e flag has no effect when using it this way
> # this is because the subprogram determines whether to export variable or not
> # it usually exports the variable
> pk secrets env -e my-software-project:* bash
> ```
> 
> Now you have a subshell that has the environment variables configured, it will also automatically export it to child programs. When you have finished your development, you can just `exit` the shell, and the environment variables are gone!

### Tasks

1. Ensure that `pk secrets env` can be used to inject environment variables and run a subprocess
2. Ensure that `pk secrets env` can be used to source environment variables and output something that can be used as a file to be sourced by a shell
3. Deal with the fact that nodejs doesn't have an exec that can replace the current node process: https://github.com/nodejs/node/issues/21664
   * You can deal with this the way babel deals with it

> * Babel [attempts to use `kexec`](https://github.com/babel/babel/blob/master/packages/babel-node/src/babel-node.js#L87-L88) where available, which [is a POSIX-only module that replaces the process literally](https://www.npmjs.com/package/kexec). Absent that, it falls back to [its own implementation](https://github.com/babel/babel/blob/master/packages/babel-node/src/babel-node.js#L90-L109) that works like the other two examples.

tegefaulkes:
### Description

This PR addresses adding the `secrets env` command to the cli

### Issues Fixed

* Fixes #31 

### Tasks

- [x] 1. parse input for selecting desired secrets from vaults
- [x] 2. obtain secrets from the Polykey agent
- [x] 3. ability to output env variables in selectable formats
- [x] 4. ability to run commands with env variables only for the duration of that command
- [x] 5. ability to have the running command replace the current process that invoked it.
- [x] 6. created tests to demonstrate and test functionality

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/npmDepsHash
+++ b/npmDepsHash
@@ -1,1 +1,1 @@
-sha256-Tkycu9+7Kwjvuaplw38b/t1N4ZRdsOack5OC0ZO0cxo=
+sha256-hS9LURKsHIb8bg/LEbGGkrHP4oAxDGeQKvopl8PhuL4=
--- a/npmDepsHash
+++ b/npmDepsHash
@@ -1,1 +1,1 @@
-sha256-buIzGNjR5CMM7BNlw5srJtVTSH/6Ru+UsMeXeNOjgZ4=
+sha256-LZ3QekhwOZj3IkbnZVdC/lbUvwhbJXROzd9oTxrMs7s=
--- a/npmDepsHash
+++ b/npmDepsHash
@@ -1,1 +1,1 @@
-sha256-LZ3QekhwOZj3IkbnZVdC/lbUvwhbJXROzd9oTxrMs7s=
+sha256-Tkycu9+7Kwjvuaplw38b/t1N4ZRdsOack5OC0ZO0cxo=
--- a/package-lock.json
+++ b/package-lock.json
@@ -40,7 +40,7 @@
         "nexpect": "^0.6.0",
         "node-gyp-build": "^4.4.0",
         "nodemon": "^3.0.1",
-        "polykey": "^1.2.1-alpha.47",
+        "polykey": "^1.2.1-alpha.49",
         "prettier": "^3.0.0",
         "shelljs": "^0.8.5",
         "shx": "^0.3.4",
@@ -1476,9 +1476,9 @@
       ]
     },
     "node_modules/@matrixai/quic": {
-      "version": "1.1.4",
-      "resolved": "https://registry.npmjs.org/@matrixai/quic/-/quic-1.1.4.tgz",
-      "integrity": "sha512-WhwoFzw2fuxDCxzY/WTZnMHg2snSGpGvifh4K6iRPfnvUfKME0ikFty7EMkRt8bQiAes4yxBbcI72/wBSeTYMA==",
+      "version": "1.2.1",
+      "resolved": "https://registry.npmjs.org/@matrixai/quic/-/quic-1.2.1.tgz",
+      "integrity": "sha512-Dz/IzRrYIV407aG3GAM7+4eoqntBZ/QOP4sBugd/5ZgiS8vwIdXQzzfGfetcM9SRp3I67v5teXDpavb4JAcWYg==",
       "dev": true,
       "dependencies": {
         "@matrixai/async-cancellable": "^1.1.1",
@@ -1493,16 +1493,16 @@
         "ip-num": "^1.5.0"
       },
       "optionalDependencies": {
-        "@matrixai/quic-darwin-arm64": "1.1.4",
-        "@matrixai/quic-darwin-x64": "1.1.4",
-        "@matrixai/quic-linux-x64": "1.1.4",
-        "@matrixai/quic-win32-x64": "1.1.4"
+        "@matrixai/quic-darwin-arm64": "1.2.1",
+        "@matrixai/quic-darwin-x64": "1.2.1",
+        "@matrixai/quic-linux-x64": "1.2.1",
+        "@matrixai/quic-win32-x64": "1.2.1"
       }
     },
     "node_modules/@matrixai/quic-darwin-arm64": {
-      "version": "1.1.4",
-      "resolved": "https://registry.npmjs.org/@matrixai/quic-darwin-arm64/-/quic-darwin-arm64-1.1.4.tgz",
-      "integrity": "sha512-/Dho39T4sjbdOeC01F4rDesssQ9nvfdH57Ap95870LoY1KwXWh+c7cAtdv9FVma2R+9cqdKNkfsMFcS68Y15KA==",
+      "version": "1.2.1",
+      "resolved": "https://registry.npmjs.org/@matrixai/quic-darwin-arm64/-/quic-darwin-arm64-1.2.1.tgz",
+      "integrity": "sha512-TutHRdQoPYPvwg8Fs6t+JmedKWn/7mJG63c5ZFyw9D58a90OzYr47re2qknKNjH/dgBIyybhBYTAltkOcXxzOg==",
       "cpu": [
         "arm64"
       ],
@@ -1512,9 +1512,9 @@
       ]
     },
     "node_modules/@matrixai/quic-darwin-x64": {
-      "version": "1.1.4",
-      "resolved": "https://registry.npmjs.org/@matrixai/quic-darwin-x64/-/quic-darwin-x64-1.1.4.tgz",
-      "integrity": "sha512-soxYopXZjoHxSMvpoNEKQQS1uQcXK2aIuaJHDgZINz4UbeLW0NsoEqMgUJKXeWZNR+aaVxhL5fE5+5Ypwej80Q==",
+      "version": "1.2.1",
+      "resolved": "https://registry.npmjs.org/@matrixai/quic-darwin-x64/-/quic-darwin-x64-1.2.1.tgz",
+      "integrity": "sha512-HMIhCHjAvvAgSap2gQyKYgWyhikDtmU9KbqtReMT3tP8KG3jNPAw2YuF7vucVR9k6FaMufISLYyBvWQSadTxIg==",
       "cpu": [
         "x64"
       ],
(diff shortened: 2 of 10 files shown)


tegefaulkes:
There will be a companion PR in `Polykey` for tracking the lib changes needed. There will also likely be a new repo for the native code needed to do kexec.

CMCDragonkai:
The repo https://github.com/MatrixAI/js-exec still needs some work - lots of configuration needed - README, and permissions, CI integration into Gitlab.

tegefaulkes:
Yeah, I'm still in the middle of working on it.

CMCDragonkai:
@tegefaulkes the `execvpe` does not work on MacOS.

For Unix compatibility you need to start with `excevp`, and simulate `execvpe` by setting the environment yourself.

Example provided in C:

```c
#include <stdio.h>
#include <stdlib.h>
#include <unistd.h>

int main(int argc, char *argv[], char *envp[]) {
    // Add or modify an environment variable
    setenv("MY_ENV_VAR", "some_value", 1);

    // Prepare arguments for execvp
    char *new_argv[] = {"program_name", "arg1", "arg2", NULL};

    // Execute the program with the modified environment
    execvp(new_argv[0], new_argv);

    // If execvp returns, it failed
    perror("execvp");
    return EXIT_FAILURE;
}
```

However, since you're using rust, I'm sure there's equivalents for it.

What's the difference between using:

* https://docs.rs/uapi/latest/uapi/fn.execvpe.html
* https://docs.rs/libc/latest/libc/fn.execvpe.html

CMCDragonkai:
Have a look: https://chat.openai.com/share/e/9c9f1315-d600-49c4-91a1-667ecb8ca0cb

Example with `libc`

```rs
extern crate libc;

use libc::{execvp, setenv};
use std::ffi::CString;
use std::os::unix::ffi::OsStrExt;
use std::process;

fn main() {
    // Set an environment variable
    let key = CString::new("MY_VARIABLE").unwrap();
    let value = CString::new("some_value").unwrap();
    unsafe {
        if setenv(key.as_ptr(), value.as_ptr(), 1) != 0 {
            eprintln!("Failed to set environment variable");
            process::exit(1);
        }
    }

    // Prepare the arguments for execvp
    let program = CString::new("/usr/bin/env").unwrap();
    let args = [
        CString::new("env").unwrap().as_ptr(), // argv[0], conventionally the program name
        std::ptr::null(),                       // Arguments must be terminated by a NULL
    ];

    // Execute the command, replacing the current process
    unsafe {
        execvp(program.as_ptr(), args.as_ptr());
        // If execvp returns, it means an error occurred
        eprintln!("Error executing execvp");
        process::exit(1);
    }
}
```

Also I see alot of vestigial comments, can you clean up the code of js-exec and make sure errors are all handled.

CMCDragonkai:
If uapi is better, you can use that but you'll need to figure out how to set the environment. https://stackoverflow.com/questions/59922661/why-doesnt-unistd-h-execvpe-work-on-macos

CMCDragonkai:
I updated a bit of the repo configuration in https://github.com/MatrixAI/js-exec. The usage in `README.md` should be updated too.

Note that since on Windows it is expected to throw an exception, can you create a `src/errors.ts` to create the domain set of errors.

Also on Windows, to simulate the same behaviour, you'd need to be able to create a subprocess and detach it and then terminate the parent process. This can be put for a future issue to address, for now a standard childprocess can be done on Windows. Which means for now using this may be a bit inefficient on Windows.

tegefaulkes:
OK, so we need to add the following before we can 
1. We need to validate secret and override names to make sure they're valid environment variable names.
1. add flags for switching the behaviour of the following; we need to switch between throw, warning or silent. default to throw.
  a. how to handle duplicate env names
  b. how to handle invalid env names
2. we need to support the windows platform. Just need to switch to an alternative such as `child_process.exec` do match the behaviour but without process replacement.
3. General polish and review.

I also need to go back to the `js-exec` and review that too. Make sure that all commentary is addressed.

tegefaulkes:
TODO:

- [x] 1. when renaming a secret we need to validate the new secret name
- [x] 2. when getting secret name from the vault we need to validate if the secret name is valid
- [x] 3. when validating the secret name we need to switch between 3 behaviours. 'ignore', 'warn', 'error'. 
- [x] 4. when handling duplicate secret names we need to switch between the following behaviour. `keep`, `overwrite`, `warn`, `error`. 
- [x] 5. when running on windows we need to switch to using `child_process.exec` to run the command. This means we can only  dynamically load `js-exec` if it's supported by the platform.
- [x] 6. general polish
- [x] 7. general polish of `js-exec` lib.

tegefaulkes:
One thing I'm concerned about is that I'm using two different format options. 
1. `--format` is not actually used, is included by default in the command and I cant actually override it.
2. `--env-format` is an option I've created with the following options. `'dotenv', 'json', 'prepend'`.

Ideally I'd just have `--format` function like `--env-format` but we'd have to remove `--format` from being automatically included in all commands and re add it in to all command implementations.

tegefaulkes:
I'ts possible to modify the options that have already been set. within the constructor when creating the command we can access the options array with `this.options`. It's in the format of `Array<Option>`. We can add or remove choices to the `this.options[x].argChoices.push('newOption')` so we just need to find the option we want to modify.

Note that while `this.options` can be accessed its not actually in the types.

tegefaulkes:
Re-based on staging and ready to merge.

CMCDragonkai:
> One thing I'm concerned about is that I'm using two different format options.
> 
> 1. `--format` is not actually used, is included by default in the command and I cant actually override it.
> 2. `--env-format` is an option I've created with the following options. `'dotenv', 'json', 'prepend'`.
> 
> Ideally I'd just have `--format` function like `--env-format` but we'd have to remove `--format` from being automatically included in all commands and re add it in to all command implementations.

Hold up! Did you have a read about https://github.com/MatrixAI/Polykey-CLI/issues/31#issuecomment-1957872145 - there's important differences between different terminal/shell environments.

And also it's important to acknowledge what we mean by the `--format` option. It's a program wide option to specify what is the format of data we should provide via STDOUT and STDERR. The default of course is `--format human` which is supposed to mean that it is "readable" by human beings.

Now if you're going to overload that option I need you to explain what `prepend` or `dotenv` is supposed to be. Plus you must document it in the CLI help too.

Here's what I'm proposing.

* `human` means human-readable - it doesn't get any more specific that that

The problem is that we may deal 3 different "shell" environments:

* Unix shells where `X=Y` works
* CMD
* Powershell

It is important to note that none of these variables should be auto exported without an additional flag being `--export` option.

So... I think `human` should automatically choose the right format by checking the parent process shell some how.

IF we cannot do that, then if you overload the `--format` then you need to decide what `human` means, and what if you can instead do `unix`, `cmd`, `powershell` formats instead.

CMCDragonkai:
So by default variables are only local to the shell. Then with an extra `--export` flag, depending on the type of the shell that is running, you would then wrap the whole thing an export system. For unix, you'd use `export X=Y`, for powershell you need `$env:X = ""`, for CMD there's a different method again.

CMCDragonkai:
You might want to address the `--export` option in a separate issue.

CMCDragonkai:
Make sure to test this on the windows and mac machines we got too!

tegefaulkes:
Added follow on issues at #145 and #144

PR #133, opened 2024-02-19
ci: merge staging to master

maxwell-aiden:
This is an automatic PR generated by the pipeline CI/CD. This will be automatically fast-forward merged if successful.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -31,13 +31,9 @@
         commitHash = toString (self.rev or self.dirtyRev);
         npmDepsHash = builtins.readFile ./npmDepsHash;
 
-        polykey-cli-app = pkgs.buildNpmPackage {
+        polykey-cli = pkgs.buildNpmPackage {
           inherit npmDepsHash;
           pname = utils.packageName;
-          meta = {
-            # Other meta attributes like description, homepage, license, etc.
-            mainProgram = "polykey";
-          };
           version = utils.packageVersion;
           src = utils.src;
           COMMIT_HASH = commitHash;
@@ -58,7 +54,7 @@
           '';
         };
 
-        polykey-cli-pkg = pkgs.buildNpmPackage {
+        polykey-cli-executable = pkgs.buildNpmPackage {
           inherit npmDepsHash;
           name = "${utils.packageName}-${utils.packageVersion}-${platform}-${arch}";
           src = utils.src;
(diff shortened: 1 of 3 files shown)


maxwell-aiden:
Pipeline Attempt on 1182543206 for 04b6eaae290014f02f8ec684ed81faed8bc231b8

https://gitlab.com/MatrixAI/open-source/Polykey-CLI/-/pipelines/1182543206

flake.nix
```diff
@@ -31,13 +31,9 @@
...
         npmDepsHash = builtins.readFile ./npmDepsHash;
 
-        polykey-cli-app = pkgs.buildNpmPackage {
+        polykey-cli = pkgs.buildNpmPackage {
           inherit npmDepsHash;
           pname = utils.packageName;
-          meta = {
-            # Other meta attributes like description, homepage, license, etc.
-            mainProgram = "polykey";
-          };
```
CMCDragonkai, line 40: Turns out this was not needed because of the `type: "app"` in the `apps`. It then just uses the `program` attribute. This is only needed if we just refer to a derivation directly

flake.nix
```diff
@@ -138,18 +134,18 @@
...
           };
         };
 
         packages = {
-          polykey-cli = polykey-cli-pkg;
-          application = polykey-cli-app;
-          docker = dockerImage;
+          default = polykey-cli;
+          executable = polykey-cli-executable;
+          docker = docker;
```
CMCDragonkai, line 148: This most closely aligns to our original `default.nix` and such. And yes the flake system is definitely much easier! Great work @brynblack and @tegefaulkes. Now running it is much easier than remember all the `nix-build` commands. I guess the only thing is that there's no parameterisation... That's interesting.

CMCDragonkai:
I didn't study the flake utils too much, but do we actually need it? https://ayats.org/blog/no-flake-utils/

maxwell-aiden:
Pipeline Succeeded on 1182543206 for 04b6eaae290014f02f8ec684ed81faed8bc231b8

https://gitlab.com/MatrixAI/open-source/Polykey-CLI/-/pipelines/1182543206

flake.nix
tegefaulkes: Okay then.

flake.nix
tegefaulkes: `nix build` does have an `--argstr` parameter but it's not exactly clear how it works. Maybe parameters are allowed?

brynblack:
> I didn't study the flake utils too much, but do we actually need it? https://ayats.org/blog/no-flake-utils/

`flake-utils` just helps with building for multiple targets, normally when defining a flake you have to re-add boilerplate to build for more than one target and it gets annoying quick. So basically its just a utility to make it easier to understand and use.

flake.nix
brynblack: No, there are no parameters that can be passed. Flakes as I have said before are reproducible and sandboxed, and if you saw my Excalidraw then this would make more sense. This was the case with `npmDepsHash`, as we could not dynamically just put it in at runtime.

There does exist a way to pass parameters, but that would mean making the flakes impure, which is considered a very bad practice as now there is a possible dependency on an arbitrary parameter for a desired output. I think its best if we keep it for now with how its done, we can always build specialisations of a derivative through hard coded parameters, and then later select that derivative. I think there also exists a way to add properties and attributes to an output, so it may be possible to do more there.

PR #137, opened 2024-02-23
Upgrade to Node 20

### LINKED ISSUE #111: Update to Node20 and switch `pkg`

amydevs:
### Description
This PR upgrades the Node version for all development environments to Node 20, and enables building of arm64 macos builds.

### Issues Fixed

* Fixes #111 

### Tasks

- [x] 1. Update `utils.nix` node package to `nodejs_20` for buildsp
- [x] 2. Update `flake.nix` `mkShell` inputs node package to `nodejs_20` for `nix develop` environments
- [x] 3. Update `scripts/choco-install.ps1` to use `nodev20.5.1`
- [x] 4. Update `scripts/brew-install.sh` to use `node@20`.
- [x] 5. Add macos Arm64 builds to `.gitlab-ci.yml`
- [x] 6. Docblock `utils.pkStdio` to warn usage when the expected `process.exitCode` is non-zero. 

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -248,6 +248,7 @@ integration:builds:
         .#packages.x86_64-linux.executable \
         .#packages.x86_64-windows.executable \
         .#packages.x86_64-darwin.executable \
+        .#packages.aarch64-darwin.executable \
       )";
       cp -r $(echo $builds | tr '\n' ' ') ./builds/;
   artifacts:
--- a/flake.nix
+++ b/flake.nix
@@ -12,7 +12,9 @@
       # The target systems and their vercel/pkg mapping
       systems = {
         "x86_64-linux" = [ "linux" "x64" ];
+        "aarch64-linux" = [ "linux" "arm64" ];
         "x86_64-windows" = [ "win32" "x64" ];
+        "aarch64-windows" = [ "win32" "arm64" ];
         "x86_64-darwin" = [ "darwin" "x64" ];
         "aarch64-darwin" = [ "darwin" "arm64" ];
       };
@@ -94,7 +96,7 @@
 
         shell = { ci ? false }: with pkgs; pkgs.mkShell {
           nativeBuildInputs = [
-            nodejs
+            nodejs_20
             prefetch-npm-deps
             shellcheck
             gitAndTools.gh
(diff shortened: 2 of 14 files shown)


CMCDragonkai:
The problem with the jest tests is:

* https://github.com/jestjs/jest/issues/9324
* https://github.com/jestjs/jest/issues/14501

We need a workaround for dealing with `process.exitCode = `.

amydevs:
> The problem with the jest tests is:
> 
>     * [JEST tests complete successfully but returns exit status 1 jestjs/jest#9324](https://github.com/jestjs/jest/issues/9324)
> 
>     * [[Bug]: In Node 20, Jest exits with non-zero exit code if `process.exitCode` is set in the tested code jestjs/jest#14501](https://github.com/jestjs/jest/issues/14501)
> 
> 
> We need a workaround for dealing with `process.exitCode = `.

@CMCDragonkai 

The workaround I have opted, is to docblock `pkStdio` which was mocking a Polykey instance rather than creating a subprocess, with warnings regarding usage when the expected exitCode is non-zero. I have then replaced the usage of `pkStdio` wherever the exitCode is expected to be non-zero.

An alternative solution could be to just set the exitCode to zero at the end of `pkStdio`.

CMCDragonkai:
Did you make sure to indicate in that docblock that the reason for this workaround is solely due a jest bug, and link the jest bug itself when it gets resolved?

CMCDragonkai:
> An alternative solution could be to just set the exitCode to zero at the end of pkStdio.

Not sure if that would work, after all isn't it the internal command that is executed that does it?

CMCDragonkai:
@amydevs @tegefaulkes you must remember to set the assignments - otherwise these won't be tracked later in R&D.

PR #138, opened 2024-02-25
NixOS Module Configuration

### LINKED ISSUE #143: NixOS Service Configuration via `flake.nix`
### Specification

The below specification is with regards to secret zero problem, however the actual PR solving this is just introducing the service configuration. For now, the "secret zero" of PK which is the root password, is just expected to exist in the home directory.

> It seems we are running into the secret zero problem with regard to the new Polykey Agent nix service. When running `pk agent start` for the first time, it is required that a password is supplied to initialise the agent. However that password needs to be managed by something, and appropriately stored somewhere. Currently, the service will look for a password file under `/root/.local/share/pass`, but this is unoptimal for our situation. Keeping it in a place which is more secure and standardised will be ideal. Possibly `/run/credentials` or `/run/keys`.
>
> My proposal is that there are a few solutions to solving this problem.
> - Systemd supports the configuration directive `EnvironmentFile` which lets the service pull environment variables from a specified file. We could then in the service configuration use that environment variable in the setup script; or
> - We set the `Environment` directive inside the Systemd service to pull from a given environment variable accessible to the system, but not from a file. Perhaps we inject the password somehow when building the system, but with flakes that doesn't seem likely; or
> - We use `systemd-ask-password` which lets the user enter a password to be used in the service. This could be a simple way of specifying it safely.
>
> Additionally, the Polykey Agent normally outputs a recovery code upon the initial setup of the agent. This will need to also be stored somewhere safe. A new CLI flag will need to be created to output the recovery codes to a specified directory (something similar to `-pf`).

@tegefaulkes @amydevs 

### Tasks

1. A flag inside of Polykey-CLI will need to be added to allow the secret output directory to change.
2. The service will need to be configured to use one of the solutions above and use the password in the service.
3. Documentation will need to be written up on how to set up the agent with the given solution.

brynblack:
### Description

This PR adds a module to Polykey-CLI that allows for the automated activation of the Polykey Agent through a `systemd` service, both in user-space and system-wide space.

Two module configurations are provided, both a `programs` module and a `services` module. Fundamentally, both have the same configuration options. The main difference here is that the programs module gives each user their own unit service that can be enabled on a per-user basis, while the system service can be managed by a root user, or members of the group `polykey` (a lot like Docker daemon).

At the moment, the module will require the user to provide a path to the password/recovery-code files, and will otherwise fail if not provided.

In the future by default, the system level service will read from a directory under `/var/lib/polykey` for the relevant password and recovery code files, and under the user-level service, will read from `~/.config/polykey`. These paths may be overwritten in the module configuration.

feature-module Spec excalidraw

### Issues Fixed

* Fixes #143 

### Tasks

- [x] 1. Implement the service module
- [x] 2. Implement the programs module
- [x] 3. Add minor finishing touches
- [x] 3. Test module and expected behaviour

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -24,7 +24,6 @@
 
             passwordFilePath = mkOption {
               type = with types; uniq str;
-              default = "/this/path/will/cause/a/failure/if/not/set";
               description = ''
               The path to the Polykey password file. This is required to be set for the module to work, otherwise this module will fail.
               '';
@@ -56,7 +55,6 @@
 
             passwordFilePath = mkOption {
               type = with types; uniq str;
-              default = "/this/path/will/cause/a/failure/if/not/set";
               description = ''
               The path to the Polykey password file. This is required to be set for the module to work, otherwise this module will fail.
               '';
--- a/flake.nix
+++ b/flake.nix
@@ -13,7 +13,153 @@
       systems = {
         "x86_64-linux" = [ "linux" "x64" ];
       };
+
     in
+    {
+      nixosModules.default = { config, ... }: with nixpkgs; with lib;
+      {
+        options = {
+          services.polykey = {
+            enable = mkEnableOption "Enable the Polykey agent. Users with the `polykey` group or root permissions will be able to manage the agent.";
+
+            passwordFilePath = mkOption {
+              type = with types; uniq str;
+              default = "/this/path/will/cause/a/failure/if/not/set";
+              description = ''
+              The path to the Polykey password file. This is required to be set for the module to work, otherwise this module will fail.
+              '';
+            };
+
+            recoveryCodeFilePath = mkOption {
+              type = with types; uniq str;
+              default = "";
+              description = ''
+              The path to the Polykey recovery code file. This is not required, but if set will read a recovery code from the provided path to bootstrap a new state with.
+              '';
+            };
+
+            recoveryCodeOutPath = mkOption {
+              type = with types; uniq str;
+              description = ''
+              The path to the Polykey recovery code file output location.
+              '';
+            };
+
+            statePath = mkOption {
+              type = with types; uniq str;
+              default = "/var/lib/polykey";
+              description = "The path to the Polykey node state directory. Will default to `/var/lib/polykey`, but can be overwritten to a custom path.";
+            };
+          };
+          programs.polykey = {
+            enable = mkEnableOption "Enable the per-user Polykey agent.";
+
+            passwordFilePath = mkOption {
+              type = with types; uniq str;
+              default = "/this/path/will/cause/a/failure/if/not/set";
+              description = ''
+              The path to the Polykey password file. This is required to be set for the module to work, otherwise this module will fail.
+              '';
+            };
+
+            recoveryCodeFilePath = mkOption {
+              type = with types; uniq str;
+              default = "";
+              description = ''
+              The path to the Polykey recovery code file. This is not required, but if set will read a recovery code from the provided path to bootstrap a new state with.
+              '';
+            };
+
+            recoveryCodeOutPath = mkOption {
+              type = with types; uniq str;
+              description = ''
+              The path to the Polykey recovery code file output location.
+              '';
+            };
+
+            statePath = mkOption {
+              type = with types; uniq str;
+              default = "%h/.local/share/polykey";
+              description = "The path to the Polykey node state directory. Will default to `$HOME/.local/share/polykey`, but can be overwritten to a custom path.";
+            };
+          };
+        };
+        config = mkMerge [
+          (mkIf config.services.polykey.enable {
+            users.groups.polykey = {};
+
+            environment.systemPackages = [
+              self.outputs.packages.${buildSystem}.default
+            ];
+
+            system.activationScripts.makeAgentPaths = ''
+              mkdir -p ${config.services.polykey.statePath}
+              chgrp -R polykey ${config.services.polykey.statePath}
+              chmod 770 ${config.services.polykey.statePath}
+            '';
+
+            systemd.services.polykey = {
+              description = "Polykey Agent";
+              wantedBy = [ "multi-user.target" ];
+              after = [ "network.target" ];
+              serviceConfig = {
+                User = "root";
+                Group = "polykey";
+                PermissionsStartOnly = true;
+                LoadCredential = [
+                  "password:${config.services.polykey.passwordFilePath}"
+                ];
+                ExecStartPre = ''
+                  -${self.outputs.packages.${buildSystem}.default}/bin/polykey \
+                  --password-file ''${CREDENTIALS_DIRECTORY}/password \
+                  --node-path ${config.services.polykey.statePath} \
+                  bootstrap  ${lib.optionalString (config.services.polykey.recoveryCodeFilePath != "") '' -rcf ${config.services.polykey.recoveryCodeFilePath}''}\
+                  --recovery-code-out-file ${config.services.polykey.recoveryCodeOutPath}
+                '';
+                ExecStart = ''
+                  ${self.outputs.packages.${buildSystem}.default}/bin/polykey \
+                  --password-file ''${CREDENTIALS_DIRECTORY}/password \
+                  --node-path ${config.services.polykey.statePath} \
+                  agent start \
+                  --recovery-code-out-file ${config.services.polykey.recoveryCodeOutPath}
+                '';
+              };
+            };
+          })
+          (mkIf config.programs.polykey.enable {
+            environment.systemPackages = [
+              self.outputs.packages.${buildSystem}.default
+            ];
+
+            system.activationScripts.makeUserAgentPaths = ''
+              mkdir -p ${config.programs.polykey.statePath}
+            '';
+
+            systemd.user.services.polykey = {
+              description = "Polykey Agent";
+              wantedBy = [ "default.target" ];
+              after = [ "network.target" ];
+              serviceConfig = {
+                ExecStartPre = ''
+                  -${self.outputs.packages.${buildSystem}.default}/bin/polykey \
+                  --password-file ${config.programs.polykey.passwordFilePath} \
+                  --node-path ${config.programs.polykey.statePath} \
+                  bootstrap  ${lib.optionalString (config.programs.polykey.recoveryCodeFilePath != "") '' -rcf ${config.programs.polykey.recoveryCodeFilePath}''}\
+                  --recovery-code-out-file ${config.programs.polykey.recoveryCodeOutPath}
+                '';
+                ExecStart = ''
+                  ${self.outputs.packages.${buildSystem}.default}/bin/polykey \
+                  --password-file ${config.programs.polykey.passwordFilePath} \
+                  --node-path ${config.programs.polykey.statePath} \
+                  agent start \
+                  --recovery-code-out-file ${config.programs.polykey.recoveryCodeOutPath}
+                '';
+              };
+            };
+          })
+        ];
+      };
+    } //
     flake-utils.lib.eachSystem (builtins.attrNames systems) (targetSystem:
       let
         platform = builtins.elemAt systems.${targetSystem} 0;
--- a/src/agent/CommandStart.ts
+++ b/src/agent/CommandStart.ts
@@ -12,7 +12,7 @@ import type { DeepPartial } from 'polykey/dist/types';
 import type { RecoveryCode } from 'polykey/dist/keys/types';
 import childProcess from 'child_process';
 import process from 'process';
-import * as fs from 'fs';
+import fs from 'fs';
 import config from 'polykey/dist/config';
 import * as keysErrors from 'polykey/dist/keys/errors';
 import * as polykeyEvents from 'polykey/dist/events';
--- a/src/agent/CommandStart.ts
+++ b/src/agent/CommandStart.ts
@@ -12,6 +12,7 @@ import type { DeepPartial } from 'polykey/dist/types';
 import type { RecoveryCode } from 'polykey/dist/keys/types';
 import childProcess from 'child_process';
 import process from 'process';
+import * as fs from 'fs';
 import config from 'polykey/dist/config';
 import * as keysErrors from 'polykey/dist/keys/errors';
 import * as polykeyEvents from 'polykey/dist/events';
@@ -28,6 +29,7 @@ class CommandStart extends CommandPolykey {
     this.name('start');
     this.description('Start the Polykey Agent');
     this.addOption(binOptions.recoveryCodeFile);
+    this.addOption(binOptions.recoveryCodeOutFile);
     this.addOption(binOptions.clientHost);
     this.addOption(binOptions.clientPort);
     this.addOption(binOptions.agentHost);
@@ -265,12 +267,18 @@ class CommandStart extends CommandPolykey {
           type: options.format === 'json' ? 'json' : 'dict',
           data: {
             ...statusLiveData!,
-            ...(recoveryCodeOut != null
+            ...(recoveryCodeOut != null && options.recoveryCodeOutFile == null
               ? { recoveryCode: recoveryCodeOut }
               : {}),
           },
         }),
       );
+      if (options.recoveryCodeOutFile != null && recoveryCodeOut != null) {
+        await fs.promises.writeFile(
+          options.recoveryCodeOutFile,
+          recoveryCodeOut,
+        );
+      }
     });
   }
 }
--- a/src/bootstrap/CommandBootstrap.ts
+++ b/src/bootstrap/CommandBootstrap.ts
@@ -1,4 +1,5 @@
 import process from 'process';
+import * as fs from 'fs';
 import CommandPolykey from '../CommandPolykey';
 import * as binUtils from '../utils';
 import * as binOptions from '../utils/options';
@@ -10,6 +11,7 @@ class CommandBootstrap extends CommandPolykey {
     this.name('bootstrap');
     this.description('Bootstrap Keynode State');
     this.addOption(binOptions.recoveryCodeFile);
+    this.addOption(binOptions.recoveryCodeOutFile);
     this.addOption(binOptions.fresh);
     this.addOption(binOptions.privateKeyFile);
     this.addOption(binOptions.passwordOpsLimit);
@@ -37,14 +39,22 @@ class CommandBootstrap extends CommandPolykey {
         logger: this.logger,
       });
       this.logger.info(`Bootstrapped ${options.nodePath}`);
-      process.stdout.write(
-        binUtils.outputFormatter({
-          type: options.format === 'json' ? 'json' : 'dict',
-          data: {
-            recoveryCode: recoveryCodeOut,
-          },
-        }),
-      );
+
+      if (options.recoveryCodeOutFile == null) {
+        process.stdout.write(
+          binUtils.outputFormatter({
+            type: options.format === 'json' ? 'json' : 'dict',
+            data: {
+              recoveryCode: recoveryCodeOut,
+            },
+          }),
+        );
+      } else if (recoveryCodeOut != null) {
+        await fs.promises.writeFile(
+          options.recoveryCodeOutFile,
+          recoveryCodeOut,
+        );      
+      }
     });
   }
 }
--- a/src/bootstrap/CommandBootstrap.ts
+++ b/src/bootstrap/CommandBootstrap.ts
@@ -1,5 +1,5 @@
 import process from 'process';
-import * as fs from 'fs';
+import fs from 'fs';
 import CommandPolykey from '../CommandPolykey';
 import * as binUtils from '../utils';
 import * as binOptions from '../utils/options';
@@ -53,7 +53,7 @@ class CommandBootstrap extends CommandPolykey {
         await fs.promises.writeFile(
           options.recoveryCodeOutFile,
           recoveryCodeOut,
-        );      
+        );
       }
     });
   }
--- a/src/utils/options.ts
+++ b/src/utils/options.ts
@@ -105,6 +105,11 @@ const recoveryCodeFile = new commander.Option(
   'Path to Recovery Code',
 );
 
+const recoveryCodeOutFile = new commander.Option(
+  '-rcof, --recovery-code-out-file <path>',
+  'Path to output recovery code. Only used if a `RecoveryCode` is generated.',
+);
+
 const background = new commander.Option(
   '-b, --background',
   'Starts the agent as a background process',
@@ -240,6 +245,7 @@ export {
   agentPort,
   connConnectTime,
   recoveryCodeFile,
+  recoveryCodeOutFile,
   passwordFile,
   passwordNewFile,
   background,
--- a/src/utils/options.ts
+++ b/src/utils/options.ts
@@ -102,12 +102,12 @@ const passwordNewFile = new commander.Option(
 
 const recoveryCodeFile = new commander.Option(
   '-rcf, --recovery-code-file <path>',
-  'Path to Recovery Code',
+  'Path to a file used to load the Recovery Code from',
 );
 
 const recoveryCodeOutFile = new commander.Option(
   '-rcof, --recovery-code-out-file <path>',
-  'Path to output recovery code. Only used if a `RecoveryCode` is generated.',
+  'Path used to write the Recovery Code if one was generated, if none was generated then this is ignored',
 );
 
 const background = new commander.Option(
--- a/flake.lock
+++ b/flake.lock
@@ -5,11 +5,11 @@
         "systems": "systems"
       },
       "locked": {
-        "lastModified": 1705309234,
-        "narHash": "sha256-uNRRNRKmJyCRC/8y1RqBkqWBLM034y4qN7EprSdmgyA=",
+        "lastModified": 1710146030,
+        "narHash": "sha256-SZ5L6eA7HJ/nmkzGG7/ISclqe6oZdOZTNoesiInkXPQ=",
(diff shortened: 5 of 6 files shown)


tegefaulkes:
This PR isn't very clear. This is related to the service configuration using `nix`? Should we call it `nix service configuration`? Also, if there are no issues directly fixed by this then just remove that section.

Also, do we have a specification for this anywhere? is #143 related?

CMCDragonkai:
@brynblack please make sure you got an issue in `nixpkgs-matrix` - that should be an epic that governs all the relevant issues of moving into module style. This PR should then be part of that - possibly `Related` or a `Fix` depending of scope. Then spec out implementation wise - and important to tell people how to use it if there's a special flag. And if we need special flag that's not quite good, it's better if things "work by default".

tegefaulkes:
Ok, so after setting up the password file and restarting the service it just works. 

```
❯ systemctl status polykey
● polykey.service - Polykey Agent
     Loaded: loaded (/etc/systemd/system/polykey.service; enabled; preset: enabled)
     Active: active (running) since Wed 2024-03-06 15:53:59 AEDT; 14s ago
   Main PID: 23832 (polykey-agent)
         IP: 29.7K in, 44.2K out
         IO: 23.2M read, 0B written
      Tasks: 78 (limit: 38078)
     Memory: 580.9M (peak: 701.9M)
        CPU: 6.707s
     CGroup: /system.slice/polykey.service
             └─23832 polykey-agent
```

Running agent status works too. so its using the correct default path.

```
❯ sudo npm run polykey -- agent status
[sudo] password for brian: 

> polykey-cli@0.2.3 polykey
> ts-node src/polykey.ts agent status

✔ Please enter the password … ************
status                  LIVE
pid                     23832
nodeId                  vu5gloq1a1abnf07q47vl74seln1haf5leeeck8dah58d198mf750
clientHost              ::1
clientPort              43095
agentHost               ::
agentPort               43627
upTime                  177
connectionsActive       2
nodesTotal              2
version                 0.1.8
sourceVersion           1.2.1-alpha.47
stateVersion            1
networkVersion          1
commitHash              2bcebd05046bbae86a7b9da67e22613720952627
libVersion              1.2.1-alpha.47-1-1
libSourceVersion        1.2.1-alpha.47
libStateVersion         1
libNetworkVersion       1
```

Note, that the service is running in root, so I have to sodo the command to have it get the correct default path. That's not ideal right now.

So there are a few things I'd like to point out.

1. It seems simple enough, the service is added be default on the new platforms changes. But it doesn't run properly without providing a password file in some *magic* path. So I need to create a `pass` file in `/run/keys/polykey` before the service can even run. Ideally we have a safe way of defining secrets and then using config parameters to use them. Is this possible at all?
2. The service itself is running as root. No path is specified so it defaults to `~/.local/share/polykey`. For the root user this is `/root/.local/share/polykey`. From a usability standpoint if we wanted to use this node we'd have to specify nodeid, ip and port directly OR run the command as root with `sudo polykey agent start`. Ideally we'd be able to seamlessly access this node without using root or specifying the path/connection info. Maybe run the service as a user we configure?
3. Can we add a way to configure flags when setting up the service? For example, can we override the `-np --node-path` that is used when running? Specify the recovery code or `--fresh` and the like?
4. Can we configure the user and group the service runs as? I think it's best practice to not run a service as root if at all possible.
5. Can we configure multiple `Polykey` services at once? I know it's possible to have multiple of the same service running under different names and configurations. Things like VPNs need this to run multiple networks with the same program.

tegefaulkes:
Not a problem with the service per se, but the agent still has issues with connections dropping out. 

```
Mar 07 09:35:59 matrix-precision-3480-syzygy polykey[44478]: WARN:polykey.PolykeyAgent:Moving Task v0pf803kstlo0180keei7clko5c from Active to Queued
Mar 07 09:36:01 matrix-precision-3480-syzygy polykey[44478]: (node:44478) [DEP0112] DeprecationWarning: Socket.prototype._handle is deprecated
Mar 07 09:36:01 matrix-precision-3480-syzygy polykey[44478]: (Use `node --trace-deprecation ...` to show where the warning was created)
Mar 07 09:36:01 matrix-precision-3480-syzygy polykey[44478]: WARN:polykey.PolykeyAgent.NodeManager:Duplicate refreshBucket task was found for bucket 255, cancelling
Mar 07 09:36:01 matrix-precision-3480-syzygy polykey[44478]: WARN:polykey.PolykeyAgent.NodeManager:Duplicate refreshBucket task was found for bucket 254, cancelling
Mar 07 09:36:01 matrix-precision-3480-syzygy polykey[44478]: pid               44478
Mar 07 09:36:01 matrix-precision-3480-syzygy polykey[44478]: nodeId            vu5gloq1a1abnf07q47vl74seln1haf5leeeck8dah58d198mf750
Mar 07 09:36:01 matrix-precision-3480-syzygy polykey[44478]: clientHost        ::1
Mar 07 09:36:01 matrix-precision-3480-syzygy polykey[44478]: clientPort        41049
Mar 07 09:36:01 matrix-precision-3480-syzygy polykey[44478]: agentHost         ::
Mar 07 09:36:01 matrix-precision-3480-syzygy polykey[44478]: agentPort         36307
Mar 07 09:40:12 matrix-precision-3480-syzygy polykey[44478]: ErrorQUICClientInternal: Failed to send data on the QUICSocket
Mar 07 09:40:12 matrix-precision-3480-syzygy systemd[1]: polykey.service: Main process exited, code=exited, status=70/SOFTWARE
Mar 07 09:40:12 matrix-precision-3480-syzygy systemd[1]: polykey.service: Failed with result 'exit-code'.
Mar 07 09:40:12 matrix-precision-3480-syzygy systemd[1]: polykey.service: Consumed 35.082s CPU time, 315.7M memory peak, 0B memory swap peak, no IO, received 154.9K IP traffic, sent 270.3K IP traffic.

```

I can trigger this by disabling wifi on the laptop.

I thought I had fixed this in quic where send failures are thrown back to the connection making the send. The error codes it handles this way is very focused so it may be a new error code which is getting bubbled up to an internal error in stead. I'll need to look into this.

CMCDragonkai:
This needs to be rock solid. We are going live for the next application.

6 Mar 2024 16:51:16 Brian Botha ***@***.***>:

brynblack:
Some things that were discussed in a meeting:
- Functionality should be added to Polykey itself to prevent "jankiness".
- There should be a flag to divert from standard output to a file.
- Create an issue for the recovery code flag input/output.
- Instead of setting things through flags, create a config format in place of flags (JSON format).
- The file umask should be 0077, the only thing that needs to be set is the `status.json` file to 0007 umask (0770 permissions).

tegefaulkes:
Just to clarify

`There should be a flag to divert from standard output to a file.` - There should be a flag to divert sensitive information such as the recovery code from first time startup to a file. I think the recovery code is the only thing that needs this but any other sensitive information would get the same treatment.

`Instead of setting things through flags, create a config format in place of flags (JSON format).` still up for discussion before creating an issue. We'd need to discuss if it's really needed and how it would work first.

src/CommandPolykey.ts
```diff
@@ -46,6 +46,7 @@ class CommandPolykey extends commander.Command {
     this.addOption(binOptions.passwordFile);
     this.addOption(binOptions.format);
     this.addOption(binOptions.verbose);
+    this.addOption(binOptions.divertSensitiveFile);
   }
```
tegefaulkes: This isn't the place to add `divertSensitiveFile` as an option. Any options added here are global to all commands. You need to add it to commands that specifically need it. That being the bootstrap and start command.

src/utils/options.ts
```diff
@@ -105,6 +105,11 @@ const recoveryCodeFile = new commander.Option(
   'Path to Recovery Code',
 );
 
+const divertSensitiveFile = new commander.Option(
+  '-dsf, --divert-sensitive-file <path>',
+  'Path to divert sensitive information',
+);
+
```
tegefaulkes: This is too generic in name and description. It's only used for outputting the `RecoveryCode` to a file. So it should be named as such. Also the description needs to explain that it is only used if a `RecoveryCode` is generated.

brynblack:
I think a configuration file wouldn't hurt to have, although it would probably end up mainly being used for NixOS systems. At the moment, the way I have it set up requires the end-user to manually place a file called `password` or `recovery-code` into the configuration directory designated by the module configuration. But with more and more features/inputs being required, we should consolidate this information into one file most likely.

As you said, JSON would be appropriate here as we already heavily use JSON, and our tooling would work a lot better around it. Maybe how it could work is Polykey will scan a default directory for the configuration file (perhaps `.config/polykey/config.json` and then parse the results into related values. This would skip the need to always provide CLI flags and is heavily useful in declarative configurations. The user can also provide an option to specify an alternative path to the configuration file.

Only concern I have is how we would propagate things like the recovery code for example. By default, that recovery code will be outputted as a file called `recovery-code` into the configuration directory, but should we leave it this way? Not sure. Is it a good idea to even be putting sensitive information into a configuration file like that in the user's own configuration directory?

src/utils/options.ts
brynblack: Only concern I have here is there already exists an option named `recoveryCodeFile` so I can't really reuse that name.

CMCDragonkai:
No we won't have a config file. It's bad UX.

14 Mar 2024 00:52:38 Brynley Llewellyn-Roux ***@***.***>:

flake.nix
```diff
@@ -16,7 +16,133 @@
...
+            recovery = mkEnableOption "Enable recovery mode for the node. Will try to source from a file under the configuration directory called `recovery-code`.";
+
+            configPath = mkOption {
+              type = with types; uniq str;
+              default = "/etc/polykey";
+              description = ''
+              The path to Polykey configuration directory.
+              A file called `password` with your desired password must be placed inside of this directory to set the node password.
+              '';
+            };
```
tegefaulkes: How are we handling permissions for this password file? It's possible that the user makes this file to permissive which would be problematic. SSH handles this by refusing to use a file depending on it's mask. Can we do the same here?

flake.nix
```diff
@@ -16,7 +16,133 @@
...
     in
+    {
+      nixosModules.default = { config, ... }: with nixpkgs; with lib;
+      {
+        options = {
+          services.polykey = {
+            enable = mkEnableOption "Enable the Polykey agent. Users with the `polykey` group or root permissions will be able to manage the agent.";
+
+            recovery = mkEnableOption "Enable recovery mode for the node. Will try to source from a file under the configuration directory called `recovery-code`.";
+
```
tegefaulkes: This still feels like a jank way to enable use of the recovery code. Did you explore any alternatives?

flake.nix
```diff
@@ -16,7 +16,133 @@
...
+                  bootstrap  ${lib.optionalString config.services.polykey.recovery '' -rcf ${config.services.polykey.configPath}/recovery-code''}\
+                  -rcof ${config.services.polykey.configPath}/recovery-code
+                '';
+                ExecStart = ''
+                  ${self.outputs.packages.${buildSystem}.default}/bin/polykey \
+                  -pf ''${CREDENTIALS_DIRECTORY}/password \
+                  -np ${config.services.polykey.statePath} \
+                  agent start \
+                  -rcof ${config.services.polykey.configPath}/recovery-code
+                '';
```
tegefaulkes: Two things, 

1. Is the `-` in front of `-${self.outputs.packages.${buildSystem}.default}/bin/polykey` meant to be there?
2. Us the full sized option name of `--password-file` and `--node-path` rather than the shortened `-pf` or `-np`. It's better to be explicit here. Make sure all options and flags use the full names.

src/utils/options.ts
```diff
@@ -209,6 +214,7 @@ export {
   agentPort,
   connConnectTime,
   recoveryCodeFile,
+  recoveryCodeOutFile as recoveryCodeOutFile,
   passwordFile,
```
tegefaulkes: What's with the `as` assignment here?

tegefaulkes (commented):
Some minor things to check.

flake.nix
brynblack: 1. The `-` tells Systemd to ignore if the command fails, so if the bootstrap command fails (e.g. the state directory already exists; if the user reboots the service) it won't pull down the entire service with it, and it will continue with executing the `ExecStart` command.
2. That sounds good, I'll implement that now.

flake.nix
brynblack: There aren't many alternatives here, as we cannot directly pass in parameters when rebuilding (flakes are self-contained). I need to explore more options, I'll write up any I find here that would be a suitable alternative.

src/utils/options.ts
brynblack: That's strange, I'm not sure why that's there. I'll look into it.

flake.nix
brynblack: Currently, its not handling permissions in an ideal manner, it will just attempt to read the password file regardless of permissions. I believe its possible to cause the module to throw an error if a condition is met, so it should be possible to check the mask and throw an error if it isn't suitable. I'll get this done ASAP.

src/utils/options.ts
brynblack: It seems it might've been my LSP decided to rename the variable to itself, that's very odd. Fixed in the next commit I'll be pushing up soon.

flake.nix
brynblack: I have an idea, it involves the `systemd-ask-password` command. When the user runs this command, it will ask for a password to use as a credential, and will store it as a credential Systemd is able to read from in service units. We could then copy that password to a more permanent directory to maintain persistence.

brynblack: Hm I think though my previous idea would pertain more to sourcing a password into it, let me rethink this. Maybe we can output to an environment variable that could be read by the user? Or we just automatically output it to a file if it is generated, and then alert the user that it was generated, and they handle it themselves?

brynblack: But there's also the question of telling the agent whether to actually initiate recovery. I suppose if the user sets an environment variable such as `POLYKEY_RECOVERY` it will recover using the configured file path in the module?

flake.nix
brynblack: I think we could add this to the CLI itself? Doing this through scripting seems very hacky.

brynblack: @CMCDragonkai do you think we should make it so PKCLI will refuse to accept any password files as an argument if they don't contain a proper file permission mask?

tegefaulkes: If we have it at all it should be a check inside of `PK-CLI`

flake.nix
tegefaulkes: Maybe a simple script in the unit file that will load a recovery code from a file to an env variable once and delete that file?

brynblack: I'm not too sure about this, already we read from a file. If it were put into an environment variable, we would need to make that password persistent anyway, and under a user-space I feel like it would just make more sense to let the user handle the file themselves, instead of making a file under some sort of system directory that does not have any read access. Maybe in the case of the system service we can do something like what you suggested, but user-level probably not.

brynblack: Okay, I suppose we can make it so that if we are recovering, we check a directory to see if a recovery code file is present; if it is, we recover using that and delete the file that was provided. Let me implement this.

CMCDragonkai:
Let's move ahead. Make proposals for solving everything here and try it out - when in doubt, do both if it is possible and we will be testing it in prod and we will see which one is better. Get this merged this week at the latest.

flake.nix
brynblack: I have made this logic now:
```nix
            recoveryCodePath = "${config.services.polykey.configPath}/recovery-code";
            recoveryCodeExists = builtins.pathExists recoveryCodePath;
```

Basically, if the user provides the file in a set location (in this case in the configuration directory), it will automatically determine if a recovery is needed or not. So if the file exists, we recover, and if it doesn't, we create a new vault altogether. Still not sure though about deleting the recovery code file, especially in the case where the user doesn't provide one (the file will be put in said configuration directory and the user can deal with it).

CMCDragonkai: What do you mean we create a new vault altogether? What does vaults have to do with recovery codes?

flake.nix
CMCDragonkai: I do not like the idea of a PK config directory. It should just be direct options in Nix or references to arbitrary file paths the user has set. We do not use config directories or config files as a rule for PK.

brynblack:
To clarify after a bunch of confusion:

- The defaults existed to make it more "plug-and-play" for the end-user (we know now stuff like the password file path should be required in the Nix module configuration to prevent implicitness and quiet errors)
- The new spec will require the user to provide the password file, and fail with some error if it is not set
- With the nodePath and the recovery code file, those could be implicit or only take effect if the user manually sets them; if the CLI already has defaults, we should follow that methodology and use the same? And with the recovery code, if the path to that file is set then it starts recovery? I need input here from everyone.

flake.nix
brynblack: Resolved in latest commit, it will now only recover if the user sets the recovery code property themselves.

tegefaulkes:
I'm going to do another review of this in a sec. Since I'm not directly working on this I usually don't have a full picture of all the decisions made and what the final constraints are. It's clear that what we thought was needed has evolved over time and we've made some changes to that. But I don't think you've updated the spec in #143 so I don't have a clear reference to todo to compare against for working out if this is done or not.

Remember that we refer to the spec while implementing, reviewing and even much later when writing up the R&D report. So it needs to be updated as the constraints change.

flake.nix
```diff
@@ -13,7 +13,153 @@
...
+        options = {
+          services.polykey = {
+            enable = mkEnableOption "Enable the Polykey agent. Users with the `polykey` group or root permissions will be able to manage the agent.";
+
+            passwordFilePath = mkOption {
+              type = with types; uniq str;
+              default = "/this/path/will/cause/a/failure/if/not/set";
+              description = ''
+              The path to the Polykey password file. This is required to be set for the module to work, otherwise this module will fail.
+              '';
```
tegefaulkes: Is `mkOption` the best thing to use here? It's not optional. The `default` of `/this/path/will/cause/a/failure/if/not/set` is a bit of a code smell too.

brynblack:
> But I don't think you've updated the spec in #143 so I don't have a clear reference to todo to compare against for working out if this is done or not.

I want to just add on to this, there never really was a specification issue per say pertaining to the entire specification of this, issue #143 was more related to the secret zero problem itself. I have instead mostly been using this PR as the specification.

brynblack:
Something else to note I noticed CI has been failing, something to do with the identities and discovering gestalts by node. Investigating.

CMCDragonkai:
@brynblack can you update the spec - and describe at a high level what this will do - as well as provide excalidraw diagram.

flake.nix
```diff
@@ -13,7 +13,153 @@
...
+          programs.polykey = {
+            enable = mkEnableOption "Enable the per-user Polykey agent.";
+
+            passwordFilePath = mkOption {
+              type = with types; uniq str;
+              default = "/this/path/will/cause/a/failure/if/not/set";
+              description = ''
+              The path to the Polykey password file. This is required to be set for the module to work, otherwise this module will fail.
+              '';
+            };
```
tegefaulkes: same here.

tegefaulkes (commented):
Using `mkOptional` for required fields seems weird. Is there any alternative?

Otherwise it looks fine to my un-trained eye. So long as everything works with manual testing then I think we're good.

flake.nix
brynblack: My bad, this was left in from a previous experiment. Fixing both.

brynblack:
> Using `mkOptional` for required fields seems weird. Is there any alternative?

Where exactly is this? Are you referring to `mkOption`? or `lib.optionalString`?

tegefaulkes (approved):
Ok, if it works and you think it's good then I think we're good.

amydevs (approved):

src/agent/CommandStart.ts
```diff
@@ -12,6 +12,7 @@ import type { DeepPartial } from 'polykey/dist/types';
 import type { RecoveryCode } from 'polykey/dist/keys/types';
 import childProcess from 'child_process';
 import process from 'process';
+import * as fs from 'fs';
```
CMCDragonkai: This is unnecessary.

CMCDragonkai: I think you can do `import fs from 'fs';`.

CMCDragonkai: On the other hand I'm pretty sure there's a consistent FS API somewhere. Are we making use of fa injection from the utils @tegefaulkes?

src/utils/options.ts
```diff
@@ -240,6 +245,7 @@ export {
   agentPort,
   connConnectTime,
   recoveryCodeFile,
+  recoveryCodeOutFile,
```
CMCDragonkai, line 248: What would be the problem with re-usung the original option `recoveryCodeFile`? Is there fundamental ambiguity here?

src/bootstrap/CommandBootstrap.ts
```diff
@@ -1,4 +1,5 @@
 import process from 'process';
+import * as fs from 'fs';
```
CMCDragonkai: Same here.

src/utils/options.ts
brynblack: We don't want to reuse the input recovery code file path, as that means it may try to overwrite the existing recovery code file. Additionally, the user may want to customise the output directory so it makes sense to keep these separate.

tegefaulkes: Its better that the difference is explicit. While yes, we can't have an input and an output function at the same time, How would we tell the input apart from output? Check that we're imputing a path rather than a BIP39 code? Switching that way is way to implicit for my liking.

CMCDragonkai: If one were to set `--recovery-code-file=/path/to/somewhere`.

Then it could mean that the program is going to read from that path.

It could meant that the program will write to the path.

In what situation where it would ever be the case where you'd want to read from path A, and then write to path B?

As far as I know atm, in every situation where you'd want to read a recovery code, or write to the recovery code, you'd want to have the same path. In fact it makes sense as the option that designates the location of the recovery code.

However disregard what I said **IF** the presence of the option `--recovery-code-file` actually changes the behaviour of polykey agent start. If this is true, such as the case where it tries to start by reading the recovery code... then maybe it should be 2 options.

If there are 2 options - it has to be documented clearly which option is actually relevant. I can imagine ambiguity here:

```
# is this about starting an agent, recovery with A, and then writing it out to B?
polykey agent start --recovery-code-file=/path/to/A --recovery-code-out-file=/path/to/B

# is this about writing out to B a generated recovery code? (What if the key already existed?)
polykey agent start --recovery-code-out-file=/path/to/B

# what about this?
polykey agent start --recovery-code-file=/path/to/B
```

I think we have to make a decision here. When an option is not used, should we throw an error to indicate to the user that the option does not make sense here? Or should we just ignore the option - and we could even warn on stderr that the option is not being used - due to _reasons_. I'm leaning to the latter atm.

CMCDragonkai:
Needs to be made ready to merge @tegefaulkes.

CMCDragonkai:
This must start being used now. I'm reviewing the implementation along with the new platform profiles in our Orchestrator.

tegefaulkes:
Working on this now. It just needs review comments addressed and then merge?

src/agent/CommandStart.ts
tegefaulkes: I'm not exactly sure what you mean? I think we have an interface we use when a fs needs to be passed into something, but for the top level like this we just import `fs`.

src/utils/options.ts
tegefaulkes: I'll make sure that they are nicely documented. The `recovery-code-out-file` option by necessity only applies when needed. When not needed it is just ignored. For it's usage there are two cases

1. node generated a new recovery code for whatever reason - so rather than printing it out we divert the recovery code to a file specified by that path.
2. Node started with existing state - Do nothing, there was no recovery code to output. 

For case 2, I'd rather just have it silently do nothing. The service will have a static configuration. So first time it starts it needs to not output the code to the logs, but for all subsequent starts we don't need to handle it. But in both the initial and subsequent starts the `recovery-code-out-file` will be set.


This option needs to be separate from the `recovery-code-file` since their intentions *will* conflict depending on state. Starting a node without state will need a clear difference between inputting a recovery code, or outputting it to a file.

src/agent/CommandStart.ts
tegefaulkes: This has been fixed.

src/bootstrap/CommandBootstrap.ts
tegefaulkes: This has been fixed.

src/utils/options.ts
tegefaulkes: I've updated the description to indicate that one is for reading the recovery code input, and the other is only used to divert the recovery code output.

tegefaulkes:
Rebased on staging.

src/agent/CommandStart.ts
CMCDragonkai: I think I was trying to avoid nodeisms, but since this is PK CLI, that doesn't really matter.

CMCDragonkai:
@CryptoTotalWar the above spec would need to be adapted for the Polykey-Docs too. It would come under deployment usage or service usage.

CryptoTotalWar:
> @CryptoTotalWar the above spec would need to be adapted for the Polykey-Docs too. It would come under deployment usage or service usage.

@CMCDragonkai Polykey-Docs Issue created to track documentation request. https://github.com/MatrixAI/Polykey-Docs/issues/40

PR #150, opened 2024-03-01
Fix `identities authenticate` crash

### LINKED ISSUE #76: `identities authenticate` fails without a desktop enviroment
### Describe the bug

Running `polykey identities authenticate` in an environment that is missing a desktop environment or a web browser fails.

I suppose this somewhat intended since the OAUTH procedure depends on a web browser. But the error throw is very low level in this case. That said I don't think we have a method of adding an identity without going through OAUTH. So how do we add identities to nodes in a server environment.

I think we can to the browser portion of the procedure on a different computer, so the only issue here is that the command is crashing while failing to spawn the browser.

### To Reproduce

1. In an enviroment that lacks the desktop such as an Ubuntu server, run the `polykey identities authenticate command` This will throw `Error: spawn xdg-open ENOENT`.

### Expected behaviour

If we fail to spawn a web browser then just proceed as normal. Don't crash the command.

### Screenshots

image

### Platform (please complete the following information)
 - Device: Ubuntu VM running in an Unraid NAS
 - OS: Ubuntu 22.04

tegefaulkes:
### Description

This PR addresses crashing when using `identities authenticate` if it fails to open a browser.

### Issues Fixed

* Fixes: #76 

### Tasks

- [X] 1. Transplant `Browser` utility function from `Polykey` into `Polykey-CLI`.
- [X] 2. Fix `Browser` throw when failing to spawn browser rather than crashing the process.

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/npmDepsHash
+++ b/npmDepsHash
@@ -1,1 +1,1 @@
-sha256-/nogmxBBx/8+zSgaz/2N0yByJ30P9TnCTF0Dx69N4Qo=
+sha256-S68GU9VhfUQ5+iBWuGMoZrLCpAiyda3kvdR8c80QPL8=
--- a/package-lock.json
+++ b/package-lock.json
@@ -41,7 +41,7 @@
         "nexpect": "^0.6.0",
         "node-gyp-build": "^4.4.0",
         "nodemon": "^3.0.1",
-        "polykey": "^1.2.1-alpha.49",
+        "polykey": "^1.2.1-alpha.50",
         "prettier": "^3.0.0",
         "shelljs": "^0.8.5",
         "shx": "^0.3.4",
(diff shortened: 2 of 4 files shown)


CMCDragonkai:
I'm not sure it makes sense to have this in the CLI. The whole point is to share the functionality between CLI, desktop and mobile. This the library can provide the necessary platform detection.

tegefaulkes:
I can apply the fix to `Polykey` then.

tegefaulkes:
Alright, I've fixed the `browser` utility function in `Polykey` and using that now.

PR #178, opened 2024-04-26
Creating audit domain

### LINKED ISSUE #177: Audit CLI commands for inspecting audit event history
### Specification

The Audit domain was added to the `Polykey` library and is being used for the `polykey` website dashboard. But we haven't yet exposed it to the CLI. It will be useful for auditing connection and discovery events. But also any other types of events when they are added into the audit domain.

The audit command will be generic to handling all kinds of audit events. Event types to iterate over will be selected by the event path. We should allow the ability to specify multiple paths as a varidic in a dot path format of `a.b.c`. It will be possible to specify a subpath, parent path or duplicate paths at the same time so we'll need to trim the list to the minimal relevant paths. A union of a duplicate is the same path, A union of a child and parent path is just the parent path.

When iterating over the multiple paths we're going to use a iterator zipper method. Where each path has it's own iterator. Each select from each iterator to be read from based on the strict ordering of the `AuditEventId` which is a monotonic logical ordering based on time.

Further features and formatting needs to be specced out.

### Additional context

* Related: #162 - Started creating the audit domain but that work is being split to a new PR.

### Tasks

1. expand this spec.

tegefaulkes:
### Description

This PR addresses adding the Audit domain to the CLI.

I have some work in progress done, but it needs to be made more generic. The specific `audit discovery` command needs to be generalised to a `audit` command that handles all audit events.

We may need to add formatters for specific event types. For example the discovery events have verticies as the `GestaltIdEncoded`. It may be fine to output that as is. But ideally we'd keep gestaltId formatting consistent, so a node would be   `NodeIdEncoded` and identitiy as `identitiyId:providerId`.

### Issues Fixed

* Fixes #177 

### Tasks

- [x] 1. Create a generic `polykey audit` command.
- [x] 2. Support providing multiple event paths in a dot path format `a.b.c`.
- [x] 3. Create tests checking functionality.

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/npmDepsHash
+++ b/npmDepsHash
@@ -1,1 +1,1 @@
-sha256-aSQGZ+PTzc0GCs8SvMw6fnvx6BN20/l3C0h2ox3s7Mg=
+sha256-mto5sb8+L5pb0ZXGruJlNmPz3XU1fOSF14mh5VHoW7I=
--- a/npmDepsHash
+++ b/npmDepsHash
@@ -1,1 +1,1 @@
-sha256-mto5sb8+L5pb0ZXGruJlNmPz3XU1fOSF14mh5VHoW7I=
+sha256-3x+Xu52V6lAVNJ+DP0lcjz43/EHQs4IWuWxLhrs5ruA=
--- a/package-lock.json
+++ b/package-lock.json
@@ -41,7 +41,7 @@
         "nexpect": "^0.6.0",
         "node-gyp-build": "^4.4.0",
         "nodemon": "^3.0.1",
-        "polykey": "^1.6.1",
+        "polykey": "^1.6.3",
         "prettier": "^3.0.0",
         "shelljs": "^0.8.5",
         "shx": "^0.3.4",
@@ -1541,9 +1541,9 @@
       ]
     },
     "node_modules/@matrixai/quic": {
-      "version": "1.2.7",
-      "resolved": "https://registry.npmjs.org/@matrixai/quic/-/quic-1.2.7.tgz",
-      "integrity": "sha512-e2RH/xbW4XzoaAKLFgSYmMFW164gVbAFytTYnWtUxS2kY+eJtErFHpvYS1aWfV6LUV2mJLmI7yBfE4IuDUInew==",
+      "version": "1.2.8",
+      "resolved": "https://registry.npmjs.org/@matrixai/quic/-/quic-1.2.8.tgz",
+      "integrity": "sha512-hGowHtyyu4yaE/L5oy0yOseRllCWX+qi93XKRAbxPgQQXvqEzWk3huGMOGRdf13OdI0ThaTZGAhRT0603oIe3w==",
       "dev": true,
       "dependencies": {
         "@matrixai/async-cancellable": "^1.1.1",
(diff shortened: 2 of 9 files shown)


tegefaulkes:
I've done most of the work for this already with the `audit discovery` command. This needs to be generalised into a `audit` command. Some polish needs to be applied, mainly supporting different option. 

One main feature will need to be implemented as well. We need to support selecting multiple event paths at once using a dot path format of `a.b`, `a.b.c` etc. This will be provided as a varadic option. Any paths that are duplicate or a subset of parent paths need to be removed from the list before iterating over them. 

Simplifying the list is as simple as sorting the paths and removing and duplicate paths or paths that are strictly a sub path. When sorted only the subsequent paths need to be compared.

Once we have a simplified subset of the paths we can create an iterator for each path. Then combine the output of these iterators into a single iterator by comparing the head event of each path and reading them off in the order of the event ID.

tegefaulkes:
I could add the logic for handling multiple event paths to the `audit` command directly. But I wonder if it's better to apply it to the RPC handler instead. The only issue with that is making changes to the RPC command may break the dashboard. I'll have to co-ordinate that with @amydevs .

amydevs:
hmm i think multiple paths might be something to consider to add to the actual Audit class rather than just the CLI

tegefaulkes:
I agree, there is too much surface area for problems if the logic was implemented this side of the RPC. The core of the logic can be implemented either side so I'm doing some basic prototyping of it here first before transplanting it to the `audit` domain.

CMCDragonkai:
If this is going to take longer than this cycle, prefer addressing this https://github.com/MatrixAI/Polykey-CLI/issues/181. That directly affects the flow as shown in our demo video.

This happened literally on Friday on the latest staging branch. So it's definitely a problem.

tegefaulkes:
As part of working on this I'm making changes to the auditEventsGet.ts handler in the `Polykey` lib. The change is to allow selecting multiple paths when getting events. The problem with this is it will very likely be a breaking change for the dashboard that uses it. It would also be a major version change for the lib.

It's a lot more hassle than it's worth TBH. So as an alternative I can make a separate RPC command for using multiple audit paths when getting audit events. This would make it a minor version change and avoid having to fix up the dashboard.

tegefaulkes:
This is ready for review now.

tegefaulkes:
For the output I have it formatted as the following.

```
    >	
      id  	06659319332f7000bda6ed4d0f737d12
      path	discovery.vertex.queued
      data	
        vertex	node-A
    >	
      id  	0665931933337000a7059459c3f6ebfd
      path	discovery.vertex.queued
      data	
        vertex	node-B
        parent	node-A
    >	
      id  	066593193337700094c8b217c9a324db
      path	discovery.vertex.queued
      data	
        vertex	node-C
        parent	node-A
    >	
      id  	06659319333b7001a29fbc5ff3008e4a
      path	discovery.vertex.queued
      data	
        vertex	node-D
        parent	node-B
```

I find having a semi-empty space between each entry is much easier to read. However I think the standard approach is to have it like this?

```
    id  	0665931f14c470008512455a5736c54a
    path	discovery.vertex.queued
    data	
      vertex	node-A
    id  	0665931f14d070008fdc80ad92a46dd0
    path	discovery.vertex.queued
    data	
      vertex	node-B
      parent	node-A
    id  	0665931f14d0700189600c8ed5c8a8b5
    path	discovery.vertex.queued
    data	
      vertex	node-C
      parent	node-A
    id  	0665931f14d47001ae9bead8f8915880
    path	discovery.vertex.queued
    data	
      vertex	node-D
      parent	node-B
```

Which I personally really dislike. But I'll change it if that's the intended format.

tegefaulkes:
I also need to note that this PR is blocked by infrastructure prep work that @brynblack should be working on. I'll need feedback from her on that progress.

CMCDragonkai:
If blocked, you should finish off the other sanity checks for our UX fixes. Brynley will be working on deployment of PKE too - so lots of putting orchestrator into production!

tegefaulkes:
The `Polykey` PR has some details for what needs to be done. Here are some links for reference @amydevs @brynblack 

https://github.com/MatrixAI/Polykey/pull/730 - Generally relevant.
https://github.com/MatrixAI/Polykey/pull/730#issuecomment-2138515041 - Main details for what needs to be done.

tegefaulkes:
This is fully prepped now.

PR #194, opened 2024-06-05
fix deadlock error when pipe ends while prompting password

### LINKED ISSUE #139: Undefined behaviour when piping
### Describe the bug

So it's possible to use the pipe operation to input into a `polykey` command. For example, `echo password | npm run polykey -- agent start`. This will actually input the password on the first prompt.

```
// Start an agent normally
brian@matrix-precision-3480-syzygy:~/workspace/Polykey-CLI/ > npm run polykey -- agent start -np ./tmp/test1 

> polykey-cli@0.1.8 polykey
> ts-node src/polykey.ts agent start -np ./tmp/test1

✔ Enter new password … ******** // Password entered manually
✔ Confirm new password … ********
(node:1189978) [DEP0112] DeprecationWarning: Socket.prototype._handle is deprecated
(Use `node --trace-deprecation ...` to show where the warning was created)
pid             1189978
nodeId          v9q4nj2j4sqipj9nlm197fsfofvn61pcvghreml5cjpiv748o6960
clientHost      ::1
clientPort      43369
agentHost       ::
agentPort       41059
recoveryCode    celery grit genre truly weapon tilt cause plug donate forward tackle junior sphere cave sun cloth assist catalog risk bomb staff rate goddess poverty
^C

// With state created we run again while piping in the password
brian@matrix-precision-3480-syzygy:~/workspace/Polykey-CLI/ > echo "password" | npm run polykey -- agent start -np ./tmp/test1

> polykey-cli@0.1.8 polykey
> ts-node src/polykey.ts agent start -np ./tmp/test1

✔ Please enter the password … ******** // Password is provided by the pipe
(node:1190120) [DEP0112] DeprecationWarning: Socket.prototype._handle is deprecated
(Use `node --trace-deprecation ...` to show where the warning was created)
pid             1190120
nodeId          v9q4nj2j4sqipj9nlm197fsfofvn61pcvghreml5cjpiv748o6960
clientHost      ::1
clientPort      46567
agentHost       ::
agentPort       37515
^C

```

If we try and start `Polykey` using a pipe and starting from scratch we end up with a promise deadlock.

```
brian@matrix-precision-3480-syzygy:~/workspace/Polykey-CLI/ > echo "password" | npm run polykey -- agent start -np ./tmp/test3

> polykey-cli@0.1.8 polykey
> ts-node src/polykey.ts agent start -np ./tmp/test3

✔ Enter new password … ********
? Confirm new password › ErrorPolykeyCLIAsynchronousDeadlock
```

Trying with a newline has the same problem

```
brian@matrix-precision-3480-syzygy:~/workspace/Polykey-CLI/ > echo "password\npassword" | npm run polykey -- agent start -np ./tmp/test3

> polykey-cli@0.1.8 polykey
> ts-node src/polykey.ts agent start -np ./tmp/test3

✔ Enter new password … ****************

? Confirm new password › ErrorPolykeyCLIAsynchronousDeadlock

```

My guess is that the pipe closing is causing the problem here. 

### To Reproduce

1. start polykey with `echo password | agent start` from fresh state
2. polykey shold throw an `ErrorPolykeyCLIAsynchronousDeadlock` error.

### Expected behavior

Hard to say, maybe it should input the first password no problem, but still wait for the confirmation input? Is it also possible to include newlines in the password like this? Do we want to support that? Do we need to sanitise password inputs?

### Platform (please complete the following information)
 - Device: precision laptop 
 - OS: nixos
 - Version 23

### Notify maintainers
@tegefaulkes

tegefaulkes:
### Description

This PR addresses the promise deadlock error when piping into the `agent start` command with fresh state.

I've applied a fix, however some tests are failing now with new behaviour. So I'll need to look deeper into this. It may just be `pkExpect` breaking but some concurrent X tests are failing too.

### Issues Fixed

* Fixes #139 

### Tasks

- [x] 1. Update the prompt utility functions to return undefined when the `stdin` ends while awaiting the prompt.
- [x] 2. Make sure all tests are working

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased


tegefaulkes:
Minor logic bug. It's fixed now. Prepping for merge.

tegefaulkes:
This is the simplest and best solution I can think of right now. The core of the problem is that the `process.stdin` functions differently when piping into the node program. In such a way that `prompts` doesn't handle it properly.

Prompts fails to process this in two ways 
1. A prompt takes a line as an input, but with the pipe wit will take the full stream as the first line. So if we piped `password\npassword`, it gets `passwordpassword` in the first prompt. 
2. It can't handle the end of stream properly. So once the pipe has fully dumped it's data it keeps waiting for input. Resulting in a promise deadlock.

Initially I tried to solve this by creating a promise to track the end of the `stdin` stream. I could race the prompt with this to prevent the deadlock. but this results with a leaked prompts promise. Also the input from this isn't usable anyway.

The simplest solution is to not prompt at all when handling a pipe. The pipe still works, we just can't function interactively while doing it.

I looked into using the `ttys` library https://github.com/TooTallNate/ttys/blob/master/index.js. With this we can have the pipe but also the prompt still functions interactively. The library is simple so we can inline the code. However I'm not sure if it's very portable since it's trying to open a `/dev/tty` with `fs.openSync('/dev/tty', 'r')`.

The only other solution I can think of right now is to detect the piping and have our own stream processor to extract the provided password. Similar to https://stackoverflow.com/questions/59308117/how-to-differentiate-between-standard-inputs-via-pipe-and-via-prompt-in-nodejs

tegefaulkes:
I'm putting this on hold for now pending some discussion.

CMCDragonkai:
Did you synthesize the solution from https://www.npmjs.com/package/ttys?

tegefaulkes:
We're going to go with the current behaviour I've implemented. Where we handle the pipe as if the user cancelled the password prompt.

tegefaulkes:
I had to remove the tests using `nExpect` to check for prompting. It seems to be using a pipe to function so it's incompatible with the fix.

tegefaulkes:
Ready to merge.

Issue #245, opened 2024-07-11
Implement `secrets ls` command


tegefaulkes:
### Specification

The `secrets ls` command will list the contents of a directory.

There are 3 main features to support here.
1. List directory normally
2. List in list format `-l`
3. Include hidden files with `-a`
4. Paths can support wildcards and globing.
5. Multiple paths can be specified.
6. `stdin` is ignored

```
Polykey/tmp/asd on  staging [?] 
❯ ls
asdasd  best  dir1  dir2  one  test  two

Polykey/tmp/asd on  staging [?] 
❯ ls -l
total 32
-rw-r--r-- 1 brian operators  8 Jul 11 11:18 asdasd
-rw-r--r-- 1 brian operators 15 Jul 11 11:18 best
drwxr-xr-x 2 brian operators  2 Jul 11 11:19 dir1
drwxr-xr-x 2 brian operators  2 Jul 11 11:19 dir2
-rw-r--r-- 1 brian operators  0 Jul 11 11:18 one
-rw-r--r-- 1 brian operators  5 Jul 11 11:18 test
-rw-r--r-- 1 brian operators  0 Jul 11 11:18 two

Polykey/tmp/asd on  staging [?] 
❯ ls -a
.  ..  asdasd  best  dir1  dir2  .hidden  one  test  two

Polykey/tmp/asd on  staging [?] 
❯ ls *est
best  test

Polykey/tmp/asd on  staging [?] 
❯ ls **/*     
asdasd  best  dir1/a  dir2/b  dir2/c  one  test  two

dir1:
a

dir2:
b  c

Polykey/tmp/asd on  staging [?] 
❯ ls -l **/*
-rw-r--r-- 1 brian operators  8 Jul 11 11:18 asdasd
-rw-r--r-- 1 brian operators 15 Jul 11 11:18 best
-rw-r--r-- 1 brian operators  0 Jul 11 11:42 dir1/a
-rw-r--r-- 1 brian operators  0 Jul 11 11:43 dir2/b
-rw-r--r-- 1 brian operators  0 Jul 11 11:43 dir2/c
-rw-r--r-- 1 brian operators  0 Jul 11 11:18 one
-rw-r--r-- 1 brian operators  5 Jul 11 11:18 test
-rw-r--r-- 1 brian operators  0 Jul 11 11:18 two

dir1:
total 1
-rw-r--r-- 1 brian operators 0 Jul 11 11:42 a

dir2:
total 1
[...]

Issue #251, opened 2024-08-06
Vault names can have `.` character when creating them, but they aren't valid paths for secrets commands


CMCDragonkai:
### Describe the bug

```
cmcdragonkai ➜ matrix-ml-1  ➜ ~/Projects/Polykey-CLI
 $ ./dist/polykey.js secrets edit 'polykey.com:/POLYKEY_COM_GITHUB_TOKEN'
error: command-argument value 'polykey.com:/POLYKEY_COM_GITHUB_TOKEN' is invalid for argument 'secretPath'. polykey.com:/POLYKEY_COM_GITHUB_TOKEN is not of the format <vaultName>:<directoryPath>
```

Notice that `polykey.com` is a valid vault name, but it cannot be referred to when editing a secret path.

The regex must be aligned between the 2. Also new fast check inputs need to be checked to ensure that all usages of "secret path" or "vault name" are consistent.

### To Reproduce

1. Create `polykey.com` vault
2. Try to edit a file inside it like `polykey.com:/somesecret`

### Expected behavior

Given that we expect that `:` to be the separation, pretty much any alphanum plus basic symbols.

The question is what symbols should be allowed for vault names. We should allow most characters on standard keyboards, and even utf-8 characters. Actually why not allow anything except `:`, since that's the distinguishing character.

### Notify maintainers

@tegefaulkes @aryanjassal

CMCDragonkai:
Consistent secret paths and vault names needs to be specified and tested consistently across all commands in #32.

tegefaulkes:
This may tie into #223 but otherwise should be an easy issue for @aryanjassal to take on.

aryanjassal:
https://github.com/MatrixAI/Polykey-CLI/blob/755e97f0ab6346c04c373adf8c0550da5ce71afb/src/vaults/CommandCreate.ts#L14

This line is responsible for allowing vault creation, as there are no checks for forbidden characters. As such, it isn't a matter of unifying the regex for both vault path during creation or otherwise, but to add a new parser during vault creation.

One solution is to just make a new parser which will use regex to parse the vault name, and then the same regex can be used in other parsers/commands to validate a vault name. However, we already have too many parsers, most of which do things *slightly* differently than the other. (https://github.com/MatrixAI/Polykey-CLI/pull/255#discussion_r1731175246). We need a way to unify (or simplify) the parsers, otherwise our codebase will become polluted with parsers which are basically the same but have slight differences.

[...]

Issue #32, opened 2021-08-30
Incorporate Unix Commands to Secrets Subcommand - `pk secrets *`


CMCDragonkai:
### Specification

Standard unix commands include things like `cp`, `mv` and `ls`.

This allows users to interact with secret vaults as if they were real filesystems. (And they sort of are).

Imagine (IN ORDER OF PRIORITY):

```
# these 2 might be merged into 1 command just `write` is sufficient
pk secrets read vault1:/a/b
pk secrets write v1:/abc

pk secrets cat vault1:/a/b
pk secrets touch vault1:/a/b
pk secrets ls vault1
pk secrets mkdir vault1:/c
pk secrets rm vault1:/a/b
pk secrets mv vault1:/a/b vault1:/a/c
pk secrets cp vault1:/a/b vault1:/a/c
pk secrets ln vault1:/a/b vault1:/d/e

# this one is special
pk secrets env

# this one is also special
pk secrets ed vault1:/a/b

# these are less important
pk secrets head vault1:/a/b
pk secrets tail vault1:/a/b
pk secrets find vault1
pk secrets grep vault1:/a/b
pk secrets sed vault1:/a/b
```

I believe some of these commands were already implemented before in the old PK codebase. You have things like `pk secrets create` and `pk secrets delete`.

```
»» ~/Projects/js-polykey/src/bin/secrets
 ♖ tree .                                                                                                          (client-refactoring) pts/5 14:47:05
.
├── create.ts
├── delete.ts
├── dir.ts
├── edit.ts
├── env.ts
├── get.ts
├── index.ts
├── list.ts
├── mkdir.ts
├── rename.ts
└── update.ts

0 directories, 11 files
```

I believe there's a problem with doing this. We are reinventing the wheel, and we'll never cover all the commands that Unix already has.

This has the benefit of reusing context that developers already know and remember when interacting with a Unix shell.

But how do we do this without having to rewrite all the code? Luckily it seems someone has already done this.

See: https://github.com/shelljs/shelljs

It has implemented all the major Unix shell commands in raw JS.

The only problem that their command source code currently directly imports the native Node fs:

https://github.com/shelljs/shelljs/blob/79ae14d30d7ce4064de05d41c7889885326b6754/src/ls.js#L2

If we would want to use the `shelljs` library, we would need to globally mock the fs as described here: https://github.com/shelljs/shelljs/issues/747#issuecomment-316301410

However that may be dangerous if that leaks into other places of the FS.

[...]

PR #255, opened 2024-08-16
Implementing UNIX-like `secrets ls` command

aryanjassal:
### Description

This PR works on implementing UNIX-like `secrets ls` experience, as if one was `ls`-ing using the `coreutils`'s `ls` command. As described in the issue spec, not all `ls` commands will be supported, only some major and useful ones.

### Issues Fixed

* Relates to #245
* Relates to https://github.com/MatrixAI/Polykey/pull/785
* REF ENG-358

### Tasks

- [x] 1. Add UNIX-like `ls` support
- [x] 2. Confirm file filtering/globbing works
- [x] 3. Add options to render the contents using various flags
- [x] 4. Update all the test

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/src/secrets/CommandList.ts
+++ b/src/secrets/CommandList.ts
@@ -14,7 +14,7 @@ class CommandList extends CommandPolykey {
     this.argument(
       '<directoryPath>',
       'Directory to list files from, specified as <vaultName>[:<path>]',
-      binParsers.parseSecretName,
+      binParsers.parseSecretPathOptional,
     );
     this.addOption(binOptions.nodeId);
     this.addOption(binOptions.clientHost);
--- a/src/secrets/CommandList.ts
+++ b/src/secrets/CommandList.ts
@@ -3,18 +3,23 @@ import CommandPolykey from '../CommandPolykey';
 import * as binUtils from '../utils';
 import * as binOptions from '../utils/options';
 import * as binProcessors from '../utils/processors';
+import * as binParsers from '../utils/parsers';
 
 class CommandList extends CommandPolykey {
   constructor(...args: ConstructorParameters<typeof CommandPolykey>) {
     super(...args);
-    this.name('list');
-    this.aliases(['ls']);
-    this.description('List all Available Secrets for a Vault');
-    this.argument('<vaultName>', 'Name of the vault to list secrets from');
+    this.name('ls');
+    this.aliases(['list']);
+    this.description('List all secrets for a vault within a directory');
+    this.argument(
+      '<directoryPath>',
+      'Directory to list files from, specified as <vaultName>[:<path>]',
+      binParsers.parseSecretName,
+    );
     this.addOption(binOptions.nodeId);
     this.addOption(binOptions.clientHost);
     this.addOption(binOptions.clientPort);
-    this.action(async (vaultName, options) => {
+    this.action(async (vaultPattern, options) => {
       const { default: PolykeyClient } = await import(
         'polykey/dist/PolykeyClient'
       );
@@ -40,39 +45,38 @@ class CommandList extends CommandPolykey {
           nodeId: clientOptions.nodeId,
           host: clientOptions.clientHost,
           port: clientOptions.clientPort,
-          options: {
-            nodePath: options.nodePath,
-          },
+          options: { nodePath: options.nodePath },
           logger: this.logger.getChild(PolykeyClient.name),
         });
-        const data = await binUtils.retryAuthentication(async (auth) => {
-          const data: Array<{ secretName: string }> = [];
+        const secretPaths = await binUtils.retryAuthentication(async (auth) => {
+          const secretPaths: Array<string> = [];
           const stream = await pkClient.rpcClient.methods.vaultsSecretsList({
             metadata: auth,
-            nameOrId: vaultName,
+            nameOrId: vaultPattern[0],
+            secretName: vaultPattern[1] ?? '/',
           });
           for await (const secret of stream) {
-            data.push({
-              secretName: secret.secretName,
-            });
+            // Remove leading slashes
+            if (secret.path.startsWith('/')) {
+              secret.path = secret.path.substring(1);
+            }
+            secretPaths.push(secret.path);
           }
-          return data;
+          return secretPaths;
         }, auth);
 
         if (options.format === 'json') {
           process.stdout.write(
             binUtils.outputFormatter({
               type: 'json',
-              data: data,
+              data: secretPaths,
             }),
           );
         } else {
           process.stdout.write(
             binUtils.outputFormatter({
               type: 'list',
-              data: data.map(
-                (secretsListMessage) => secretsListMessage.secretName,
-              ),
+              data: secretPaths,
             }),
           );
         }
--- a/src/utils/parsers.ts
+++ b/src/utils/parsers.ts
@@ -8,7 +8,8 @@ import * as gestaltsUtils from 'polykey/dist/gestalts/utils';
 import * as networkUtils from 'polykey/dist/network/utils';
 import * as nodesUtils from 'polykey/dist/nodes/utils';
 
-const secretPathRegex = /^([\w-]+)(?::)([^\0\\=]+)$/;
+const vaultNameRegex = /^[\w.-]+$/;
+const secretPathNameRegex = /^([\w-]+)(?::([^\0\\=]+))?$/;
 const secretPathValueRegex = /^([a-zA-Z_][\w]+)?$/;
 const environmentVariableRegex = /^([a-zA-Z_]+[a-zA-Z0-9_]*)?$/;
 
@@ -65,27 +66,47 @@ function parseCoreCount(v: string): number | undefined {
   }
 }
 
-function parseSecretPath(secretPath: string): [string, string, string?] {
-  // E.g. If 'vault1:a/b/c', ['vault1', 'a/b/c'] is returned
-  //      If 'vault1:a/b/c=VARIABLE', ['vault1, 'a/b/c', 'VARIABLE'] is returned
-  const lastEqualIndex = secretPath.lastIndexOf('=');
-  const splitSecretPath =
-    lastEqualIndex === -1
-      ? secretPath
-      : secretPath.substring(0, lastEqualIndex);
-  const value =
-    lastEqualIndex === -1 ? '' : secretPath.substring(lastEqualIndex + 1);
-  if (!secretPathRegex.test(splitSecretPath)) {
+function parseVaultName(vaultName: string): string {
+  // E.g. If 'vault1, 'vault1' is returned
+  //      If 'vault1:a/b/c', an error is thrown
+  if (!vaultNameRegex.test(vaultName)) {
     throw new commander.InvalidArgumentError(
-      `${splitSecretPath} is not of the format <vaultName>:<directoryPath>`,
+      `${vaultName} is not of the format <vaultName>`,
     );
   }
-  const [, vaultName, directoryPath] = splitSecretPath.match(secretPathRegex)!;
-  return [vaultName, directoryPath, value];
+  // Returns match[1], or the parsed vaultName
+  return vaultName.match(secretPathNameRegex)![1];
+}
+
+function parseSecretName(secretPath: string): [string, string?] {
+  // E.g. If 'vault1:a/b/c', ['vault1', 'a/b/c'] is returned
+  //      If 'vault1', ['vault1, undefined] is returned
+  if (!secretPathNameRegex.test(secretPath)) {
+    throw new commander.InvalidArgumentError(
+      `${secretPath} is not of the format <vaultName>[:<directoryPath>]`,
+    );
+  }
+  // Returns [vaultName, secretName?]
+  const match = secretPath.match(secretPathNameRegex)!;
+  return [match[1], match[2] || undefined];
+}
+
+function parseSecretPath(secretPath: string): [string, string, string?] {
+  // E.g. If 'vault1:a/b/c', ['vault1', 'a/b/c'] is returned
+  //      If 'vault1', an error is thrown
+  const [vaultName, secretName] = parseSecretName(secretPath);
+  if (secretName === undefined) {
+    throw new commander.InvalidArgumentError(
+      `${secretPath} is not of the format <vaultName>:<directoryPath>`,
+    );
+  }
+  return [vaultName, secretName];
 }
 
 function parseSecretPathValue(secretPath: string): [string, string, string?] {
-  const [vaultName, directoryPath, value] = parseSecretPath(secretPath);
+  const [vaultName, directoryPath] = parseSecretPath(secretPath);
+  const lastEqualIndex = secretPath.lastIndexOf('=');
+  const value = lastEqualIndex === -1 ? '' : secretPath.substring(lastEqualIndex + 1);
   if (value != null && !secretPathValueRegex.test(value)) {
     throw new commander.InvalidArgumentError(
       `${value} is not a valid value name`,
@@ -198,12 +219,13 @@ function parseEnvArgs(
 }
 
 export {
-  secretPathRegex,
   secretPathValueRegex,
   environmentVariableRegex,
   validateParserToArgParser,
   validateParserToArgListParser,
   parseCoreCount,
+  parseVaultName,
+  parseSecretName,
   parseSecretPath,
   parseSecretPathValue,
   parseSecretPathEnv,
--- a/src/utils/parsers.ts
+++ b/src/utils/parsers.ts
@@ -8,8 +8,7 @@ import * as gestaltsUtils from 'polykey/dist/gestalts/utils';
 import * as networkUtils from 'polykey/dist/network/utils';
 import * as nodesUtils from 'polykey/dist/nodes/utils';
 
-const vaultNameRegex = /^[\w.-]+$/;
-const secretPathNameRegex = /^([\w-]+)(?::([^\0\\=]+))?$/;
+const secretPathRegex = /^([\w-]+)(?::([^\0\\=]+))?$/;
 const secretPathValueRegex = /^([a-zA-Z_][\w]+)?$/;
 const environmentVariableRegex = /^([a-zA-Z_]+[a-zA-Z0-9_]*)?$/;
 
@@ -66,47 +65,44 @@ function parseCoreCount(v: string): number | undefined {
   }
 }
 
-function parseVaultName(vaultName: string): string {
-  // E.g. If 'vault1, 'vault1' is returned
-  //      If 'vault1:a/b/c', an error is thrown
-  if (!vaultNameRegex.test(vaultName)) {
-    throw new commander.InvalidArgumentError(
-      `${vaultName} is not of the format <vaultName>`,
-    );
-  }
-  // Returns match[1], or the parsed vaultName
-  return vaultName.match(secretPathNameRegex)![1];
-}
-
-function parseSecretName(secretPath: string): [string, string?] {
+function parseSecretPathOptional(
+  secretPath: string,
+): [string, string?, string?] {
   // E.g. If 'vault1:a/b/c', ['vault1', 'a/b/c'] is returned
   //      If 'vault1', ['vault1, undefined] is returned
-  if (!secretPathNameRegex.test(secretPath)) {
+  // splits out everything after an `=` separator
+  const lastEqualIndex = secretPath.lastIndexOf('=');
+  const splitSecretPath =
+    lastEqualIndex === -1
+      ? secretPath
+      : secretPath.substring(0, lastEqualIndex);
+  const value =
+    lastEqualIndex === -1
+      ? undefined
+      : secretPath.substring(lastEqualIndex + 1);
+  if (!secretPathRegex.test(splitSecretPath)) {
     throw new commander.InvalidArgumentError(
-      `${secretPath} is not of the format <vaultName>[:<directoryPath>]`,
+      `${secretPath} is not of the format <vaultName>[:<directoryPath>][=<value>]`,
     );
   }
-  // Returns [vaultName, secretName?]
-  const match = secretPath.match(secretPathNameRegex)!;
-  return [match[1], match[2] || undefined];
+  const [, vaultName, directoryPath] = splitSecretPath.match(secretPathRegex)!;
+  return [vaultName, directoryPath, value];
 }
 
 function parseSecretPath(secretPath: string): [string, string, string?] {
   // E.g. If 'vault1:a/b/c', ['vault1', 'a/b/c'] is returned
   //      If 'vault1', an error is thrown
-  const [vaultName, secretName] = parseSecretName(secretPath);
+  const [vaultName, secretName, value] = parseSecretPathOptional(secretPath);
   if (secretName === undefined) {
     throw new commander.InvalidArgumentError(
-      `${secretPath} is not of the format <vaultName>:<directoryPath>`,
+      `${secretPath} is not of the format <vaultName>:<directoryPath>[=<value>]`,
     );
   }
-  return [vaultName, secretName];
+  return [vaultName, secretName, value];
 }
 
 function parseSecretPathValue(secretPath: string): [string, string, string?] {
-  const [vaultName, directoryPath] = parseSecretPath(secretPath);
-  const lastEqualIndex = secretPath.lastIndexOf('=');
-  const value = lastEqualIndex === -1 ? '' : secretPath.substring(lastEqualIndex + 1);
+  const [vaultName, directoryPath, value] = parseSecretPath(secretPath);
   if (value != null && !secretPathValueRegex.test(value)) {
     throw new commander.InvalidArgumentError(
       `${value} is not a valid value name`,
@@ -219,13 +215,13 @@ function parseEnvArgs(
 }
 
 export {
+  secretPathRegex,
   secretPathValueRegex,
   environmentVariableRegex,
   validateParserToArgParser,
   validateParserToArgListParser,
   parseCoreCount,
-  parseVaultName,
-  parseSecretName,
+  parseSecretPathOptional,
   parseSecretPath,
   parseSecretPathValue,
   parseSecretPathEnv,
--- a/tests/secrets/list.test.ts
+++ b/tests/secrets/list.test.ts
@@ -42,19 +42,33 @@ describe('commandListSecrets', () => {
   });
 
   test(
+    'should fail when vault does not exist',
+    async () => {
+      command = ['secrets', 'ls', '-np', dataDir, 'DoesntExist'];
+      const result = await testUtils.pkStdio([...command], {
+        env: {
+          PK_PASSWORD: password,
+        },
+        cwd: dataDir,
+      });
+      expect(result.exitCode).toBe(64); // Sysexits.USAGE
+    },
+    globalThis.defaultTimeout * 2,
+  );
+
+  test(
     'should list secrets',
     async () => {
       const vaultName = 'Vault4' as VaultName;
       const vaultId = await polykeyAgent.vaultManager.createVault(vaultName);
 
       await polykeyAgent.vaultManager.withVaults([vaultId], async (vault) => {
-        await vaultOps.addSecret(vault, 'MySecret1', 'this is the secret 1');
-        await vaultOps.addSecret(vault, 'MySecret2', 'this is the secret 2');
-        await vaultOps.addSecret(vault, 'MySecret3', 'this is the secret 3');
+        await vaultOps.addSecret(vault, 'MySecret1', '');
+        await vaultOps.addSecret(vault, 'MySecret2', '');
+        await vaultOps.addSecret(vault, 'MySecret3', '');
       });
 
-      command = ['secrets', 'list', '-np', dataDir, vaultName];
-
+      command = ['secrets', 'ls', '-np', dataDir, vaultName];
       const result = await testUtils.pkStdio([...command], {
         env: {
           PK_PASSWORD: password,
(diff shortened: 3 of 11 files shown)


aryanjassal:
ls


This is the current state of the `ls` command. I'm not sure how to progress with the formatting of `ls` long output. The file permissions are redundant, and so are the users and groups. Instead of trying to retain perfect compatibility with UNIX commands, we will have to make a compromise.

For now, we need to remove permissions and users columns and fill it with something else. That is something we need to discuss. I have been thinking that we could do something like printing the vault ID instead of the permissions, and printing the type of file (directory or file) in place of the user and operators.

CMCDragonkai:
The metadata of `ls` is a bit redundant... given that it's always going to be the "root" user and "root" group. And permissions are always rwx for all.

So in a way we can cut down the number of irrelevant information while still preserving some interesting data. Like `d` for directories, file size and date time of last edit.

But other ideas is possible.

CMCDragonkai:
1. I think we stick with simple ls showing individual files, one after another.
2. Let's hold off on any options yet except maybe the long option..
3. I feel like hidden files may need to be shown by default. No need for option.

CMCDragonkai:
The key thing here is that LS doesn't really need to match the Unix LS command. It's not that well defined anyway.

CMCDragonkai:
And more advanced features to LS can be added in later.

I want to see the other commands implemented as they are more important.

aryanjassal:
I have noticed that `globWalk` works differently to how `ls` works. In `ls`, when any directories are listed, they are automatically entered and explored. However, in `globWalk`, that doesn't happen. I can understand that this decision makes sense, but this results in a lot of extra code to work around it. In this PR, I have to make a RPC call, then check if the result is a directory. If it is, then we need to make another RPC call with a different pattern specifically to explore the directory, which seems to me like a lot of unnecessary effort.

Moreover, by default, the 'glob' select (which is used internally to mean 'select everything') actually excludes hidden files, which is different to the behaviour in shell. This is what I have to current do to properly fetch all results, including hidden files, by default.
```ts
function formatPattern(pattern: string, directory: boolean = false): string {
function formatPattern(pattern: string, directory: boolean = false): string {
  if (pattern == '/') return '{*,.*}';
  if (pattern.charAt(0) == '/') pattern = pattern.substring(1);

  if (directory) pattern = `${pattern}/{*,.*}`;
  else pattern = `{*${pattern},.*${pattern}}`;
  return pattern;
}
```

The NPM page on `minimatch` mentioned that enabling the `dot` option would allow it to match filenames starting with `.`, but I haven't seen it work, so I will keep using this formatter for the time being.

I'm not sure if this is the best way to be handling this, so I would like @tegefaulkes to provide me feedback on this.

aryanjassal:
> And more advanced features to LS can be added in later.
> I want to see the other commands implemented as they are more important.

I am nearing completion on both the PRs in PK and PKCLI. The only major task that I have to do in https://github.com/MatrixAI/Polykey/pull/785 is adding the second RPC handler for streaming file contents over `RawHandler` and trim the functionality of `serializerStream` to only handle serialising file contents.

aryanjassal:
If two options coincide with their short name, then the option which was registered first would be set. For example, if I have two options: `'-l, --long'` and `'-l, --longer'`, then pass `-l` as a flag, I will see that only one of them will have the value of `-l` that was passed.

If the `long` option is added/registered first, then the value of `-l` will always update the `long` option. If the `longer` option is added/registered first, then `longer` will always be updated.

```console
[aryanj@matrix-34xx]$ pk test flags
Options:
  -l --long            Use a long listing format (default: false)
  -l --longer          Use a longer format (default: false)

[aryanj@matrix-34xx]$ pk test flags -l
long: true
longer: false

[aryanj@matrix-34xx]$ pk test flags -l -l
long: true
longer: false

[aryanj@matrix-34xx]$ pk test flags -l --longer
long: true
longer: true
```

If both the short and long name are the same, then the second registered option will always be `undefined`.
```console
[aryanj@matrix-34xx]$ pk test flags -l
long: true
another long: undefined
```

tegefaulkes:
Remember the discussion. We're not using the `globWalk` utility anymore. I think we can keep the functionality of that around for now but for now we don't need to implement it. That said, since globbing is a shell level construct, globbing normal fs paths will still work with the command so only the vault paths will not support it properly.

I think the we should at least stick to the format `ls -l` has so people can use our version interchangeably within their scripts. Same for the normal format. It should be simplified but I think each entry on a separate line would be more annoying to process than say, each entry separated by a space. That said, we could include the `--zero` and `-1` option formats.

Lastly, when checking the conflicting short options. Remember that we have multiple level commands such as `secrets` being one command and `secrets ls` being a sub command. You need to check how the short options conflict across the levels of these commands. and if ordering of the option matters here. `secrets -l ls` vs `secrets ls -l` vs `secrets -l ls -l`.

aryanjassal:
> Remember the discussion. We're not using the globWalk utility anymore. [snip]

I thought that we weren't going to use the 'glob' feature of the `globWalk` utility, as otherwise, how am I supposed to get the list of the files along their stats from the file system?

> I think the we should at least stick to the format ls -l has so people can use our version interchangeably within their scripts. [snip]

I also agree here, but it is true that most of that information is useless for us. We might be able to implement another flag which can use a newer formatting made for Polykey.

The `ls` utility also checks if it is running in interactive mode, and if it isn't then lists the files and directories one-by-one, otherwise lists them in the space-based format. I remember reading somewhere that we have checks for running in interactive mode, so we might be able to use them to implement this feature?

aryanjassal:
> Lastly, when checking the conflicting short options. [snip]

There are no options for the `secret` sublevel. As such, I added the `-l, --longer` option in the `CommandPolykey` module, directly on the `polykey` command level.

The results were as expected. As `-l, --longer` was registered before `ls`'s `-l, --long`, the root level `-l --longer` took precedence.

```console
[aryanj@matrix-34xx]$ pk secrets ls
Options:
  -np, --node-path <path>      Path to Node State (default: "/home/aryanj/.local/share/polykey", env: PK_NODE_PATH)
  -l --longer                  longer (default: false)
  -ni, --node-id <id>           (env: PK_NODE_ID)
  -ch, --client-host <host>    Client Host Address (env: PK_CLIENT_HOST)
  -cp, --client-port <port>    Client Port (env: PK_CLIENT_PORT)
  -l --long                    Use a long listing format (default: false)

[aryanj@matrix-34xx]$ pk secrets ls -l
long: false
longer: true
...

[aryanj@matrix-34xx]$ pk secrets -l ls
long: false
longer: true
...

[aryanj@matrix-34xx]$ pk -l secrets ls
long: false
longer: true
...
```

tegefaulkes:
I can't tell if you applied the two `-l` flags to `secrets` and `secrets ls` separately or both of them to `secrets ls`.

aryanjassal:
> I can't tell if you applied the two `-l` flags to `secrets` and `secrets ls` separately or both of them to `secrets ls`.

The `--longer` flag was applied to the `polykey` namespace, as there were no flags in the `secrets` namespace. The `--long` flag has been applied only on the `ls` namespace.

```console
[aryanj@matrix-34xx]$ pk --longer secrets ls
works

[aryanj@matrix-34xx]$ pk --long secrets ls
fails

[aryanj@matrix-34xx]$ pk secrets --long ls
fails

[aryanj@matrix-34xx]$ pk secrets --longer ls
works

[aryanj@matrix-34xx]$ pk secrets ls --longer
works

[aryanj@matrix-34xx]$ pk secrets ls --long
works

[aryanj@matrix-34xx]$ pk secrets ls --long --longer
works
```

CMCDragonkai:
> > Remember the discussion. We're not using the globWalk utility anymore. [snip]
> 
> I thought that we weren't going to use the 'glob' feature of the `globWalk` utility, as otherwise, how am I supposed to get the list of the files along their stats from the file system?
> 
> > I think the we should at least stick to the format ls -l has so people can use our version interchangeably within their scripts. [snip]
> 
> I also agree here, but it is true that most of that information is useless for us. We might be able to implement another flag which can use a newer formatting made for Polykey.
> 
> The `ls` utility also checks if it is running in interactive mode, and if it isn't then lists the files and directories one-by-one, otherwise lists them in the space-based format. I remember reading somewhere that we have checks for running in interactive mode, so we might be able to use them to implement this feature?

It's not recommended for people to create scripts depending on ls. Like I said ls is not a standardized command. So regular ls should just do the simple thing which in our case is listing the contents one line at a time.

As for -l option, yea it could work for us to show more data. It should still use one of the standard output formatters we use though.

CMCDragonkai:
> If two options coincide with their short name, then the option which was registered first would be set. For example, if I have two options: `'-l, --long'` and `'-l, --longer'`, then pass `-l` as a flag, I will see that only one of them will have the value of `-l` that was passed.
> 
> If the `long` option is added/registered first, then the value of `-l` will always update the `long` option. If the `longer` option is added/registered first, then `longer` will always be updated.
> 
> ```
> [aryanj@matrix-34xx]$ pk test flags
> Options:
>   -l --long            Use a long listing format (default: false)
>   -l --longer          Use a longer format (default: false)
> 
> [aryanj@matrix-34xx]$ pk test flags -l
> long: true
> longer: false
> 
> [aryanj@matrix-34xx]$ pk test flags -l -l
> long: true
> longer: false
> 
> [aryanj@matrix-34xx]$ pk test flags -l --longer
> long: true
> longer: true
> ```
> 
> If both the short and long name are the same, then the second registered option will always be `undefined`.
> 
> ```
> [aryanj@matrix-34xx]$ pk test flags -l
> long: true
> another long: undefined
> ```

So does this mean that if there is a global option like -l, this would conflict with our local option -l? At any case you then want to avoid any short or long options that is likely to be in conflict with any global option that currently has been defined on PK.

Again I would reduce the scope of this issue/pr to focus on the basic functionality first, not necessarily all the options of the Unix commands. It would take too long to complete otherwise.

aryanjassal:
> So does this mean that if there is a global option like -l, this would conflict with our local option -l? At any case you then want to avoid any short or long options that is likely to be in conflict with any global option that currently has been defined on PK.
> Again I would reduce the scope of this issue/pr to focus on the basic functionality first, not necessarily all the options of the Unix commands. It would take too long to complete otherwise.

Yes, if a global option and a local option both have conflicts, then the global one will take precedence.

I am working on the basic functionality, and that seems to be almost done. I can currently list the files in a vault, list files within a directory inside the vault, and list them in long format looking like this:
```console
[aryanj@matrix-34xx]$ pk secrets ls myvault:/ -l
FILE  65 Aug 19 10:07 .hidden-file.txt
FILE 175 Jun 13 16:53 another.txt
FILE  54 Jun 13 16:53 newsecret.txt
FILE  17 Jun 13 16:51 secret1.txt
 DIR   0 Aug 19 12:21 test

[aryanj@matrix-34xx]$ pk secrets ls myvault:test -l
FILE 45822 Aug 19 12:21 test/.hidden-document.docx
FILE 34478 Aug 19 11:20 test/image.jpg
FILE  3364 Aug 19 11:19 test/something.txt
```

tegefaulkes:
BTW, GlobWalk just walks the tree based on a glob pattern. It isn't an implementation for LS. So if it only matches the directory it will only output the directory and not its contents.

CMCDragonkai:
I prefer a suffix of `/` for directories and drop the entire FILE vs DIR column.

CMCDragonkai:
I wonder would the hardlink count be useful.

aryanjassal:
> I prefer a suffix of `/` for directories and drop the entire FILE vs DIR column.

Now, the output of the long list looks like this:
```console
[aryanj@matrix-34xx]$ pk secrets list myvault:/ -l
  65 Aug 19 10:07 .rippling-dns-record.txt
 175 Jun 13 16:53 another.txt
  54 Jun 13 16:53 newsecret.txt
  17 Jun 13 16:51 secret1.txt
   - Aug 19 12:21 test/

[aryanj@matrix-34xx]$ pk secrets list myvault:/ -lh
Size Date Modified Path
  65 Aug 19 10:07  .rippling-dns-record.txt
 175 Jun 13 16:53  another.txt
  54 Jun 13 16:53  newsecret.txt
  17 Jun 13 16:51  secret1.txt
   - Aug 19 12:21  test/
```

CMCDragonkai:
Let's add back in the hardlink counts.

> Yes, if a global option and a local option both have conflicts, then the global one will take precedence.

If this is true, then how is your `-lh` option working? Would't the short option of `-h` in the local case conflict with the `-h` of the global help option?

I'm basically pointing out here, that if the global option takes precedence, you cannot have any conflicting local options.

Global options stem from the `CommandPolykey`. Local options are from the individual subcommand.

CMCDragonkai:
BTW if you're doing prefix offsets, that means you are buffering the output in the case of the long option?

CMCDragonkai:
Here is the ACTUAL posix spec of `ls`: https://pubs.opengroup.org/onlinepubs/009695399/utilities/ls.html.

Have a read of the details there.

The `ls` that is in your terminal is GNU ls which isn't necessarily the stasndard https://unix.stackexchange.com/questions/754981/what-are-non-gnu-versions-of-terminal-commands.

CMCDragonkai:
```
d
Directory.
b
Block special file.
c
Character special file.
l (ell)
Symbolic link.
p
FIFO.
-
Regular file
```

Those are the entry type characters. In EFS, I don't believe we have block and character files. But we do have symlinks.

So...

```
d
l
-
```

Can be used.

CMCDragonkai:
> The default format shall be to list one entry per line to standard output; the exceptions are to terminals or when one of the -C, [[XSI](javascript:open_code('XSI'))] [Option Start] -m, or -x [Option End] options is specified. If the output is to a terminal, the format is implementation-defined.

CMCDragonkai:
> If no operands are specified, ls shall write the contents of the current directory. If more than one operand is specified, ls shall write non-directory operands first; it shall sort directory and non-directory operands separately according to the collating sequence in the current locale.

That does imply buffering up to sort by alphanum. We would just assume english. But technically unix paths are allowed any character except `/`. However I believe in the case of PK , would limit to a common set of characters available to all OS including windows limitations.

Would we then use LANG and other things for sorting? That doesn't even exist on Windows, so I guess we would just sort alphanum for now. Whatever JS does with https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort.

CMCDragonkai:
> If the -l option is specified without -L, the following information shall be written:
>
> "%s %u %s %s %u %s %s\n", <file mode>, <number of links>,
>    <owner name>, <group name>, <number of bytes in the file>,
>    <date and time>, <pathname>

So I would do something like this:

```
d 3 vair3u8kqfqoheulk277bp71j5vi80vd2bnavki02ekeounp72e80  11 Aug 15 15:22 tmp/
- 1 vair3u8kqfqoheulk277bp71j5vi80vd2bnavki02ekeounp72e80 946 May  5  2023 ADDITIONAL_TERMS
l 1 vair3u8kqfqoheulk277bp71j5vi80vd2bnavki02ekeounp72e80  71 May 10 21:13 result -> /nix/store
```

What do you think?

Maybe even `<nodeId>` is superfluous (because it's always going to be the same node ID), in which case:

```
d 3  11 Aug 15 15:22 tmp/
- 1 946 May  5  2023 ADDITIONAL_TERMS
l 1  71 May 10 21:13 result -> /nix/store
```

It's just the lack of separation between hardlinks and the file size is a bit difficult to decipher.

Without hardlinks...

```
d  11 Aug 15 15:22 tmp/
- 946 May  5  2023 ADDITIONAL_TERMS
l  71 May 10 21:13 result -> /nix/store
```

---

Hardlink information - primarily indicates whether the file data is going to be deleted. It is useful information as it will even tell you how many files in the directory.

CMCDragonkai:
Take note that the date time showing, changes dpeending on whether it is a regular time date vs year.

I think also it is localized against the local timezone too. So by default the files contain the UTC timezone, you'd want to localise that time date too.

CMCDragonkai:
Due to both alphabetical ordering AND prefix offsetting for formatting, that means the long format is the same as the table output formatter... However I believe they use space character, not tab character, so I guess it's not a table output formatter.

Either both styles involve buffering. It's not a stream based listing command, unless we implement `secrets find` (which we could...).

CMCDragonkai:
Oh and the `/` suffix for directories should only occur for the long mode. Not for regular `ls`, since we would want to make sure that when piping, any commands that make use of `ls` output should expect that all content entries are just the name of the file and nothing else.

CMCDragonkai:
In your tests, you should also test for possiblities of filenames as per https://stackoverflow.com/a/31976060. As well as dealing with potential newlines within a file name. And how the `ls` would show that too. We may need to do escaping.

For example if a file name is `a b c`, I believe, `ls` will show `'a b c'` with the single quotes.

Furthermore `\n` would get quoted too so `a\nb` would be `'a\nb'`.

So single quotation would occur for some whitespace characters.

However `a®b` stays `a®b`.

aryanjassal:
> If this is true, then how is your -lh option working? Would't the short option of -h in the local case conflict with the -h of the global help option?

Most likely because `-h` is defined by commander, and the reason I was able to override it is probably to allow for the user to override displaying the help menu. I will change it, then.

> BTW if you're doing prefix offsets, that means you are buffering the output in the case of the long option?

Yes, I am buffering the data by adding it to an array to apply further formatting to it. I haven't currently sorted the entries, just storing them in their iNode order and listing them in same.

> Oh and the / suffix for directories should only occur for the long mode. Not for regular ls, since we would want to make sure that when piping, any commands that make use of ls output should expect that all content entries are just the name of the file and nothing else.

Yes, this is how it is handled currently. However, there is one catch. If we want to list the contents of the root directory, we do:
```console
[aryanj@matrix-34xx]$ pk secrets ls newvault:
[wrong command]

[aryanj@matrix-34xx]$ pk secrets ls newvault:/
/.hidden-file.txt
/another.txt
/newsecret.txt
/secret1.txt
/test

[aryanj@matrix-34xx]$ pk secrets ls newvault:/test
/test/.document.docx
/test/image.jpg
/test/something.txt

[aryanj@matrix-34xx]$ pk secrets ls newvault:test
test/.document.docx
test/image.jpg
test/something.txt
```
Basically, leading slashes changes the output of the listing, and I'm not sure if this needs to change, or not.

> It's just the lack of separation between hardlinks and the file size is a bit difficult to decipher.

That is why I added an option to display the header in long formatting, to make it easier to understand. Maybe I can enable it by default?

> Take note that the date time showing, changes dpeending on whether it is a regular time date vs year.
> I think also it is localized against the local timezone too. So by default the files contain the UTC timezone, you'd want to localise that time date too.

By default, the date data is stored in milliseconds since 1 Jan 1970. I am currently using JavaScript's date utility, which automatically performs the conversion of the date based on the timezone of the location where it was called from.

> n your tests, you should also test for possiblities of filenames as per https://stackoverflow.com/a/31976060. As well as dealing with potential newlines within a file name. And how the ls would show that too. We may need to do escaping.

I haven't done this bit yet, but shouldn't be too difficult to implement.

aryanjassal:
> However I believe they use space character, not tab character, so I guess it's not a table output formatter.

At this point, I'm basically writing a table formatter, but replacing the tab character with a space. Maybe, instead of doing that, I can simply modify the `outputFormatterTable` to optionally take a custom separator, defaulting to `\t`.

CMCDragonkai:
> > BTW if you're doing prefix offsets, that means you are buffering the output in the case of the long option?
> 
> Yes, I am buffering the data by adding it to an array to apply further formatting to it. I haven't currently sorted the entries, just storing them in their iNode order and listing them in same.

Yes I'm just changing my tune here and saying that we can buffer it up and sort it.

> > Oh and the / suffix for directories should only occur for the long mode. Not for regular ls, since we would want to make sure that when piping, any commands that make use of ls output should expect that all content entries are just the name of the file and nothing else.
> 
> Yes, this is how it is handled currently. However, there is one catch. If we want to list the contents of the root directory, we do:
> 
> ```
> [aryanj@matrix-34xx]$ pk secrets ls newvault:
> [wrong command]
> 
> [aryanj@matrix-34xx]$ pk secrets ls newvault:/
> /.hidden-file.txt
> /another.txt
> /newsecret.txt
> /secret1.txt
> /test
> 
> [aryanj@matrix-34xx]$ pk secrets ls newvault:/test
> /test/.document.docx
> /test/image.jpg
> /test/something.txt
> 
> [aryanj@matrix-34xx]$ pk secrets ls newvault:test
> test/.document.docx
> test/image.jpg
> test/something.txt
> ``` 

No that's not what I meant here. I'm saying long mode in the sense of the `-l` option for the output of the command. And you should still enable the ability to do `ls newvault` and `ls newvault:/` simultaneously and `newvault:` should be an error.

CMCDragonkai:
> > However I believe they use space character, not tab character, so I guess it's not a table output formatter.
> 
> At this point, I'm basically writing a table formatter, but replacing the tab character with a space. Maybe, instead of doing that, I can simply modify the `outputFormatterTable` to optionally take a custom separator, defaulting to `\t`.

No that won't work, the table formatter should be using tab as tab spacing in the terminal fixes the tab width.

However I noticed that `ls` itself might actually be using spaces directly. So in the long mode, you'd format it similarly using the list format but with the appropriate spaces. See you have to find the longest field of each column and prefix accordingly if you want to match the exact same style.

You could have a different table output formatter thie time call it `outputFormatterTableWithSpaces` that does the calculation.

aryanjassal:
> No that won't work, the table formatter should be using tab as tab spacing in the terminal fixes the tab width.

In the latest commit, I actually implemented this already. So far, nothing has broken, and everything works as intended.

This is the output that the table formatter gives:
```console
[aryanj@matrix-34xx]$ pk secrets ls newvault:/ -lH
Type Size Links Date Modified Name
-    1    65    Aug 19 10:07  .rippling-dns-record.txt
-    1    175   Jun 13 16:53  another.txt
-    1    54    Jun 13 16:53  newsecret.txt
-    1    17    Jun 13 16:51  secret1.txt
d    2    0     Aug 19 12:21  test/

[aryanj@matrix-34xx]$ pk secrets ls newvault:/ -l
- 1 65  Aug 19 10:07 .rippling-dns-record.txt
- 1 175 Jun 13 16:53 another.txt
- 1 54  Jun 13 16:53 newsecret.txt
- 1 17  Jun 13 16:51 secret1.txt
d 2 0   Aug 19 12:21 test/
```

aryanjassal:
I have noticed two problems with how the table formatter formats the `ls` output. Firstly, `ls` outputs contents with padding applied to the beginning, but the table formatter does so with the padding applied to the end.

```
[aryanj@matrix-34xx]$ pk secrets ls newvault -l
- 1 65  Aug 19 10:07 .hidden-file.txt
- 1 175 Jun 13 16:53 another.txt
- 1 54  Jun 13 16:53 longer secret.txt
- 1 17  Jun 13 16:51 secret1.txt
d 2 0   Aug 19 12:21 test/

[aryanj@matrix-34xx]$ # What the correct version would output
[aryanj@matrix-34xx]$ pk secrets ls newvault -l
- 1  65 Aug 19 10:07 .hidden-file.txt
- 1 175 Jun 13 16:53 another.txt
- 1  54 Jun 13 16:53 longer secret.txt
- 1  17 Jun 13 16:51 secret1.txt
d 2   0 Aug 19 12:21 test/
```
I will add an option to the table formatter to change the text alignment from left aligned to right aligned.

The second issue I had was the auto-escaping done by the table formatter. It would escape things like quotations, but not space. For example, it would render `filename with spaces.txt` as-is, but if I try to manually insert the quotations, it would be escaped like this: `"\'filename with spaces.txt\'"`.

My first idea was to simply have an option to override the regex and the replacer, but that seemed too janky, so Brian helped me come up with another solution. Instead of overriding the escape function and regex, I can just disable it, and provide it with the padded strings by myself. The disabling of the escaping function will again be an option in the table formatter.

I need to use the table formatter, as the output of `ls` is basically covered by a table formatter, but with minor changes. Instead of rewriting an entire formatter in the command just to change a line or two of code, I can instead use and extend the table formatter instead, making my own code easier to write and providing extra functionality should it be needed in the future for any reasons.

aryanjassal:
> I will add an option to the table formatter to change the text alignment from left aligned to right aligned.

This actually does not work.

```console
[aryanj@matrix-34xx]$ pk secrets ls newvault -l
- 1  65 Aug 19 10:07  .hidden-file.txt
- 1 175 Jun 13 16:53       another.txt
- 1  54 Jun 13 16:53 longer secret.txt
- 1  17 Jun 13 16:51       secret1.txt
d 2   0 Aug 19 12:21             test/
```

It also aligns the files to the right, not just the text like hard links, block sizes, etc. basically any field which can have a variable length. Thus, the easiest solution right now appears to create a utility function which will apply padding for any field manually, then manually padding the hard links and file size field. Or, alternatively, just ignoring this issue and keeping everything left-aligned.

aryanjassal:
> 6. Support specifying multiple paths.
> 7. Seamlessly support using normal and secret paths.

For 6., it would be slightly more complex supporting that as we would need to perform multiple RPC calls and somehow handle displaying them all. This would especially be an issue as this style would be inherently incompatible with how the table formatter works, so this would force me to basically write a custom formatter in the `ls` command handler, which seems needlessly excessive.

For 7., I don't think supporting it is a good idea. We are only displaying information for the files within the vault, and only the information which is relevant to it. The filesystem would have more information than the information we show for vault files.

Moreover, I support syntax where you can just specify the vault name and all the files in the vault are automatically listed out. If we do end up implementing this functionality, then it would break this syntax.
```console
[aryanj@matrix-34xx]$ pk secrets ls newvault
```

My suggestion is to let the user use their own system utilities to interact with their local file system. For example, scripts made in windows relying on its `dir` utility will break if they replace the command with Polykey's `ls` command. Even the scripts written on linux using GNU `ls` will break as we don't follow the same layout, so I don't think we should display system files too.

Brian's argument was that we would need to seamlessly interact with the user's file system for commands like `cp`, but we already do that when we are using `pk secrets create`. So, overall, I think these last two points need to be thought about a bit more.

@tegefaulkes @CMCDragonkai

tegefaulkes:
The entire point of #32 is for it to function seamlessly like the unix commands. To do that both 6. and 7. are requirements for this.

Specifying multiple paths is pretty essential to the LS command. In fact, its why it works at all with the glob expansion the shell does. But the main thing is, without using the globWalk utility to specify a globstar pattern we have not way of outputting multiple files unless we specify multiple paths. Making multiple RPC calls shouldn't be a problem here. It should be expected at this stage.

7. is an essential feature of the other unix style commands. we should include it here if only to make all of the new commands consistent.

aryanjassal:
As per our discussion, I am removing the extended features included in this PR, including the detailed view, globbing, etc, as those features aren't high on the priority list now.

src/secrets/CommandList.ts
```diff
@@ -1,20 +1,26 @@
...
-    this.description('List all Available Secrets for a Vault');
-    this.argument('<vaultName>', 'Name of the vault to list secrets from');
+    this.name('ls');
+    this.aliases(['list']);
+    this.description('List all available secrets for a vault');
+    this.argument(
+      '<secretPattern>',
+      'Pattern to be matched, specified as <vaultName>[:<pattern>]',
+      binParsers.parseSecretName,
+    );
```
tegefaulkes: Description should be updated. Mention that it's a basic implementation of the LS command. Also the argument should be updated again. We're not matching a pattern for now.

src/secrets/CommandList.ts
```diff
@@ -40,23 +46,26 @@ class CommandList extends CommandPolykey {
...
-          return data;
+          for await (const secret of stream) data.push(secret);
+
+          // Remove leading slashes
+          return data.map((secret) => {
+            if (secret.path.startsWith('/')) {
+              secret.path = secret.path.substring(1);
+            }
+            return secret;
+          });
```
tegefaulkes: Why use a `.map` here? You can modify the data when pushing it into the array.

src/secrets/CommandList.ts
```diff
@@ -70,9 +79,7 @@ class CommandList extends CommandPolykey {
           process.stdout.write(
             binUtils.outputFormatter({
               type: 'list',
-              data: data.map(
-                (secretsListMessage) => secretsListMessage.secretName,
-              ),
+              data: data.map((secrets) => secrets.path),
```
tegefaulkes: You like your maps, why wasn't this logic done in a previous map over the data or when it was pushed into the array? You can consolidate it all together.

src/utils/parsers.ts
```diff
@@ -65,6 +66,19 @@ function parseCoreCount(v: string): number | undefined {
...
+  if (!secretPathNameRegex.test(secretPath)) {
+    throw new commander.InvalidArgumentError(
+      `${secretPath} is not of the format <vaultName>[:<directoryPath>]`,
+    );
+  }
+  // Returns [vaultName, secretName?]
+  const match = secretPath.match(secretPathNameRegex)!;
+  return [match[1], match[2] || undefined];
+}
+
```
tegefaulkes: Why do we have a whole new parser for this? Does the existing not cover it? What change needed to be made that needs this?

tests/secrets/list.test.ts
tegefaulkes, line 1: We need tests for the edge cases here as well. Such as

1. Directory doesn't exist.
2. Vault doesn't exist.
3. Tried to ls a secret.
4. Tried to ls a directory.
5. What about a nested directory `dir1/dir2`?

src/utils/parsers.ts
aryanjassal: Existing parsers require a pattern of `vaultName:secretPath`, but a requirement for this PR is that it should list the contents of the entire vault when `polykey secrets list vaultName` is run, which would otherwise raise an error during parsing.

src/secrets/CommandList.ts
aryanjassal: Oh, this is a remnant from when I had the long formatting of the files, as I had to retain all the data for long mode, but I didn't update this logic after its removal.

src/utils/parsers.ts
CMCDragonkai: I think one can unify the parsers, if the parser always returns `[vaultName, secretPath]`. And all other users of this can just throw the relevant exception...?

There must be examples of this compositionality elsewhere.

Or if the parsers actually just throw the exception directly, then it would be best to name each parser appropriately. Like `parseX` and `parseXY`.

aryanjassal: > There must be examples of this compositionality elsewhere.

I asked [ChatGPT](https://chatgpt.com/share/e/1ac30d0d-56f7-4f4e-a8d5-4a6585c38344) about this, and it did come up with a way to compose complex parsers based on simple parser "modules". (see https://github.com/MatrixAI/Polykey-CLI/issues/251#issuecomment-2311481854).

Basically, we can define parsers which do simple things like only validate the `vaultName`, or the `secretPath`, and we can compose a new parser on the fly with that. Moreover, an option for adding optional parsers also exist. For example, in the case of `ls` command, the parser can look like this:

```ts
// composeParser(mandatoryParsers, optionalParsers)
const parser = composeParser([vaultName], [secretPath]);

// Run parsing using the parser
// let result = Parse(parser);

// result = {
//   vaultName: 'vaultName'
// }
// --- or ---
// result = {
//   vaultName: 'vaultName',
//   secretPath: 'pathToSecret',
// }

tegefaulkes: There's no need to get that complex. We can just modify the existing parser to allow for just the vaultname. So parsing `'vaultName:'` would return `['vaultName', '',]`. `'vaultName:path'` gives us `['vaultName', 'path']` as usual. But now we allow just `'vaultName'` to return `['vaultname', undefined]`

Wherever we use the existing parser we need to add logic to check that the path is defined if it is required. Actually, we can create a 2nd parser to wrap the changed one so that it enforces the path requirement. We'd have to do this to ensure that the parsing error is handled properly by commander. So...

1. Modify the existing parser to allow parsing just the vault name without the `:` delimiter or the path being provided. Without the delimiter the path will be undefined.
2. creating a 2nd parser function that wraps the first just to enforce requiring the path.
3. Modify existing usage of the parser to use the newly created 2nd parser.

aryanjassal: I've done that. However, I'm also trying to look for a solution for #251.

Maybe I'll write one which returns only the vault name, which would also solve that issue. As I am already working on simplifying the parsers, I don't see a reason why to add that small change in a whole new PR.

aryanjassal:
Brian helped me with some minor issues and cleaning up the code for the parsers. This PR now looks ready. Merging.

PR #262, opened 2024-08-22
Refactor NixOS modules and implement `home-manager` modules

brynblack:
### Description

This PR updates the existing NixOS modules configuration to remove `programs.polykey` from the global user-space, and instead isolate it into a per-user based configuration utilising `home-manager`. 

### Tasks
- [x] 1. Remove `programs.polykey` NixOS module
- [x] 2. Add home-manager module for user-level service

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased


CMCDragonkai:
We still need the regular system module.

CMCDragonkai:
Also make sure to follow and reference the new flake output schema.

CMCDragonkai:
In your PR spec, you should be removing all the comments `<!-- --->`

CMCDragonkai:
The CI still needs to be updated with the system registries (that needs to be done on both https://github.com/MatrixAI/nixpkgs-matrix-private and Orchestrator) and then task 3. can only be done after this is merged, because you will get the commit revision and update the overlay.

brynblack:
I've merged it as it seems all good.

Issue #266, opened 2024-08-26
Implement `secrets edit` command


aryanjassal:
### Specification

If a secret doesn't exist when the `edit` command is invoked, then currently, the command just fails. Instead, the user needs to run `polykey secrets create` to copy a file from their file system and make that a secret, *then* they are able to use `polykey secrets edit` command. This is not great for UI/UX.

Moreover, each platform has their own preferences for file editor. As such, identify and launch the corresponding file editor to edit the secret. Windows would have `notepad.exe`, and MacOS and Linux editors would require to be inferred from environment. To ensure everything is working smoothly, integration tests need to be done on all the given platforms.

> Generally speaking you rely on an environment variable: `$VISUAL` then if not set, rely on `$EDITOR`, and if not set, then you default onto something that would exist on most platforms.
>
> [https://unix.stackexchange.com/questions/4859/visual-vs-editor-what-s-the-difference](https://unix.stackexchange.com/questions/4859/visual-vs-editor-what-s-the-difference)
>
> The "default" depends on some magic.
>
> Here's our list depending on the OS.
>
> 1. Linux - VISUAL, EDITOR, `nano`, `ed` - I wouldn't default to vi or vim because it's actually a more advanced editor.
> 2. Windows - VISUAL, EDITOR, get the `.txt` file association first, then default to `notepad.exe` and wait for it to finish. Windows never had a default terminal text editor.
> 3. MacOS - VISUAL, EDITOR, `pico`, `nano`
>
> If the defaults cannot be found, you need to error out, indicating that no default editor could be found, and users can instead choose to set VISUAL, EDITOR or pass in the "editor" program with `--ed` or something.
>
> Furthermore it is important that the temporary being created can only be edited by the same user as the user who ran `polykey secrets edit`, otherwise it can be intercepted. The file should have a limited umask like `700`. Or equivalent for the Windows.

The behaviour of `secrets edit` command shouldn't be what is simplest. It should be what makes the most sense.

For example, instead of making a blank secrets file if it doesn't exist, ignore the part where the secret contents are fetched instead. Point the editor, say, `vim` to the temporary file, but don't make the file yet. If the editor exists without making the file, the action was cancelled. Do nothing. No need to make a file in the vault. If, on the other hand, a file which wasn't there before was created, then the user wants to make a new file with that contents. Then, make a new file with the contents that the user just entered. Otherwise, it should behave as it already has been.

[...]

PR #267, opened 2024-08-26
`secrets edit` creates a secret if it doesn't exist

aryanjassal:
### Description

This PR tracks improving the `secrets edit` command to create a file in the vault if it didn't exist, otherwise edit the file contents as usual.

This iteration will focus on creating a new secret if it didn't exist at the time of editing it. Other features from the issue, including platform agnostic editor search or using `RawHandler` to stream file contents instead of `UnaryHandler` will be implemented in future iterations.

### Issues Fixed

* Relates to #266
* REF ENG-398

### Tasks

- [x] 1. Create file if it doesn't exist in the vault
- [x] 2. Write tests for `secrets edit`

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/src/secrets/CommandEdit.ts
+++ b/src/secrets/CommandEdit.ts
@@ -1,8 +1,8 @@
 import type PolykeyClient from 'polykey/dist/PolykeyClient';
 import fs from 'fs';
 import path from 'path';
-import * as errors from '../errors';
 import CommandPolykey from '../CommandPolykey';
+import * as errors from '../errors';
 import * as binUtils from '../utils';
 import * as binOptions from '../utils/options';
 import * as binParsers from '../utils/parsers';
@@ -24,6 +24,7 @@ class CommandEdit extends CommandPolykey {
     this.action(async (secretPath, options) => {
       const os = await import('os');
       const { execSync } = await import('child_process');
+      const vaultsErrors = await import('polykey/dist/vaults/errors');
       const { default: PolykeyClient } = await import(
         'polykey/dist/PolykeyClient'
       );
@@ -44,6 +45,10 @@ class CommandEdit extends CommandPolykey {
       this.exitHandlers.handlers.push(async () => {
         if (pkClient != null) await pkClient.stop();
       });
+
+      const tmpDir = await fs.promises.mkdtemp(
+        path.join(os.tmpdir(), 'polykey-'),
+      );
       try {
         pkClient = await PolykeyClient.createPolykeyClient({
           nodeId: clientOptions.nodeId,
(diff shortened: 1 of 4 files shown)


aryanjassal:
Based on the discussion we had in #255, for rapid iteration over these commands, do I merge this PR which introduces automatic file creation, or do I keep working on it and implement the platform-specific `$EDITOR` commands, which would take longer and is currently blocked by me not having access to a Windows or a Mac at this moment?

@CMCDragonkai @tegefaulkes

tegefaulkes:
If we're keeping to the iterative style then we should merge the current fix and move editor changes to a new issue.

There's a little more work to do here however. We need to use the new secret contents serialiser to stream over the secret contents.

aryanjassal:
> There's a little more work to do here however. We need to use the new secret contents serialiser to stream over the secret contents.

Oh, do you mean that we stop using `vaultsSecretsGet` and instead use the file contents serialiser proposed in https://github.com/MatrixAI/Polykey/pull/785?

tegefaulkes:
Yep, we nee to use the new serialiser to get the contents but also send the changed file back.

aryanjassal:
https://github.com/MatrixAI/Polykey-CLI/issues/266#issuecomment-2323542197

src/secrets/CommandEdit.ts
```diff
@@ -54,41 +54,66 @@ class CommandEdit extends CommandPolykey {
...
+                metadata: auth,
+                nameOrId: secretPath[0],
+                secretName: secretPath[1],
+              });
+            } catch (e) {
+              if (e.cause.name === 'ErrorSecretsSecretUndefined') {
+                secretExists = false;
+              } else {
+                throw e;
+              }
```
tegefaulkes: This is not how we check the type of secret. We want to use `instanceof` here.

```ts
if (e.cause instanceof ErrorSecretsSecretUndefined) ...
```

I also think we have a utility for unwrapping remote errors. It's called `remoteErrorCause` in the utilities. This will return the cause for that error.

On top of that you need to use `instanceof` to check its actually an instance of that error.

src/secrets/CommandEdit.ts
```diff
@@ -54,41 +54,66 @@ class CommandEdit extends CommandPolykey {
...
+                secretName: secretPath[1],
+                secretContent: content,
+              });
+              await pkClient.rpcClient.methods.vaultsSecretsEdit({
+                metadata: auth,
+                nameOrId: secretPath[0],
+                secretName: secretPath[1],
+                secretContent: content,
+              });
+            });
```
tegefaulkes: This is broken. It seems to create a new secret and then just edit it. But there is no logic branch for editing an existing secret. So editing a secret is just broken right now.

src/secrets/CommandEdit.ts
```diff
@@ -54,41 +54,66 @@ class CommandEdit extends CommandPolykey {
...
+            throw new errors.ErrorPolykeyCLIFileRead(e.message, {
+              data: {
+                errno: e.errno,
+                syscall: e.syscall,
+                code: e.code,
+                path: e.path,
+              },
+              cause: e,
+            });
+          }
```
tegefaulkes: This error handling doesn't make sense. You always need to check the error you're getting and never assume its going to be the error you think it is.  So always check the error you're getting then you can decide what to do with it. In this case we wrap the error. If its not the error we expect then we MUST re-throw that error. Never ignore an error unless its explicitly intended logic and always comment as to why. But its extremely rare that we just ignore errors.

This is important since improper handling of errors makes it extremely hard to debug a problem. You've seen it yourself today with that `read 0` error.

tegefaulkes (changes requested):
We need to add tests to show functionality of this command. Its actually a problem that there are no tests already but understandable because its kinda hard to test interactive input. You can create a test node or bash script and use that as a psudo editor for testing. The script will do two things, return to the test the contents of the file so we can verify it, but also edit the file with the contents we provide it.

These tests are essential to check for regressions when future changes are made so we really need tests here that cover all the intended functionality of the command. There are two main cases.

1. edit the file and check that the contents are updated. With the current code this would actually fail.
2. Edit a new file and show that it is created with the contents.

src/secrets/CommandEdit.ts
aryanjassal: I've fixed this in the newer commits. I will be writing test to ensure this doesn't happen again.

src/secrets/CommandEdit.ts
```diff
@@ -54,40 +55,85 @@ class CommandEdit extends CommandPolykey {
...
+              }
+            }
+            return exists;
+          },
+          meta,
+        );
+        // If the editor exited with a code other than zero, then execSync
+        // will throw an error. So, in the case of saving the file but the
+        // editor crashing, the program won't save the updated secret.
         execSync(`$EDITOR \"${tmpFile}\"`, { stdio: 'inherit' });
```
aryanjassal, line 89: Apparently when you invoke environment variables using `SVARIABLE` in `execSync`, it doesn't respect the variable value in `process.env`. For me, while I had set `process.env.EDITOR` to a script I made for testing the `secrets edit` command, the `execSync` kept opening my actual system editor.

However, logging the value of `process.env.EDITOR` showed me that the value was set correctly. So, when I replaced `$EDITOR` with `process.env.EDITOR`, the command started respecting the actual environment variable and using the script correctly.

It is weird, because `ExecSync` by default uses `NodeJS.ProcessEnv` for its environment variables, but for some reason, it didn't work with the tests.

aryanjassal:
https://github.com/MatrixAI/Polykey-CLI/blob/6c483ad6eea0a76ddb0a9b45e37cd4fca8ea4a8d/src/secrets/CommandEdit.ts#L89

Apparently when you invoke environment variables using `SVARIABLE` in `execSync`, it doesn't respect the variable value in `process.env`. For me, while I had set `process.env.EDITOR` to a script I made for testing the `secrets edit` command, the `execSync` kept opening my actual system editor.

However, logging the value of `process.env.EDITOR` showed me that the value was set correctly. So, when I replaced `$EDITOR` with `process.env.EDITOR`, the command started respecting the actual environment variable and using the script correctly.

It is weird, because `ExecSync` by default uses `NodeJS.ProcessEnv` for its environment variables, but for some reason, it didn't work with the tests.

src/secrets/CommandEdit.ts
```diff
@@ -1,8 +1,9 @@
 import type PolykeyClient from 'polykey/dist/PolykeyClient';
 import fs from 'fs';
 import path from 'path';
-import * as errors from '../errors';
+import * as vaultsErrors from 'polykey/dist/vaults/errors';
 import CommandPolykey from '../CommandPolykey';
```
tegefaulkes: Something like this needs to be imported asynchronously in the command. Have a look at `src/nodes/CommandFind.ts:26` for reference.

We do this to only import what we need when we need it. Otherwise we end up importing all of polykey regardless of the command we're running.

src/secrets/CommandEdit.ts
```diff
@@ -54,41 +55,85 @@ class CommandEdit extends CommandPolykey {
...
+                errno: e.errno,
+                syscall: e.syscall,
+                code: e.code,
+                path: e.path,
+              },
+              cause: e,
+            });
+          }
+          return;
         }
```
tegefaulkes: This is bad, we can't assume we'll never get any other error. The logic seems to be, if `ENOENT` and it existed then we want an error, if `ENOENT` and file didn't exist then we just allow it. but right now all other errors are ignored and that is really bad.

The 3rd case of other errors happening must be re-thrown. This is important.

src/secrets/CommandEdit.ts
```diff
@@ -54,41 +55,85 @@ class CommandEdit extends CommandPolykey {
...
-              secretName: secretPath[1],
-            }),
-          meta,
-        );
-        const secretContent = response.secretContent;
+
         const tmpDir = await fs.promises.mkdtemp(
           path.join(os.tmpdir(), 'polykey-'),
         );
         const tmpFile = path.join(tmpDir, 'pksecret');
```
tegefaulkes: I feel like the name of the secret we're editing rather than a hard coded `pksecret` name. This is because editors will tell you the name of the file it's editing and it should match up for convenience sake. Its not a big issue though.

src/secrets/CommandEdit.ts
```diff
@@ -54,41 +55,85 @@ class CommandEdit extends CommandPolykey {
...
+              secretContent: content,
+            });
+          } else {
+            await pkClient.rpcClient.methods.vaultsSecretsEdit({
+              metadata: auth,
+              nameOrId: secretPath[0],
+              secretName: secretPath[1],
+              secretContent: content,
+            });
+          }
```
tegefaulkes: This logic is slightly confusing, why are we even checking if the file exists here if we never reach this part of the code if it didn't. Clean it up.

src/secrets/CommandEdit.ts
```diff
@@ -54,41 +55,85 @@ class CommandEdit extends CommandPolykey {
...
+        }, meta);
         await this.fs.promises.rm(tmpDir, { recursive: true, force: true });
```
tegefaulkes: This can leak the secret if we run into any problems before this point. You need to remove the temp file in the finally block just below to ensure that it's always removed. You also need to handle any `ENOENT` errors when trying this in case it directory you want to remove doesn't exist anymore. Any other error should still be thrown.

tegefaulkes (changes requested):
Some small changes are we're good.

tegefaulkes (approved):

Issue #278, opened 2024-09-10
Adding long list `-l` format and hidden file support to `secrets ls` command


aryanjassal:
### Specification

In the previous issue, https://github.com/MatrixAI/Polykey-CLI/issues/245 ([ENG-358](https://linear.app/matrix-ai/issue/ENG-358/implement-secrets-ls-command)), listing the contents of a single directory was implemented instead of listing the contents of the entire vault. This issue now includes the remaining tasks for: long format listing, showing hidden files, using wildcards and specifying multiple paths.

The `secrets ls` command will list the contents of a directory.

There are 3 main features to support here.
1. List in list format `-l`
2. Include hidden files with `-a`
3. Paths can support wildcards and globing.
4. Multiple paths can be specified.
5. `stdin` is ignored

```
Polykey/tmp/asd on  staging [?] 
❯ ls
asdasd  best  dir1  dir2  one  test  two

Polykey/tmp/asd on  staging [?] 
❯ ls -l
total 32
-rw-r--r-- 1 brian operators  8 Jul 11 11:18 asdasd
-rw-r--r-- 1 brian operators 15 Jul 11 11:18 best
drwxr-xr-x 2 brian operators  2 Jul 11 11:19 dir1
drwxr-xr-x 2 brian operators  2 Jul 11 11:19 dir2
-rw-r--r-- 1 brian operators  0 Jul 11 11:18 one
-rw-r--r-- 1 brian operators  5 Jul 11 11:18 test
-rw-r--r-- 1 brian operators  0 Jul 11 11:18 two

Polykey/tmp/asd on  staging [?] 
❯ ls -a
.  ..  asdasd  best  dir1  dir2  .hidden  one  test  two

Polykey/tmp/asd on  staging [?] 
❯ ls *est
best  test

Polykey/tmp/asd on  staging [?] 
❯ ls **/*     
asdasd  best  dir1/a  dir2/b  dir2/c  one  test  two

dir1:
a

dir2:
b  c

Polykey/tmp/asd on  staging [?] 
❯ ls -l **/*
-rw-r--r-- 1 brian operators  8 Jul 11 11:18 asdasd
-rw-r--r-- 1 brian operators 15 Jul 11 11:18 best
-rw-r--r-- 1 brian operators  0 Jul 11 11:42 dir1/a
-rw-r--r-- 1 brian operators  0 Jul 11 11:43 dir2/b
[...]

PR #279, opened 2024-09-10
Adding `secrets rm` command

### LINKED ISSUE #247: Implement `secrets rm` command
### Specification

The `secrets rm` command will remove one or more files specified by paths.

1. Will remove the file at the specified path.
2. Multiple paths can be specified.
3. Supports wildcard and globing.
4. Supports a `-r` recursive option for deleting directories with contents.
5. Supports normal and secret paths.

### Additional context

Related #32

### Tasks

1. Implement `secret rm` command.
2. Supports multiple specified paths.
3. supports wildcard and globing
4. supports a `-r` option to recursively delete directories and their contents.

aryanjassal:
### Description

Implement `secrets rm` command following the behaviour of UNIX. For example, passing the `-r` flag would recursively delete a directory and it's contents, and we can supply multiple secret paths to the command.

### Issues Fixed

* Related to #247 ([ENG-360](https://linear.app/matrix-ai/issue/ENG-360/implement-secrets-rm-command))
* Related to https://github.com/MatrixAI/Polykey/pull/803
* Closes ENG-360

### Tasks

- [x] 1. Add new `secrets rm` command (functionally similar to `secrets del`)
- [x] 2. Add support for specifying and deleting multiple files
- [x] 3. Supply `--recursive` to delete directories and their contents
- [x] 4. Add tests

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased


src/utils/options.ts
```diff
@@ -302,6 +302,11 @@ const order = new commander.Option(
   .choices(['asc', 'desc'])
   .default('asc');
 
+const recursive = new commander.Option(
+  '-r --recursive',
+  'If enabled, specified directories will be removed along with their contents',
+).default(false);
+
```
tegefaulkes: I think we've decided that short form options like `-r` shouldn't be used for the sub commands. Its too easy to conflict with other commands.

tests/secrets/remove.test.ts
```diff
@@ -0,0 +1,154 @@
+import type { VaultName } from 'polykey/dist/vaults/types';
+import path from 'path';
+import fs from 'fs';
+import Logger, { LogLevel, StreamHandler } from '@matrixai/logger';
+import PolykeyAgent from 'polykey/dist/PolykeyAgent';
+import { vaultOps } from 'polykey/dist/vaults';
+import * as keysUtils from 'polykey/dist/keys/utils';
+import * as testUtils from '../utils';
+
+describe('commandDeleteSecret', () => {
```
tegefaulkes: Update the describe name.

tests/secrets/remove.test.ts
tegefaulkes, line 1: The current tests look fine. But we're missing tests for when specifying multiple vaults. E.G. `secrets rm vault1:secret1 vault1:secret2 vault2:secret1 vault2:secret2`

What happens in this case? is it supported? I can think of...
1. Its supported as works across multiple vaults. - would be the best
2. Not supported and throws an error stating as such. - Acceptable.
3. Is supported for multiple vaults, but doesn't group all of the vault paths for a single vault into a single `vaultsSecretsRemove` call. Not great, it will break the condition where doing multiple removes in a go would only have 1 commit in the vault log.
4. Not supported but does some undefined behaviour like deletes all the paths provided but only in one of the vaults. - really bad.

tegefaulkes: Also add a test that checks if all the files are removed within a single vault commit. Looking over the code I don't think that condition is upheld currently.

src/secrets/CommandRemove.ts
```diff
@@ -45,20 +47,19 @@ class CommandDelete extends CommandPolykey {
...
               nameOrId: secretPath[0],
-              secretName: secretPath[1],
-            }),
-          auth,
-        );
+              secretNames: [secretPath[1]],
+              options: { recursive: options.recursive },
+            });
+          }
+        }, meta);
```
tegefaulkes: This isn't good. Removing all of the secrets needs to be done in a single `vaultsSecretRemove` RPC call. It needs to take the array of vault secret paths and do all the logic in the handler.

Alternatively for now we can just handle a single vault at a time. This probably makes more sense for now so all you need to do is check that all of the secret paths are for the same vault, if not then throw an error.

tegefaulkes (changes requested):
I've found some small problems.

src/utils/options.ts
aryanjassal: I don't think that should be done for this command. Deleting a directory is impossible without the `--recursive` flag det, and deleting directories is a common enough operation that it would be a major annoyance for the users. 

Moreover, it would conflict with programmers' muscle memory, defeating the purpose of UNIX command parity that we are trying to implement.

src/secrets/CommandRemove.ts
aryanjassal: I think I can sort the paths by the vault name and then split the groups based on that. Then, instead of sending one RPC call per file, we can do that per vault. Also, if we see that the vault name is undefined, we can try to look for the path on the system fs.

src/utils/options.ts
tegefaulkes: You misunderstand me. I just mean remove the `-r` so it's only `--recursive`. I'd like to have the short version but we've found that it's too likely to conflict with the higher level command options.

src/secrets/CommandRemove.ts
tegefaulkes: In any case, minimise the processing on the client side. Have the RPC handler handle this logic.

You can definitely sort and group the array by vault. It should be minimal effort to have it support multiple vaults then. 

Let's not bother with handling the local FS as well just yet.

src/utils/options.ts
aryanjassal: Oh sorry, I also wasn't sufficiently clear. I meant that deleting directories is a common enough operation that typing `--recursive` each time we want to do that seems like a pain.

On that note, I don't think we even can implement the -f option, as that option forcefully removes unwritable files, but the vaults don't have any concept of actual file perms, so it will be redundant.

CMCDragonkai:
Yes in that case -f is useless don't put that in. Add in -r though.

CMCDragonkai:
It's about consistent semantics not just convenience.

src/utils/options.ts
CMCDragonkai: See: https://github.com/MatrixAI/Polykey-CLI/issues/278#issuecomment-2342559234 for the problem.

src/secrets/CommandRemove.ts
aryanjassal: > In any case, minimise the processing on the client side. Have the RPC handler handle this logic.

Currently I am passing a single vault name as the `nameOrId` parameter of the RPC call. Instead of doing this, I can pass in the separated secret paths, so the handler can deal with the raw parsed input by itself. I don't like this because it differs from the parameter structure that every other command uses and would feel weird to use.

```ts
// Current implementation

const vaultName = 'vault';
const secretNames = ['secret1', 'secret2'];
await pkClient.rpcClient.methods.vaultsSecretsRemove({
  metadata: auth,
  nameOrId: vaultName,
  secretNames: secretNames,
  options: { recursive: options.recursive },
});

// This method

const secretNames = [
  ['vault1', 'secret1'], 
  ['vault2', 'secret2'],
];
await pkClient.rpcClient.methods.vaultsSecretsRemove({
  metadata: auth,
  secretNames: secretNames,
  options: { recursive: options.recursive },
});
```

Instead, I think I should do the calculation on client-side, grouping the secrets by the vault name. This would make the usage consistent for all RPC handlers.

```ts
// Proposed implementation
const secretNames = [
  ['vault1', 'secret1', 'secret2', 'secret3'], 
  ['vault2', 'secret4', 'secret5'],
];
for (const group of secretNames) {
  await pkClient.rpcClient.methods.vaultsSecretsRemove({
    metadata: auth,
    nameOrId: group.shift(), // Pop vault name
    secretNames: group, // Rest of the secrets
    options: { recursive: options.recursive },
  });
}
```
I like the second approach much more, but it comes at a cost of more processing for the client. I'm not sure which route to proceed with.

tegefaulkes: I lean to keeping the logic more clustered to where it's needed. to me it makes the most sense to have the utility function `secretsRemove` deal with all the processing and grouping.

aryanjassal: Then I will proceed with the first method, passing a single `secretNames` instead of passing `nameOrId`. It will be expected that `secretNames` will have both the vault name and the secrets to delete from each vault.

Or maybe there is a better way to proceed with it?

It just makes sense to me that all RPC handlers modifying vault contents will modify only a single vault, keeping the command structure and usage consistent, then the client will extend that functionality if needed.

aryanjassal:
https://github.com/MatrixAI/Polykey-CLI/blob/40fb687c9209408965043c2b322f3a2643e275bc/tests/secrets/stat.test.ts#L51-L64

I've noticed something interesting. In almost every test, we define `command` as a list of commands. But, for some reason, we unwrap the list, then wrap the contents into an array again, like `[...command]`. It can all be simplified to simply passing in the array directly, like this:

```ts
const command = ['secrets', 'rm', '-np', dataDir, `${vaultName}:secret`];

const result = await testUtils.pkStdio(command, {
  env: { PK_PASSWORD: password },
  cwd: dataDir,
});
```
This would have the result of making the tests slightly faster and more readable. I have also noticed that in some tests, `command` is defined globally, and in others, it is defined as a `const` for each test (see file linked above for this behaviour). I think we should define it in each test to ensure the command remains scoped to that test.

tegefaulkes (approved):

aryanjassal:
All the tasks have been finished for this PR and approval has also been granted. Merging.

CMCDragonkai:
> https://github.com/MatrixAI/Polykey-CLI/blob/40fb687c9209408965043c2b322f3a2643e275bc/tests/secrets/stat.test.ts#L51-L64
> 
> I've noticed something interesting. In almost every test, we define `command` as a list of commands. But, for some reason, we unwrap the list, then wrap the contents into an array again, like `[...command]`. It can all be simplified to simply passing in the array directly, like this:
> 
> ```ts
> const command = ['secrets', 'rm', '-np', dataDir, `${vaultName}:secret`];
> 
> const result = await testUtils.pkStdio(command, {
>   env: { PK_PASSWORD: password },
>   cwd: dataDir,
> });
> ```
> 
> This would have the result of making the tests slightly faster and more readable. I have also noticed that in some tests, `command` is defined globally, and in others, it is defined as a `const` for each test (see file linked above for this behaviour). I think we should define it in each test to ensure the command remains scoped to that test.

Spreading is usually intended for shallow copies. Not sure what that's for.

aryanjassal:
rm

CMCDragonkai:
Interesting it's 1 commit message.

Do note to use ls not list in order to be consistent in our messaging.

Issue #243, opened 2024-07-10
Implement `secrets cat` command


tegefaulkes:
### Specification

As per the #32 issue refactor, we need to implement a simplified `cat` command as `secrets cat`.

This will function identically to `cat` while omitting some of the options for now. It pretty much combines the functionality of `read` and `write` commands along with the concatination `cat` is known for.

Uses are

1. read the contents of a secret `secrets cat secretPath` or file `secrets cat filePath`.
2. write the `stdin` to a secret `cat > secretPath` or file `cat > filePath`. `stdin` will support pipes as well.
3. concatenating files together in the output. `secrets cat secretPath filePath secretPath/*`.

Pretty much the the functionality boils down to two things

1. Will output the contents of all specified files to `stdout` in the order they are specified.
2. If no files are specified then it will just output `stdin` to `stdout`.

~~Redirection into a secret path might not be possible with how commander parses inputs and how shell works, at least, not with the `cat > secretPath` syntax. We'll need to prototype that. If it doesn't work, we'd have to use an option `--redirect secretPath` to enable the functionality.~~ (see [comment](https://github.com/MatrixAI/Polykey-CLI/issues/243#issuecomment-2351954854))

### Tasks

1. Create a `secrets cat` command
2. If no arguments are provided it will take `stdin` and output to `stdout`.
3. If arguments are provided it will write their contents to `stdout` in the order they are provided.
4. Support specifying normal files along side secret paths.
5. Support specifying wildcards and possibly globbing for the paths.
6. ~~Support redirecton into a secret using `secretPath` syntax. Failing that provide an option that handles that behaviour.~~ (see [comment](https://github.com/MatrixAI/Polykey-CLI/issues/243#issuecomment-2351954854))

tegefaulkes:
@aryanj you want to work on this one next after completing your current work.

tegefaulkes:
For streamlining the basic functionality. just focus on creating the cat command and having it handle multiple paths within a single vault. Don't worry about using the seralizer utility for now. I'm starting to think I need to review how it works and it's use based on what we've learned.

Focus on tasks 1 and 3. Use a `ServerHandler` to stream the contents as `ContentMessage`  much like the commands are already doing. <br><br>In the future we'll have to update all the commands with the serialiser optimisation. And the other tasks for this issue will be a future PR.

aryanjassal:
[...]

PR #282, opened 2024-09-16
Adding `secrets cat` command and removing `secrets get` command

aryanjassal:
### Description

This PR tracks the status of implementing `secrets cat` command, which can be used to concatenate multiple secrets together.

### Issues Fixed

* Relates to https://github.com/MatrixAI/Polykey-CLI/issues/243
* Closes ENG-356

### Tasks

- [x] 1. Add `secrets cat` command
- [x] 2. Remove `secrets get` command
- [x] 3. Add support for concatenating multiple files from multiple vaults together
- [x] 4. Redirect `stdin` to stdout if no arguments are provided
- [x] 5. Add tests for new functionality

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased


src/secrets/CommandRemove.ts
```diff
@@ -49,13 +50,30 @@ class CommandDelete extends CommandPolykey {
...
+              first = false;
+            }
+            writer.close();
+          })();
+          return response;
         }, meta);
+
+        if (!(await response.output).success) {
+          throw new errors.ErrorPolykeyCLIFileRead('Failed to delete file');
+        }
```
aryanjassal: I need feedback on this. How do I handle a non-zero success code?

In Polykey, the RPC will either return a true success code, or throw an error. It will never return a non-zero success code. To ignore the return code (which, in this case, is what we can do) and simply await the output, we can do this:

```ts
await response.output
```

Or do I need to properly handle it? I guess handling it properly would be future-proof, but I would need to make proper error throwing and handling. I'm not sure if an error already exists which can be thrown at a non-zero return code.

src/secrets/CommandCat.ts
```diff
@@ -8,17 +8,19 @@ import * as binProcessors from '../utils/processors';
 class CommandGet extends CommandPolykey {
   constructor(...args: ConstructorParameters<typeof CommandPolykey>) {
     super(...args);
-    this.name('get');
-    this.description('Retrieve a Secret from the Given Vault');
+    this.name('cat');
+    this.description('Retrieve one or more Secrets from a Vault');
```
tegefaulkes: Should be more descriptive of what the cat command does. For example the cat man page says `cat - concatenate files and print on the standard output`

src/secrets/CommandCat.ts
```diff
@@ -49,16 +51,55 @@ class CommandGet extends CommandPolykey {
...
+
+            process.stdin.once('error', (code) => cleanup(code));
+            process.stdin.once('close', () => cleanup());
+            process.stdin.once('end', () => {
+              process.stdout.write(stdout);
+              resolve(0);
+            });
+          });
+          return;
+        }
```
tegefaulkes: This needs to be rewritten.
1. You're accumulating the stdin and writing it to out as a whole chunk. There's no reason to accumulate it in memory like that.
2. The clean up code is broken. IT doesn't remove any of the actual event listeners you register. Its using `removeAllListeners` which we never want to use because it will break anyone else adding their own listeners.
3. The clean up always needs to run once the stdin and stdout have ended. Seems to only be called when there's an error.

src/secrets/CommandCat.ts
```diff
@@ -49,16 +51,55 @@ class CommandGet extends CommandPolykey {
...
+            writer.close();
+          })();
+          return response;
+        }, meta);
+
+        let rawSecretContent: string = '';
+        for await (const chunk of response.readable) {
+          rawSecretContent += chunk.secretContent;
+        }
+        const secretContent = Buffer.from(rawSecretContent, 'binary');
```
tegefaulkes: This applies more when we get the proper content streaming going, but you want to avoid writing all the contents to a string and then writing a file in a single go. You should get a file fd and write it in smaller parts as you get them.

src/secrets/CommandRemove.ts
tegefaulkes: In a few places where we do something that doesn't really return anything useful we just include the success message. But logically we'd only get `success: true` or an error. The fact that it resolves without error indicates success here. But the fact that we include a success message at all means we might as well check it.

tests/secrets/cat.test.ts
```diff
@@ -0,0 +1,148 @@
...
+    const vaultIds: Array<VaultId> = [];
+    for (const vaultName of vaultNames) {
+      vaultIds.push(await polykeyAgent.vaultManager.createVault(vaultName));
+    }
+    const secretNames = ['secret-name1', 'secret-name2', 'secret-name3'];
+    const secretVaultNames = [
+      [0, secretNames[0]],
+      [1, secretNames[1]],
+      [0, secretNames[2]],
+    ];
```
tegefaulkes: This is a very confusing way to structure all your test data. It's better to keep it all flattened and simple. You don't need an array and loop over it to create vaults. You can just keep each vault name in its own variable, create each vault manually and save their Ids in their own variables. `vault1`, `vaultId1` etc etc/

tegefaulkes (changes requested):
Some changes needed. I made some comments but also address what we discussed.

src/secrets/CommandRemove.ts
aryanjassal: If, after checking, the result is not true, then what do I do? What error should I throw? Currently, I'm just throwing a generic error as I don't know what else can fit the bill.

Do I create a new error for situations where the return code was not true, and throw that? Do I throw another, existing error?

src/secrets/CommandCat.ts
aryanjassal: So, I should open up a temporary file, then stream the file contents into that file to accumulate them, then after the incoming stream has concluded, I can just read the file contents and stream that to the output, saving memory for large files and concatenations.

Is that what you meant?

tegefaulkes: If you were writing to a file then you'd open the file to get the fd and write to it in chunks rather than get all the contents in memory and do a single `writeFile`.

In this case we're writing to stdout so we just need to do `process.stdout.write()` each chunk as we get them. We never want to save the whole contents of a file in memory like this.

Actually, looking closer at the code, the format option doesn't make sense for this command, we shouldn't be using it or the output formatter. We should be writing the contents directly to stdout. Or just use the raw type on the output formatter. I don't think having a JSON format for CAT makes sense here.

tests/secrets/cat.test.ts
tegefaulkes, line 1: How does the command handle a missing vault or secret? What error do we get? If we cat 3 secrets but the 3rd one errors what does the output look like?

tegefaulkes:
Weird, I replied to your question as part of a new review so it posted it twice.

tests/secrets/cat.test.ts
aryanjassal: Regular `cat` will output the file contents of the files which exist, and print out a generic error string for files which don't.

```console
[aryanj@matrix-34xx:~]$ cat file1 file2 wrongfile file3
this is the contents of file 1
this is the contents of file 2
cat: wrongfile: No such file or directory
this is the contents of file 3
```

Our command will output file contents correctly until we hit a file which doesn't exist. At that point, it will raise an error and exit.

```console
[aryanj@matrix-34xx:~]$ polykey secrets cat vault:file1 vault:wrongfile vault:file3
this is the contents of file 1
ErrorPolykeyRemote: Remote error from RPC call
  ...
  command   vaultsSecretsGet
  ...
  cause: ErrorSecretsSecretUndefined: Secret does not exist - Secret with name: wrongfile does not exist
```

I feel we shouldn't ignore the error and continue to spit out other files which do exist. It feels counter-intuitive, when an error is being reported to the user, but we are continuing with code flow as usual.

What do you think, @tegefaulkes?

src/secrets/CommandCat.ts
aryanjassal: Following a later comment, I have ditched accumulating the file contents and using output formatters. I print out the contents to `stdout` directly.

On that note, I was using `process.stdout.end()` to indicate that we will not be sending any more data to the output stream, and closing it. 

However, after research, I have discovered that this practice is discouraged with standard out streams. They are managed by NodeJS, and will end when the process terminates. 

If we manually end the streams, then any attempts of writing to `stdout`, be it intentional or accidental, will raise an error which will propagate to the top, crashing the program.

As such, I have removed all usages of `process.stdout.end()`.

src/secrets/CommandRemove.ts
aryanjassal: Right now, the control flow reaching the `await response.output` statement indicates a success. While we *can* test for success to be true, I don't think it makes sense to do that, especially if we need to gracefully handle it by throwing an error, etc.

That can be implemented when needed. For now, a simple `await` is sufficient.

tests/secrets/cat.test.ts
tegefaulkes: I think we should match the behaviour of the cat command.

aryanjassal: To match the functionality, instead of passing an array of file paths, I need to iterate over each path and concatenate it separately.

I need to output the error to standard error instead of standard out, as `cat` does this. It ensures, when piping, the error isn't piped, but the user can still see the error. 

Anyways, I'm not sure how I can handle this from the RPC where it can throw a pseudo-error, where the error contents are valid and go to `stderr`, but it doesn't cause us to exit the RPC's control flow.

As such, I went the naive route by running `vaultsSecretsGet()` on each secret path, and ignoring the error if the secret is undefined.

aryanjassal: After discussing this with Brian, we settled on adding an option in the metadata which decides if we fail on file not found, or just return an error message to the client, and continue concatenating other files.

The option of making a new RPC handler was explored, but making an entire new RPC handler to change like 10 lines of code seemed redundant and caused code duplication.

The approach of making a new RPC call for each path is inefficient and will result in significant bottlenecks when concatenating multiple files.

aryanjassal:
`cat` in UNIX, when being piped into another command, buffers the contents internally, accumulating `stdout`. Any errors written to `stderr` are printed immediately, and `stdout` is printed after all the files have been fully read. This results in the following output:
```console
[aryanj@matrix-34xx:~]$ cat file1 nofile file2
content1
cat: nofile: No such file or directory
content2

[aryanj@matrix-34xx:~]$ cat file1 nofile file2 | cat
cat: nofile: No such file or directory
content1
content2
```
This implementation directly streams contents to `stdout` as we receive them. This results in the output being as follows:
```console
[aryanj@matrix-34xx:~]$ polykey secrets cat vault:file1 vault:nofile vault:file2
content1
ErrorSecretsSecretUndefined: nofile: No such file or directory
content2

[aryanj@matrix-34xx:~]$ polykey secrets cat vault:file1 vault:nofile vault:file2 | cat
content1
ErrorSecretsSecretUndefined: nofile: No such file or directory
content2
```
It looks like the command is not doing anything, but if we redirect `stderr` to null in both cases, we get this:
```console
[aryanj@matrix-34xx:~]$ cat file1 nofile file2 2> /dev/null
content1
content2

[aryanj@matrix-34xx:~]$ polykey secrets cat vault:file1 vault:nofile vault:file2 2> /dev/null
content1
content2
```
We currently don't detect if piping is being done or not, and we don't have a way of replicating this behaviour.

CMCDragonkai:
> `cat` in UNIX, when being piped into another command, buffers the contents internally, accumulating `stdout`. Any errors written to `stderr` are printed immediately, and `stdout` is printed after all the files have been fully read. This results in the following output:
> 
> ```
> [aryanj@matrix-34xx:~]$ cat file1 nofile file2
> content1
> cat: nofile: No such file or directory
> content2
> 
> [aryanj@matrix-34xx:~]$ cat file1 nofile file2 | cat
> cat: nofile: No such file or directory
> content1
> content2
> ```
> 
> This implementation directly streams contents to `stdout` as we receive them. This results in the output being as follows:
> 
> ```
> [aryanj@matrix-34xx:~]$ polykey secrets cat vault:file1 vault:nofile vault:file2
> content1
> ErrorSecretsSecretUndefined: nofile: No such file or directory
> content2
> 
> [aryanj@matrix-34xx:~]$ polykey secrets cat vault:file1 vault:nofile vault:file2 | cat
> content1
> ErrorSecretsSecretUndefined: nofile: No such file or directory
> content2
> ```
> 
> It looks like the command is not doing anything, but if we redirect `stderr` to null in both cases, we get this:
> 
> ```
> [aryanj@matrix-34xx:~]$ cat file1 nofile file2 2> /dev/null
> content1
> content2
> 
> [aryanj@matrix-34xx:~]$ polykey secrets cat vault:file1 vault:nofile vault:file2 2> /dev/null
> content1
> content2
> ```
> 
> We currently don't detect if piping is being done or not, and we don't have a way of replicating this behaviour.

1. The pipe buffering is still streamed, it's just dependent on chunk size.
2. You can detect if you are composed but that's a CLI concern not the PK agent concern.
3. Leave the pipe buffering to the CLI - it's done by default by the OS. You don't need to do it, you just switch modes.
4. Between agent to client you should still be streaming regardless of what the CLI is doing later.

aryanjassal:
> Between agent to client you should still be streaming regardless of what the CLI is doing later.

Yes, that is what I am currently doing. Brian mentioned writing the streamed contents we receive into a temporary file, and then stream the file's contents to `stdout`, but streaming the incoming data straight to `stdout` was much simpler and faster.

tegefaulkes:
Try not to mix up what I said. 

https://github.com/MatrixAI/Polykey-CLI/pull/282#discussion_r1772361639

tegefaulkes (approved):

aryanjassal:
All the checks have been passed, all the tasks have been finished, and the PR has also been approved. Merging.

tests/secrets/cat.test.ts
CMCDragonkai: What's this metadata? How you do it internally may be up to the RPC protocol. But on the outside running `cat` may need to report the error differently.

Firstly is it reporting on STDERR or on STDOUT? That informs what we may do.

On our end, we may report an exception report on STDERR, different from regular STDOUT, yet continue concatenating onwards.

We don't have do report the same error.

Also beware of the posix cat spec, if it mentions what is actually required versus optional/implementation dependent.

You got to show the gif demo, and have it talked about it the different scenarios.

src/secrets/CommandCat.ts
CMCDragonkai: Wouldn't a JSON format simply be applied on the outside? And in that case, we would look at using JSONL designed for streaming, not a JSON array.

aryanjassal:
cat

CMCDragonkai:
> cat [ cat ](https://private-user-images.githubusercontent.com/84617406/370558446-37bcb443-9c34-4189-bf13-26e929983569.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MjcyNDI2MDgsIm5iZiI6MTcyNzI0MjMwOCwicGF0aCI6Ii84NDYxNzQwNi8zNzA1NTg0NDYtMzdiY2I0NDMtOWMzNC00MTg5LWJmMTMtMjZlOTI5OTgzNTY5LmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDA5MjUlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwOTI1VDA1MzE0OFomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTQ4MGZjZjgzMGE4ZjBlMDM1MTMwMjJiNmFhZWUxZWU0ZDMyNzU0ZmFjYjY1MjcwMjRjZjY0NDljYThlOTc4OTMmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0In0.YkUZij2VM6pyB4ktaYjS0RvXTxL0rUvmWITdLPwntKw) [ ](https://private-user-images.githubusercontent.com/84617406/370558446-37bcb443-9c34-4189-bf13-26e929983569.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MjcyNDI2MDgsIm5iZiI6MTcyNzI0MjMwOCwicGF0aCI6Ii84NDYxNzQwNi8zNzA1NTg0NDYtMzdiY2I0NDMtOWMzNC00MTg5LWJmMTMtMjZlOTI5OTgzNTY5LmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDA5MjUlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwOTI1VDA1MzE0OFomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTQ4MGZjZjgzMGE4ZjBlMDM1MTMwMjJiNmFhZWUxZWU0ZDMyNzU0ZmFjYjY1MjcwMjRjZjY0NDljYThlOTc4OTMmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0In0.YkUZij2VM6pyB4ktaYjS0RvXTxL0rUvmWITdLPwntKw)

Is it possible for you to disable your autosuggest before you record.

aryanjassal:
Yes. I can actually edit the recording itself to remove the captured autosuggest lines. I will do this and upload the new recording when I next work.

aryanjassal:
cat

I have edited the GIF to not include autocompletion anymore. I have also sped it up a little bit. @CMCDragonkai 

I will be attaching the raw `.cast` file for reference and archival purposes. The file will be attached as a text file as GitHub doesn't support attaching files with arbitrary extensions.

[cat.cast](https://github.com/user-attachments/files/17158702/cat.txt)

PR #289, opened 2024-09-27
Polykey secrets env integration

brynblack:
### Description

This PR integrates the `polykey secrets env` command into the development shell hook, to securely load development secrets into the development environment.

It replaces `.env` with `pk secrets env` to securely access secrets from the running Polykey agent under AWS.

### Tasks

- [x] 1. Replace `. ./.env` with `pk secrets env ...`
- [x] 2. Update CI to include Polykey
- [ ] 3. Pull secrets from agent (using organisation root password)

### Final checklist

* [ ] Domain specific tests
* [ ] Full tests
* [ ] Lint fixed
* [ ] Squash and rebased

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -106,7 +106,7 @@
             shellHook = ''
               echo "Entering $(npm pkg get name)"
               set -o allexport
-              . ./.env
+              . <(pk secrets env Polykey-CLI:.)
               set +o allexport
               set -v
               ${lib.optionalString ci ''


tegefaulkes:
@brynblack does this not link back to any issues?

flake.nix
```diff
@@ -101,7 +101,7 @@
             shellHook = ''
               echo "Entering $(npm pkg get name)"
               set -o allexport
-              . ./.env
+              . <(pk secrets env Polykey-CLI:.)
```
CMCDragonkai, line 109: This should just be `Polykey-CLI`, no need for `:.`.

Also again, this assumes we have a vault called `Polykey-CLI`, let's also make this command fail-safe, so that even if the vault doesn't exist, it can continue.

The usual way to do this is with `|| true` when the command returns a non-zero exit code.

Also note that `set -o allexport` ensures that all variables declared afterwards is exported.

If the `pk secrets env` command produces exported variables, that isn't necessary anymore. This is why I wanted to ensure that sourcing doesn't do that by default.

The final thing is the lack of a vault schema making this very implicit, we are essentially injecting arbitrary global variables - whereas what we want is a schema defining what we expect to be in the vault. We have an issue for this too.

Also `pk` is an alias, the official program command should be `polykey`.

Another thing is that this assumes `polykey` exists in the development shell o r available in the CLI. I usually ensure that the program exists by adding it to `nativeBuildInputs`.

Would this be a problem for user-space Polykey?

Remember we have:

1. System deployed Polykey - system configuration service
2. User deployed Polykey - home manager service
3. Project deployed Polykey?

A command running by itself without an agent - I wonder how that should work?

CMCDragonkai: @tegefaulkes @aryanjassal I want to ensure that our secrets env is foolproof when we have this running...

And by bootstrapping PK from PK... need to examine all the potential problems.

CMCDragonkai:
Where's the issue?

flake.nix
tegefaulkes: I'll create a new issue for discussing the spec of the `secrets env` command changes. There are things that need to be hashed out.

aryanjassal: > This should just be `Polykey-CLI`, no need for `:.`.

This is being tracked in https://github.com/MatrixAI/Polykey-CLI/pull/312

aryanjassal: This is now done, so this thread can now be resolved.

aryanjassal:
We need to get this issue completed. The main blocker for this issue is the usage of polykey in the CI environment.

So far, we have discussed that creating an agent the normal way isn't predictable, as it randomly generates a node id. As such, we cannot delegate authority to a randomly-created agent.

We can use the recovery passphrase to re-generate the node with the same node ID. However, how will the password and the passphrase be stored safely? If it is hard-coded in the repo or in the CI file, then it can be leaked. Now that we are no longer using a runner image in newer CIs, we cannot pre-load the information onto the runner file system either. Perhaps the passphrase and password could be a repo-specific secret.

This is not the end of all issues, however. We also need to figure out a distribution method for a Polykey agent.

Should we integrate the agent in the runner image itself? That can no longer be done, as we are transitioning to the usage of a custom action which sets up nix on the stock runners. No images are being used anymore, so this idea will not work after the CI upgrade.

We currently don't have an action which might set up polykey and run the agent in the background. Brynley has created a custom action which prepares the nix environment, so she might be able to set up a Polykey action. The simplest way I can think of doing this would be downloading the package directly from the releases page for the appropriate platform, then using the aforementioned methods to launch an agent and delegate it authority.

Alternatively, we can use the inbuilt package manager to download Polykey. (https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/customizing-github-hosted-runners#installing-software-on-ubuntu-runners). This method works on all platforms, so custom packages can be imported. However, the examples have used `apt` for Ubuntu, `brew` for MacOS, and `chocolatey` for Windows. We currently don't have Polykey published to these package managers. There might be ways to 'sideload' package managers and get Polykey, but that likely won't be supported and an API change might introduce failures in all our CIs.

This is partially related to https://github.com/MatrixAI/Polykey/issues/222. Implementing egress schema would help with smooth CI by restricting the exported secrets for each workflow.

This leads to another point - how to delegate secrets. Should each repo get a vault? Or each branch? Technically, each runtime is unique, so do we need a unique vault per runtime? (of course not!)

Once support for egress schema has been implemented, a vault can be dedicated for a single repo, with egress schema controlling the secrets that are exported. This idea seems the best to me, but perhaps I could be overlooking something.

Eventually, we would use a tool like Orchestrator or even Polykey Enterprise to manage secret delegation, but initially, we would need someone to delegate and manage these secrets, which can get cumbersome very quickly.

However, there is another issue with dogfooding Polykey to this extent. 

What if a commit deploys a change which breaks Polykey on the CI, but isn't caught by the tests? Then, the newer Polykey version would be published, but break in CI, which would prevent any other commits to trigger the CI, becoming deadlocked in a broken state. We would need to manually downgrade polykey to allow the CI to trigger, then upgrade it to the default version.

That issue is specific to a scenario where Polykey is always fetched from the latest version. However, what about the case where polykey would be fetched from a fixed version instead? In that case, if a new major update has released, polykey would need manual update in all repos to bring the pin to the latest version.

I haven't thought up of a solution to this issue yet. This potential failure case might need more discussion.

This is what I could understand about the current state of this issue; about getting Polykey into the CI environments. I might be missing some key details, though. Some discussion might be needed. Thoughts, @CMCDragonkai @brynblack?

CMCDragonkai:
As we had discussed during our earlier meeting, I believe it make sense to have a Polykey agent running separately orchestrated by the Orchestrator. This agent will run in our cloud. Then our CI workers will instead pull secrets down according to the schema.

You want to incorporate:

* https://github.com/matrixai/polykey/issues/222 to support the egress `secrets.schema.json`
* https://github.com/matrixai/polykey/issues/835 to a lesser extent this is important for any usage of "desktop CLI tools" - but ideally we prefer to use env variables where we can
* https://github.com/MatrixAI/Orchestrator/issues/42
* https://github.com/MatrixAI/Orchestrator/issues/36
* https://github.com/MatrixAI/Orchestrator/issues/37
* https://github.com/MatrixAI/Orchestrator/issues/33
* https://github.com/MatrixAI/Orchestrator/issues/1

There's the issue of secret-zero. How do you "give the initial secret" to the polykey client to call the agent?

There was recently some discussion about "workload identities" I did with ChatGPT, and I believe there's a token delegation - short-lived delegation process that should be investigated: [ChatGPT-IAM and Service Accounts.pdf](https://github.com/user-attachments/files/18647045/ChatGPT-IAM.and.Service.Accounts.pdf)

But I think atm, the easiest way is to actually just pass the root password to each job at a Organisation level.

But @brynblack you have not been keeping up closing issues, so there's way too much entropy here. I'm going to be trying to get @Abby010 to help (along with some debug/tracing problems for PK in production). It's time to think operationally now.

CMCDragonkai:
@brynblack comment when consolidated.

brynblack:
I updated the spec a little, still not sure if the PR is the best place to put all this discussion and/or perform reconciliation. Isn't the best solution for now just to put the root password for the agent into GitHub secrets manager and then use that for connecting to the agent?

Issue #242, opened 2024-07-10
Implement `secrets write` command


tegefaulkes:
### Specification

As part of the #32 issue refactoring, we need to implement `secrets write` command. This command is the most basic method of writing secrets.

#### ~~`secrets read`~~

~~This command needs to read the contents of a secret in a vault. The contents will be outputted to `stdout`. Pretty much just the existing `secrets get` command. I don't think any changes need to be made to it.~~

~~The command takes the form of `pk secrets read [secretPath]`. The output can be redirected with `pk secrets read [secretPath] > some file` and piping into it shouldn't do anything.~~ (see [https://github.com/MatrixAI/Polykey-CLI/issues/242#issuecomment-2378351024](https://github.com/MatrixAI/Polykey-CLI/issues/242#issuecomment-2378351024))

#### `secrets write`

This command needs to update or create the contents of a file. There will be two ways to provide the secret contents here. Coompared to current implementation, this pretty much is just `secrets create` but allows you to update a secret as well.

1. Interactively via `stdin`. so if you call `secrets write secretPath` you can type in you contents to the terminal and send `EOF` (ctrl+d) to finish writing.
2. Piping data in.

In both cases the behaviour should match `cat > file`.

### Additional context

* Related #32

### Tasks

1. ~~Implement `secrets read` command.~~
2. Implement `secret write` command.
3. ~~create tests demonstrating behaviour for read.~~
4. create tests demonstrating behaviour for write.

CMCDragonkai:
Why not just go straight to Unix commands?

tegefaulkes:
I'm following the spec you wrote out where `Imagine (IN ORDER OF PRIORITY):` and `read`, `write` were top of the list. I can skip theses then since `cat` will basically cover that functionality anyway.

CMCDragonkai:
The `read` and `write` commands are not normal unix commands. The historical context for this is https://github.com/MatrixAI/Polykey-CLI/issues/32#issuecomment-1767788230 (which should be added as the specific additional context above).

I would like `cat` to be an alias of the `read` command though as it's quite a common thing for shell users to use.

tegefaulkes:
While technically using some shell shenanigans you can use `cat` to both read and write files.  That functionality will not work too well for PK and the vaults. We will still use `cat` to cover the `read` case here. But we still need to create a `write` command that will write the `stdin` into the secret we want.

aryanjassal:
[...]

PR #292, opened 2024-09-27
Implementing `secrets write` command

aryanjassal:
### Description

The secrets write command will take data written to `stdin` and write that into a secret file of our choosing.

To simplify this, work has been put into <b>[Polykey#814](https://github.com/MatrixAI/Polykey/pull/814)</b> to automatically write to a new file if it exists, and create the file if it doesn't.

As the PR in Polykey modifies the behaviour of `VaultsSecretsEdit` (removes it, actually, and replaces it with `VaultsSecretsWriteFile`), the `secrets edit` command also needs to be updated with this new behaviour.

### Issues Fixed

* Relates to <b>[Polykey-CLI#242](https://github.com/MatrixAI/Polykey-CLI/issues/242)</b>
* REF ENG-355
* Supporting work in <b>[Polykey#814](https://github.com/MatrixAI/Polykey/pull/814)</b>

### Tasks

- [x] 1. Create `secrets write` command
- [x] 2. Update `secrets edit` with new RPC handler
- [x] 3. Write tests using fast-check
- [x] 4. Remove `secrets update`

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased


tests/secrets/writeFile.test.ts
```diff
@@ -0,0 +1,136 @@
...
+  });
+  afterEach(async () => {
+    await polykeyAgent.stop();
+    await fs.promises.rm(dataDir, {
+      force: true,
+      recursive: true,
+    });
+  });
+
+  test.prop([stdinArb, contentArb], { numRuns: 5 })(
```
aryanjassal: Using multiple runs on a test which uses RPC calls is very time-taking. Running five tests takes almost 10 seconds.

I'm not sure what to do with this. Is this a reasonable trade-off, or should I reduce the runs even more? Or this small number of runs cannot hope to catch any real bugs, so I need to run more tests?

CMCDragonkai: That shouldn't be the case. Can you flamegraph this?

aryanjassal: I'm not sure how to use flamegraph, so I have asked Brian's help with this.

In the meantime, I ran some more tests, and got an average of 400ms after trying to `cat` a secret. However, I got an average of 2789ms when I tried to run `secrets write` with the same string as what the file contained for the `secrets cat` tests (it being `"testing"`).

`secrets cat` relies on a Duplex Handler to stream file paths to the handler and stream file contents from the handler. However, `secrets write` relies on a Unary Handler to do so, as we want to upgrade to a Raw Handler, but it needs too much work for it to be included in the first iteration.

As such, I believe the performance impact is due to the different handlers. This is also reinforced by the fact that running a similar test on `secrets list` gets a similar average time of 519ms while it is using a Server Handler.

tegefaulkes: I doubt it's the kind of handler that's the problem. Keep in mind that the main difference between list, cat and write is the fact that you're writing to a file this time. You wan't to sanity check that the writing itself isn't just this slow normally. Fortunately we can compare to the vaults test in `Polykey` to see how long a write takes there. Take that as the base line.

tegefaulkes: We did some digging and narrowed it down to the middlewear. Possibly the password hashing step. Needs some more digging so we create a new issue to track it. https://github.com/MatrixAI/Polykey-CLI/issues/294

src/secrets/CommandWrite.ts
```diff
@@ -1,29 +1,24 @@
...
+import * as binProcessors from '../utils/processors';
+import * as binParsers from '../utils/parsers';
 import * as binUtils from '../utils';
 import * as binOptions from '../utils/options';
-import * as binParsers from '../utils/parsers';
-import * as binProcessors from '../utils/processors';
 
-class CommandUpdate extends CommandPolykey {
+class CommandStat extends CommandPolykey {
   constructor(...args: ConstructorParameters<typeof CommandPolykey>) {
```
tegefaulkes: `CommandStat` inside a `CommandWrite.ts` file? The class name should reflect the file name.

src/secrets/CommandRemove.ts
tegefaulkes, line 1: This one is called `CommandDelete` inside a `CommandRemove.ts`

tegefaulkes (commented):
Some minor changes.

src/secrets/CommandWrite.ts
aryanjassal: Oops, I forgot to do this when I copied over the command boilerplate from another command. I will update this.

tegefaulkes:
Looks OK to merge.

aryanjassal:
All tasks have been done and approval has been granted. Merging.

aryanjassal:
I got Terminalizer working on NixOS. It was a little involved.

First, it is a npm package which doesn't exist under the nix repository. So, I had to install it globally. As I cannot use the regular `npm install -g terminalizer` to install packages on NixOS, I made a directory in my home called `.npm-globals`. In this directory, I then installed the package, then added `~/.npm-globals/node-modules/.bin` to my `$PATH`, which allowed `terminalizer` to be found as an executable.

This allowed me to record terminal, but rendering was failing as it needed to import Electron, and it was a dynamically linked library. Nix was complaining about it, so I launched a `nix-shell` with `steam-run` installed with the following command:

```console
[aryanj@matrix-34xx:~]$ NIXPKGS_ALLOW_UNFREE=1 nix-shell -p steam-run
```

This launched me into an environment which simulates regular linux libraries. Then, running `terminalizer render <recording>` allowed me to finally render the recording and get the fancy MacOS-like frame.

Do note that rendering Terminalizer recordings is much slower than asciinema recordings. AGG (asciinema gif generator) can generate a gif within a second, while for this recording, Terminalizer rendering took a bit more than 4 minutes.

On the other hand, editing recorded files is very easy, as each input/output is stored in a YAML file as a pair of delay and content fields. In asciinema, if I wanted to remove a section from the recording, I had to modify the time stamps across the entire file, otherwise there would have been an awkward delay present in the recording. Terminalizer removes that completely, as it uses delay, which calculates the delay from previous content chunk. This makes it trivial to edit recordings. The output file being YAML also means that full syntax highlighting/LSP support also exists.

Overall, except the workarounds needed for Nix and the long rendering times, I would prefer using Terminalizer more than asciinema.

Here is the rendered gif. The recording file will also be attached for reference and archival purposes. We can edit the background color and padding. I have added ample padding, so we can crop it to our needs later than need to re-export with a larger padding.

render1727747739986

[write.yml](https://github.com/user-attachments/files/17198213/write.txt)

Issue #246, opened 2024-07-11
Implement `secrets mkdir` command


tegefaulkes:
### Specification

`secrets mkdir` is pretty simple, it will create a directory at the specified paths. 
1. Creates a directory at the specified path.
2. Can take multiple paths to create multiple directories.
3. Does not support wildcards or globing
4. If a parent directory of a path doesn't exist then it will throw `ENOENT`
5. If `-p` is specified then it will create missing parent directories.

### Additional context

Related #32 

### Tasks
1. Implement `secrets mkdir` command
2. Supports multiple paths for creation.
3. Supports `-p` flag for recursively creating directories.
4. Can take either normal paths or secret paths.

aryanjassal:
UNIX `mkdir` doesn't stop on error. It keeps going, making what directories it can make, leaving what it cannot. I think that instead of returning an `ENOENT` or other error, it should catch it, and ignore it, returning an error string that can be printed out instead.

Currently, we just return a success message. Instead of that, we can use duplex streaming to return an error message or success flag every time a directory is created, or we can accumulate the errors and return them in one large chunk at the end of the whole process.

I would also need to make a new `vaultOps` for returning an error string instead of throwing an error, so if one path is invalid in multiple vaults, it won't stop creating more paths, but return an error string instead, and continue. Either this, or I can implement a flag to continue on errors in the existing `vaultOps.mkdir`.

I need some inputs on this.

tegefaulkes:
Go with the duplex method then. Keep in mind that you may need to support making directories across multiple vaults at once. And multiple directories in a single vault must be done in a single commit in the vault. That will affect how you implement things.

aryanjassal:
> Go with the duplex method then. Keep in mind that you may need to support making directories across multiple vaults at once. And multiple directories in a single vault must be done in a single commit in the vault. That will affect how you implement things.

Yes, I am taking inspiration from how I handled this in `VaultsSecretsRemove` to delete files from all vaults in a single commit.

PR #300, opened 2024-10-04
Allow `mkdir` command to create multiple directories

aryanjassal:
### Description

The current `mkdir` implementation only supports creating a single directory at a time. This needs to be expanded to work similar to UNIX's `mkdir` command, which can make multiple directories at the same time, and does not stop when it encounters an error.

This PR implements similar features to `secrets mkdir` command, allowing the creation of multiple directories across vaults in a single command.

Note: as per <b>[Polykey#819 (comment)](https://github.com/MatrixAI/Polykey/pull/819#discussion_r1787095322)</b>, a warning or notice needs to be sent to `stderr`, warning users that the directory, while it is created, is not part of the source tree and a file must exist within the directory in order for the directory to be detected as a part of the source tree.

### Issues Fixed

* Relates to <b>[Polykey-CLI#246](https://github.com/MatrixAI/Polykey-CLI/issues/246)</b>
* REF ENG-359
* Git doesn't track empty directories: <b>[Polykey#819 (thread)](https://github.com/MatrixAI/Polykey/pull/819#discussion_r1782474124)</b>

### Tasks

- [x] 1. Update `secrets mkdir` to align with UNIX `mkdir`
- [x] 2. Update tests for `secrets mkdir`
- [x] 3. Add warning informing users about empty directories being out-of-tree

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased


aryanjassal:
How do I handle sending the warning to the user? Should a random warning message be simply sent to `stderr` like this? Is the amount of information sufficient, or does it need rewording?

```console
[aryanj@matrix-34xx:~]$ pk secrets mkdir vault:dir
WARNING: Empty directories will not be affected by modifications to the vault state (version change, sharing/cloning, etc.).
``` 

It seems a bit random how `secrets rm`, `secrets cat`, and now `secrets mkdir` all have their own versions of error/warning messages with no standard.

Like I discussed in <b>[Polykey#819](https://github.com/MatrixAI/Polykey/pull/819)</b>, it would probably be better if instead of passing an error message, we instead return an object containing details of the error (serialising the error will probably take up too much bandwidth for minimal to no return), which can then be interpreted by the client (CLI) to adapt to their own needs. For example, when we make a Polykey Desktop, then the error messages we have are designed for UNIX CLI and would feel awkward on other clients.

Instead of relying on UNIX's behaviour and messages, which itself doesn't follow a strict standard, we should devise our own standard of printing warnings and errors. This is different to the formatters, as the formatters will format a certain type of output in a certain way. We, however, need a standard of message content.

We also need a standard way to display these warnings/errors. It could be made using a new formatter, it could be made using the logger. It doesn't matter how we get consistency, but I'd say consistency in user experience is something pretty important.

tegefaulkes:
Yeah we'll want a standard way of dealing with this but it doesn't need to be a fully serialized version of the Error object itself. Create a standard error message under the client types and have all rm, cat and mkdir conform to this structure when sending the details of the error. This is only for expected errors from the FS. If it was an unexpected error the we still throw it using the normal methods.

You'll need a way to tell the errors from the normal messages. for this you'd use a typed object for pattern matching the messages.

So we'd have
```ts
type Message = {
 //...whatever
}

type ErrorMessage = {
 type: 'error',
//...whatever
}

type HandlerMessage = Message | ErrorMessage;

```

So typescript is smart enough to tell the two apart. You can pattern match for one or the other implcity by checking if `message.type === 'error'`. If that is true then we know it's a error message otherwise its a normal message.

```ts
if (message.type === 'error') {
   // message will by typed as a ErrorMessage here implicitly.
   } else {
      // message is now a Message here.
   }
```

Use this to tell them apart on the client side and construct our formatted message. As for what to make the error message. Just include the minimal details of the error we need to know. Message, error name, code and exit code if applicable.

aryanjassal:
> So typescript is smart enough to tell the two apart. You can pattern match for one or the other implcity by checking if `message.type === 'error'`. If that is true then we know it's a error message otherwise its a normal message.

I implemented this here, <b>[Polykey#819](https://github.com/MatrixAI/Polykey/pull/819)</b>, but I ran into an interesting issue.

Typescript doesn't actually quite work like this. If only one object has the `type` attribute, then typescript will complain of a missing attribute. As such, both types need to feature a common attribute for it to work like this.

As `SuccessMessage` is a single boolean, and it is used everywhere, I chose to add the `success` type to `ErrorMessage` instead, to push up a proof-of-concept for this. The PoC is working as intended, so we now need to decide what to do with the common attribute set. Do we have `success` as the common attribute, or `type`?

In my opinion, having `type` would make sense, but a lot of code would need to change for that. Or, we can have a `SuccessTypeMessage` and `ErrorTypeMessage` to differentiate the `{ success: boolean; }` from `{ type: 'success'; success: boolean; }`.

Need some thoughts on this @tegefaulkes.

tegefaulkes:
In the example I gave I figured you could drop the type parameter for one of the messages and still pattern match on that. In that case you'd have to check if the type parameter actually exists which is a little more annoying. So if both messages have the `type` defined then you can just pattern match on that and get the implicit type reduction.

Let's stick with `type` then. But would a lot a code change due to this? As far as I can tell it would only affect the response messages of the secret command handlers that need it. And obviously their CLI code using it but that's not a huge change.

aryanjassal:
> Let's stick with type then. But would a lot a code change due to this? As far as I can tell it would only affect the response messages of the secret command handlers that need it. And obviously their CLI code using it but that's not a huge change.

Yes, a couple of handlers use this and a couple places in the CLI too. I still think that there could be two success messages. One is the `SuccessMessage` that we are already using, and other is the `SuccessTypeMessage` which could have the type attached to it, however it probably would result in code duplication and redundancy.

As such, I will make the relevant changes across Polykey's handlers and CLI's relevant commands and tests.

src/secrets/CommandMkdir.ts
```diff
@@ -50,16 +54,41 @@ class CommandMkdir extends CommandPolykey {
...
+              const error = result as ErrorMessage;
+              let message: string = '';
+              if (error.code === 'ENOENT') {
+                message = 'No such secret or directory';
+              } else if (error.code === 'EEXIST') {
+                message = 'Secret or directory exists';
+              } else {
+                throw new ErrorPolykeyCLIUncaughtException(
+                  `Unexpected error code ${error.code}`,
+                );
```
aryanjassal: I need to check if the incoming error is not what we expected, even if the two possible codes are `ENOENT` and `EEXIST`. However, I'm not sure what is the best way to check that and confirm the error codes are valid.

Currently, I am throwing an `ErrorPolykeyCLIUncaughtException`, but that doesn't seem as if it should be thrown intentionally. Should I not worry about invalid error codes? Or is my current approach to check for valid error code is fine?

src/secrets/CommandMkdir.ts
```diff
@@ -50,16 +54,48 @@ class CommandMkdir extends CommandPolykey {
...
+          }
+          await writer.close();
+          return response;
+        }, meta);
+
+        // As git does not track empty directories in version control, we need
+        // to inform the user about this.
+        process.stderr.write(
+          'WARNING: Empty directories will not be affected by modifications to the vault state (version change, sharing/cloning, etc.)',
         );
```
aryanjassal: I'm not sure how to word the warning for this. This also needs a standard. However, this is the one (and possibly only) situation where this might come up, so we might not need to worry about this.

Also, how do we deal with formatting long strings? The formatter enforces max character length to approximately 80 characters, but it is not easily possible to do that with strings. However, for strings of this length, there are ways of wrapping them elegantly. Is it worth doing that to make the code look cleaner?

```ts
const message = [
  'this is a super long line of',
  'a long string which will all',
  'be printed on the same line.'
].join(' ');

// message = 'this is a super long line of a long string which will all be printed on the same line';
```

src/secrets/CommandMkdir.ts
tegefaulkes: When checking a single value like this you should use a `switch() case` block. It's much cleaner.

If  the default case is expected to never happen then always use a `utils.never('why this shouldn't happen')` as an invalid logic error.

```
switch(error.code){
  case 'ENOENT': message = 'whatever'
  break;
  case 'EEXIST: message = 'whatever'
  break;
  default:
    utils.never('some message');
}
```

tests/secrets/mkdir.test.ts
```diff
@@ -94,6 +95,106 @@ describe('commandMkdir', () => {
       expect(stat2.isDirectory()).toBeTruthy();
     });
   });
+  test('should fail without recursive set', async () => {
+    const vaultName = 'vault' as VaultName;
+    const dirName1 = 'dir1';
+    const dirName2 = 'dir2';
+    const dirNameNested = `${dirName1}/${dirName2}`;
```
tegefaulkes: You should be using `path.join()` when constructing paths like this rather than use a template string. Keep in mind that the fs assumes unix style paths but `path.join()` will switch based on platform. There is a utility that handles `path.join()` for the efs. You'll probably need to check the vaults code to find it.

As a general rule never construct path strings manually like this.

tests/secrets/mkdir.test.ts
```diff
@@ -94,6 +95,106 @@ describe('commandMkdir', () => {
...
+      dataDir,
+      `${vaultName}:${dirNameNested}`,
+    ];
+
+    const result = await testUtils.pkStdio([...command], {
+      env: { PK_PASSWORD: password },
+      cwd: dataDir,
+    });
+    expect(result.exitCode).toBe(0);
+    expect(result.stderr).toBeDefined();
```
tegefaulkes: An exit code of `0` indicates that the command succeeded. Any non-zero code is a command failure. Also keep in mind any messaging, error or otherwise comes out on `stderr`. The existence of output on `stderr` isn't an indication of an error.

aryanjassal: > The existence of output on `stderr` isn't an indication of an error.

Ah, I hadn't fully updated the tests.

My intention for checking if `stderr` was defined was to make sure the warning is printed. I guess I can keep the `stderr` check in only one of the tests and use explicit `toInclude` or similar check if the error string exists in `stderr`.

> An exit code of `0` indicates that the command succeeded. Any non-zero code is a command failure.

I haven't implemented this yet. I will implement it so that the program will exit with a non-zero exit code if an error occurs by exiting the process using an exit command.

aryanjassal: There might be an issue in `src/secrets/CommandEnv.ts:193`. There, we run `process.exit()` inside the command itself. From my testing, I have realised that a) doesn't run the `finally` block and b) will exit the `jest` process if tests are being run on it, with the exit code as specified in the exit command.

We need to look into it to make sure that it is intentional.

aryanjassal: I've been trying to solve this issue. Setting `process.exitCode` does not work and the process always returns an exit code of zero. We may need to have a chat about this.

aryanjassal: While not fully compliant, we will instead be throwing an error at the end if a directory operation failed. That is not the case with standard mkdir, but this is a side-effect of Polykey CLI's behaviour.

The command will now look like this:

```console
[aryanj@matrix-34xx:~]$ pk secrets mkdir vault:test vault:fail/test vault:test2
WARNING: Empty directories will not be affected by modifications to the vault state (version change, sharing/cloning, etc.)
NOENT: cannot create directory test: No such secret or directory
ErrorPolykeyCLIMakeDirectory: Failed to create one or more directories

[aryanj@matrix-34xx:~]$ echo $?
1
```

src/secrets/CommandMkdir.ts
```diff
@@ -50,16 +56,57 @@ class CommandMkdir extends CommandPolykey {
...
+              nameOrId: vault,
+              dirName: path,
+              metadata: first
+                ? { ...auth, options: { recursive: options.recursive } }
+                : undefined,
+            });
+          }
+          await writer.close();
+          return response;
+        }, meta);
```
tegefaulkes: You're never updating `first` to `false`. It won't break things given how metadata works, however it will be sending extra information in each message.

src/secrets/CommandMkdir.ts
```diff
@@ -50,16 +56,57 @@ class CommandMkdir extends CommandPolykey {
...
+            process.stderr.write(
+              `${result.code}: cannot create directory ${result.reason}: ${message}\n`,
+            );
+          }
+        }
+        if (shouldError) {
+          throw new ErrorPolykeyCLIMakeDirectory(
+            'Failed to create one or more directories',
+          );
+        }
```
tegefaulkes: `shouldError` is a poor name for this. Use something like `isErrored`.

src/secrets/CommandMkdir.ts
tegefaulkes: Don't write a warning for this to `stderr`. This information should be part of documentation and the command description instead.

src/secrets/CommandMkdir.ts
```diff
@@ -50,16 +56,57 @@ class CommandMkdir extends CommandPolykey {
...
+          }
+          await writer.close();
+          return response;
+        }, meta);
+
+        // As git does not track empty directories in version control, we need
+        // to inform the user about this.
+        process.stderr.write(
+          'WARNING: Empty directories will not be affected by modifications to the vault state (version change, sharing/cloning, etc.)\n',
         );
```
tegefaulkes: This is not how we should warn people about this behaviour. This information should be part of the command description and documentation.

src/secrets/CommandMkdir.ts
aryanjassal: https://github.com/MatrixAI/Polykey-CLI/pull/300#discussion_r1798761117 now that we are no longer sending users a message on stderr, this is no longer relevant

src/secrets/CommandMkdir.ts
aryanjassal: Okay, so I just moved this warning into the command description itself.

```ts
    this.description(
      'Create a Directory within a Vault. Empty directories are not a part of the vault and will not be shared when cloning a Vault.',
    );
```

src/secrets/CommandMkdir.ts
aryanjassal: To avoid setting the variable to false each iteration, I have done the following which is more verbose, but avoids this issue. Should I change this and keep it simple?

```ts
if (first) {
  first = false;
  await writer.write({
    nameOrId: vault,
    dirName: path,
    metadata: {
      ...auth,
      options: { recursive: options.recursive },
    },
  });
} else {
  await writer.write({ nameOrId: vault, dirName: path });
}
```

tegefaulkes: Updating a boolean every iteration of a loop is not a problem. Don't worry about micro-optimisations like that. Just keep the code here clean.

aryanjassal:
All the tasks have been done in this PR and the CI is also passing. Merging.

aryanjassal:
mkdir
[pkcli-mkdir.yml](https://github.com/user-attachments/files/17372557/pkcli-mkdir.txt)

CMCDragonkai:
> mkdir [ mkdir ](https://private-user-images.githubusercontent.com/84617406/376456567-22a9e3fc-88f9-46cf-9638-7fa96b53ca51.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MjkwNTQ0NzksIm5iZiI6MTcyOTA1NDE3OSwicGF0aCI6Ii84NDYxNzQwNi8zNzY0NTY1NjctMjJhOWUzZmMtODhmOS00NmNmLTk2MzgtN2ZhOTZiNTNjYTUxLmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDEwMTYlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQxMDE2VDA0NDkzOVomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTZjNTRiYTI0MTljNmM3NjI4MDJjNzg0YTg0ZjM0OTA3YzdiY2Y5ZmE5ZWJjM2Y0M2Q3MGI5MDIzNDNjOTE4Y2ImWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0In0.HfEGuKiV0463srP1EiAyJIQQxrwFD4fExaWADoKcWWo) [ ](https://private-user-images.githubusercontent.com/84617406/376456567-22a9e3fc-88f9-46cf-9638-7fa96b53ca51.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MjkwNTQ0NzksIm5iZiI6MTcyOTA1NDE3OSwicGF0aCI6Ii84NDYxNzQwNi8zNzY0NTY1NjctMjJhOWUzZmMtODhmOS00NmNmLTk2MzgtN2ZhOTZiNTNjYTUxLmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDEwMTYlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQxMDE2VDA0NDkzOVomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTZjNTRiYTI0MTljNmM3NjI4MDJjNzg0YTg0ZjM0OTA3YzdiY2Y5ZmE5ZWJjM2Y0M2Q3MGI5MDIzNDNjOTE4Y2ImWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0In0.HfEGuKiV0463srP1EiAyJIQQxrwFD4fExaWADoKcWWo) [pkcli-mkdir.yml](https://github.com/user-attachments/files/17372557/pkcli-mkdir.txt)

That is good, but why the background so large? And is there a way to scroll the terminal? Also speed up the actual typing. It needs to be done quickly.

CMCDragonkai:
Also the textual resolution looks worse than ascinnema...

aryanjassal:
> That is good, but why the background so large? And is there a way to scroll the terminal? Also speed up the actual typing. It needs to be done quickly.

The background is so large so we can crop it to a fitting size later on, as re-rendering it will take some time.

There is a way to scroll the terminal. It just used by actual terminal size by default, as I was recording in a vertically split terminal. I can modify the rows and columns in the recording and can change it pretty easily.

As for speed, there should definitely be a way to do that. I will have to look into how to get that working though.

> Also the textual resolution looks worse than ascinnema...

Probably because its so zoomed out. I will experiment with settings and upload a functional `config.yml` file on <b>[MatrixAI-Graph#79](https://github.com/MatrixAI/MatrixAI-Graph/issues/79)</b>

aryanjassal:
> > That is good, but why the background so large? And is there a way to scroll the terminal? Also speed up the actual typing. It needs to be done quickly.
> 
> The background is so large so we can crop it to a fitting size later on, as re-rendering it will take some time.
> 
> There is a way to scroll the terminal. It just used by actual terminal size by default, as I was recording in a vertically split terminal. I can modify the rows and columns in the recording and can change it pretty easily.
> 
> As for speed, there should definitely be a way to do that. I will have to look into how to get that working though.
> 
> > Also the textual resolution looks worse than ascinnema...
> 
> Probably because its so zoomed out. I will experiment with settings and upload a functional `config.yml` file on [MatrixAI-Graph#79](https://github.com/MatrixAI/MatrixAI-Graph/issues/79)

There aren't any actual options to modify the speed. Instead, we can specify the frame delay, which will change the delay between each input/output (a frame) to the specified time in milliseconds. Or, we can change the max idle time, which caps the time between each frame to a limit.

The textual resolution being worse than asciinema is a known issue, and something which I (or others) haven't found a solution to.

So, as an alternative, I have discovered [termsvg](https://github.com/MrMarble/termsvg), which uses asciinema as its base, records in a slightly modified format, and renders a svg of the terminal with a MacOS-like border. However, while customisation in this is not nearly as extensive as in Terminalizer, and the file format is also difficult to change and modify, it is faster and easier to render than Terminalizer and offers higher image quality.

I haven't gotten to updating the GIF yet as my focus has been on #305, but I can do this once the critical and important issues have been closed.

Issue #304, opened 2024-10-14
Specify only the vault name and infer the secret path to point to root


aryanjassal:
### Specification

Currently, we have a parser which parses the input path to ensure valid parameters. In the case of vault paths, the parser check if the path follows the format like `vaultName:secretPath`. This is not always ideal, as we want to allow specifying the vault's root directory by simply writing `vaultName`.

This is already partially implemented in the `secrets ls` command, but this needs to be made a standard. For command which do not operate on directories, like `secrets write`, this should be another error thrown from within the command, but this shouldn't be an error thrown at the parsing stage.

```
# Currently this results in Commander reporting a parser error
$ polykey secrets mkdir vaultX
error: command-argument value 'vaultX' is invalid for argument 'secretPath'. vaultX is not of the format <vaultName>:<directoryPath>[=<value>]

Usage: polykey secrets mkdir [options] <secretPath>

Create a Directory within a Vault

Arguments:
  secretPath                   Path to where the directory to be created, specified as <vaultName>:<directoryPath>

Options:
  -np, --node-path <path>      Path to Node State (default: "/home/aryanj/.local/share/polykey", env: PK_NODE_PATH)
  -pf, --password-file <path>  Path to Password
  -f, --format <format>        Output Format (choices: "human", "json", default: "human")
  -v, --verbose                Log Verbose Messages (default: 0)
  -r, --recursive              Create the directory recursively
  -ni, --node-id <id>           (env: PK_NODE_ID)
  -ch, --client-host <host>    Client Host Address (env: PK_CLIENT_HOST)
  -cp, --client-port <port>    Client Port (env: PK_CLIENT_PORT)
  -h, --help                   display help for command
```

```
# This should be handled by the command itself
$ polykey secrets mkdir vaultX
ErrorPolykeyCLIDirectoryExists: Directory / already exists
```

`ErrorPolykeyCLIX: Directory / already exists`.

### Additional context

* Related comment [**Polykey-CLI#296 (comment)**](https://github.com/MatrixAI/Polykey-CLI/issues/296#issuecomment-2409200171)

### Tasks

1. Update parser to allow returning undefined secret paths, which is to refer to the root directory
2. Raise errors where the command should not accept directories as an argument
3. Make tests for every single command ensuring regression testing against this behaviour

tegefaulkes:
@aryanjassal this is done right? Linear might have re-opened it.

aryanjassal:
[...]

PR #305, opened 2024-10-14
Allow parsing just the vault name without requiring the secret path

aryanjassal:
### Description

When specifying a secret path, currently the format needs to be `<vaultName>:<secretPath>`, otherwise a parser error is thrown. This should not be the case, and instead paths like `<vaultName>` should also be allowed, which would point to the root of the directory.

To do this, a parser for `vaultName` will also be implemented, ensuring consistency in vault's allowed characters. This is important as currently, there are no parsers for vault name, but there is a regex validation for vault names when they are being used in the secret commands. This means that inadvertently vaults can be created which cannot be actually used in any commands like `matrix.ai`.

### Issues Fixed

* Relates to <b>[Polykey-CLI#304](https://github.com/MatrixAI/Polykey-CLI/issues/304)</b> (REF ENG-425)
* Relates to <b>[Polykey-CLI#251](https://github.com/MatrixAI/Polykey-CLI/issues/251)</b> (REF ENG-379)
* Closes <b>[Polykey-CLI#310](https://github.com/MatrixAI/Polykey-CLI/issues/310)</b> (REF ENG-431)

### Tasks

- [x] 1. Create parser for vaultName
- [x] 2. Use the vaultName parser to parse the vault name consistently in vault and secret commands
- [x] 3. Make the default parser accept missing secret names
- [x] 4. Convert each command to use this new parser and substitute `'/'` if the secret path is undefined
- [x] 5. Run tests to ensure no regression
- [x] 6. Add tests for parsers
- ~~`[ ]` 7. Split tests for `vaults/scanNode.test.ts`~~ out of scope for this PR
- [x] 8. Add tests for every handler to confirm no secret to be an alias for the root directory
  - [x] `cat`
  - [x] `create`
  - [x] `dir`
  - [x] `edit`
  - [x] `env`
  - [x] `list`
  - [x] `mkdir`
  - [x] `remove`
  - [x] `rename`
  - [x] `stat`
  - [x] `write`
- [x] 9. Change `vaults list` to `vaults ls` 

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/src/secrets/CommandRemove.ts
+++ b/src/secrets/CommandRemove.ts
@@ -4,6 +4,7 @@ import * as binUtils from '../utils';
 import * as binOptions from '../utils/options';
 import * as binParsers from '../utils/parsers';
 import * as binProcessors from '../utils/processors';
+import * as errors from '../errors';
 
 class CommandRemove extends CommandPolykey {
   constructor(...args: ConstructorParameters<typeof CommandPolykey>) {
@@ -20,9 +21,17 @@ class CommandRemove extends CommandPolykey {
     this.addOption(binOptions.clientPort);
     this.addOption(binOptions.recursive);
     this.action(async (secretPaths, options) => {
-      secretPaths = secretPaths.map((path: string) =>
-        binParsers.parseSecretPathValue(path),
-      );
+      for (let i = 0; i < secretPaths.length; i++) {
+        const value: string = secretPaths[i];
+        const parsedValue = binParsers.parseSecretPath(value);
+        // The vault root cannot be deleted
+        if (parsedValue[1] == null) {
+          throw new errors.ErrorPolykeyCLIRemoveSecret(
+            'EPERM: Cannot remove vault root',
+          );
+        }
+        secretPaths[i] = parsedValue;
+      }
       const { default: PolykeyClient } = await import(
         'polykey/dist/PolykeyClient'
       );
@@ -50,28 +59,25 @@ class CommandRemove extends CommandPolykey {
           options: { nodePath: options.nodePath },
           logger: this.logger.getChild(PolykeyClient.name),
         });
-        const response = await binUtils.retryAuthentication(async (auth) => {
+        await binUtils.retryAuthentication(async (auth) => {
           const response =
             await pkClient.rpcClient.methods.vaultsSecretsRemove();
-          await (async () => {
-            const writer = response.writable.getWriter();
-            let first = true;
-            for (const [vault, path] of secretPaths) {
-              await writer.write({
-                nameOrId: vault,
-                secretName: path,
-                metadata: first
-                  ? { ...auth, options: { recursive: options.recursive } }
-                  : undefined,
-              });
-              first = false;
-            }
-            await writer.close();
-          })();
-          return response;
+          const writer = response.writable.getWriter();
+          let first = true;
+          for (const [vault, path] of secretPaths) {
+            await writer.write({
+              nameOrId: vault,
+              secretName: path,
+              metadata: first
+                ? { ...auth, options: { recursive: options.recursive } }
+                : undefined,
+            });
+            first = false;
+          }
+          await writer.close();
+          // Wait for the program to generate a response (complete processing).
+          await response.output;
         }, meta);
-        // Wait for the program to generate a response (complete processing).
-        await response.output;
       } finally {
         if (pkClient! != null) await pkClient.stop();
       }
--- a/src/secrets/CommandRename.ts
+++ b/src/secrets/CommandRename.ts
@@ -4,6 +4,7 @@ import * as binUtils from '../utils';
 import * as binOptions from '../utils/options';
 import * as binParsers from '../utils/parsers';
 import * as binProcessors from '../utils/processors';
+import * as errors from '../errors';
 
 class CommandRename extends CommandPolykey {
   constructor(...args: ConstructorParameters<typeof CommandPolykey>) {
@@ -13,13 +14,19 @@ class CommandRename extends CommandPolykey {
     this.argument(
       '<secretPath>',
       'Path to where the secret to be renamed, specified as <vaultName>:<directoryPath>',
-      binParsers.parseSecretPathValue,
+      binParsers.parseSecretPath,
     );
     this.argument('<newSecretName>', 'New name of the secret');
     this.addOption(binOptions.nodeId);
     this.addOption(binOptions.clientHost);
     this.addOption(binOptions.clientPort);
     this.action(async (secretPath, newSecretName, options) => {
+      // Ensure that a valid secret path is provided
+      if (secretPath[1] == null) {
+        throw new errors.ErrorPolykeyCLIRenameSecret(
+          'EPERM: Cannot rename vault root',
+        );
+      }
       const { default: PolykeyClient } = await import(
         'polykey/dist/PolykeyClient'
       );
--- a/src/utils/parsers.ts
+++ b/src/utils/parsers.ts
@@ -10,7 +10,6 @@ import * as nodesUtils from 'polykey/dist/nodes/utils';
 
 const vaultNameRegex = /^(?!.*[:])[ -~\t\n]*$/s;
 const secretPathRegex = /^(?!.*[=])[ -~\t\n]*$/s;
-const vaultNameSecretPathRegex = /^([\w\-\.]+)(?::([^\0\\=]+))?$/;
 const secretPathValueRegex = /^([a-zA-Z_][\w]+)?$/;
 const environmentVariableRegex = /^([a-zA-Z_]+[a-zA-Z0-9_]*)?$/;
 
@@ -67,32 +66,6 @@ function parseCoreCount(v: string): number | undefined {
   }
 }
 
-function parseSecretPathOptional(
-  secretPath: string,
-): [string, string?, string?] {
-  // E.g. If 'vault1:a/b/c', ['vault1', 'a/b/c'] is returned
-  //      If 'vault1', ['vault1, undefined] is returned
-  // splits out everything after an `=` separator
-  const lastEqualIndex = secretPath.lastIndexOf('=');
-  const splitSecretPath =
-    lastEqualIndex === -1
-      ? secretPath
-      : secretPath.substring(0, lastEqualIndex);
-  const value =
-    lastEqualIndex === -1
-      ? undefined
-      : secretPath.substring(lastEqualIndex + 1);
-  if (!vaultNameSecretPathRegex.test(splitSecretPath)) {
-    throw new commander.InvalidArgumentError(
-      `${secretPath} is not of the format <vaultName>[:<directoryPath>][=<value>]`,
-    );
-  }
-  const [, vaultName, directoryPath] = splitSecretPath.match(
-    vaultNameSecretPathRegex,
-  )!;
-  return [vaultName, directoryPath, value];
-}
-
 function parseVaultName(vaultName: string): string {
   if (!vaultNameRegex.test(vaultName)) {
     throw new commander.InvalidArgumentError(
@@ -102,19 +75,48 @@ function parseVaultName(vaultName: string): string {
   return vaultName;
 }
 
-function parseSecretPath(secretPath: string): [string, string, string?] {
-  // E.g. If 'vault1:a/b/c', ['vault1', 'a/b/c'] is returned
-  //      If 'vault1', an error is thrown
-  const [vaultName, secretName, value] = parseSecretPathOptional(secretPath);
-  if (secretName === undefined) {
+// E.g. If 'vault1:a/b/c', ['vault1', 'a/b/c', undefined] is returned
+//      If 'vault1', ['vault1', undefined, undefined] is returned
+//      If 'vault1:abc=xyz', ['vault1', 'abc', 'xyz'] is returned
+//      If 'vault1:', an error is thrown
+//      If 'a/b/c', an error is thrown
+// Splits out everything after an `=` separator
+function parseSecretPath(inputPath: string): [string, string?, string?] {
+  // The colon character `:` is prohibited in vaultName, so it's first occurence
+  // means that this is the delimiter between vaultName and secretPath.
+  const colonIndex = inputPath.indexOf(':');
+  // If no colon exists, treat entire string as vault name
+  if (colonIndex === -1) {
+    return [parseVaultName(inputPath), undefined, undefined];
+  }
+  // Calculate vaultName and secretPath
+  const vaultNamePart = inputPath.substring(0, colonIndex);
+  const secretPathPart = inputPath.substring(colonIndex + 1);
+  // Calculate contents after the `=` separator (value)
+  const equalIndex = secretPathPart.indexOf('=');
+  // If `=` isn't found, then the entire secret path section is the actual path.
+  // Otherwise, split the path section by the index of `=`. First half is the
+  // actual secret part, and the second half is the value.
+  const secretPath =
+    equalIndex === -1
+      ? secretPathPart
+      : secretPathPart.substring(0, equalIndex);
+  const value =
+    equalIndex !== -1 ? secretPathPart.substring(equalIndex + 1) : undefined;
+  // If secretPath exists but it doesn't pass the regex test, then the path is
+  // malformed.
+  if (secretPath != null && !secretPathRegex.test(secretPath)) {
     throw new commander.InvalidArgumentError(
-      `${secretPath} is not of the format <vaultName>:<directoryPath>[=<value>]`,
+      `${inputPath} is not of the format <vaultName>[:<secretPath>][=<value>]`,
     );
   }
-  return [vaultName, secretName, value];
+  // We have already tested if the secretPath is valid, and we can return it
+  // as-is.
+  const vaultName = parseVaultName(vaultNamePart);
+  return [vaultName, secretPath, value];
 }
 
-function parseSecretPathValue(secretPath: string): [string, string, string?] {
+function parseSecretPathValue(secretPath: string): [string, string?, string?] {
   const [vaultName, directoryPath, value] = parseSecretPath(secretPath);
   if (value != null && !secretPathValueRegex.test(value)) {
     throw new commander.InvalidArgumentError(
@@ -125,36 +127,7 @@ function parseSecretPathValue(secretPath: string): [string, string, string?] {
 }
 
 function parseSecretPathEnv(secretPath: string): [string, string?, string?] {
-  // The colon character `:` is prohibited in vaultName, so it's first occurence
-  // means that this is the delimiter between vaultName and secretPath.
-  const colonIndex = secretPath.indexOf(':');
-  // If no colon exists, treat entire string as vault name
-  if (colonIndex === -1) {
-    return [parseVaultName(secretPath), undefined, undefined];
-  }
-  // Calculate contents before the `=` separator
-  const vaultNamePart = secretPath.substring(0, colonIndex);
-  const secretPathPart = secretPath.substring(colonIndex + 1);
-  // Calculate contents after the `=` separator
-  const equalIndex = secretPathPart.indexOf('=');
-  const splitSecretPath =
-    equalIndex === -1
-      ? secretPathPart
-      : secretPathPart.substring(0, equalIndex);
-  const valueData =
-    equalIndex === -1 ? undefined : secretPathPart.substring(equalIndex + 1);
-  if (splitSecretPath != null && !secretPathRegex.test(splitSecretPath)) {
-    throw new commander.InvalidArgumentError(
-      `${secretPath} is not of the format <vaultName>[:<secretPath>][=<value>]`,
-    );
-  }
-  const parsedVaultName = parseVaultName(vaultNamePart);
-  const parsedSecretPath = splitSecretPath.match(secretPathRegex)?.[0] ?? '/';
-  const [vaultName, directoryPath, value] = [
-    parsedVaultName,
-    parsedSecretPath,
-    valueData,
-  ];
+  const [vaultName, directoryPath, value] = parseSecretPath(secretPath);
   if (value != null && !environmentVariableRegex.test(value)) {
     throw new commander.InvalidArgumentError(
       `${value} is not a valid environment variable name`,
@@ -263,7 +236,6 @@ export {
   validateParserToArgParser,
   validateParserToArgListParser,
   parseCoreCount,
-  parseSecretPathOptional,
   parseVaultName,
   parseSecretPath,
   parseSecretPathValue,
--- a/tests/secrets/cat.test.ts
+++ b/tests/secrets/cat.test.ts
@@ -40,7 +40,7 @@ describe('commandCatSecret', () => {
     });
   });
   test('should retrieve a secret', async () => {
-    const vaultName = 'Vault3' as VaultName;
+    const vaultName = 'vault' as VaultName;
     const vaultId = await polykeyAgent.vaultManager.createVault(vaultName);
     const secretName = 'secret-name';
     const secretContent = 'this is the contents of the secret';
@@ -60,6 +60,22 @@ describe('commandCatSecret', () => {
     });
     expect(result.exitCode).toBe(0);
     expect(result.stdout).toBe(secretContent);
+  });
+  test('should fail when reading root without secret path', async () => {
+    const vaultName = 'vault' as VaultName;
+    const vaultId = await polykeyAgent.vaultManager.createVault(vaultName);
+    const secretName = 'secret-name';
+    const secretContent = 'this is the contents of the secret';
+    await polykeyAgent.vaultManager.withVaults([vaultId], async (vault) => {
+      await vaultOps.addSecret(vault, secretName, secretContent);
+    });
+    const command = ['secrets', 'cat', '-np', dataDir, vaultName];
+    const result = await testUtils.pkStdio(command, {
+      env: { PK_PASSWORD: password },
+      cwd: dataDir,
+    });
+    expect(result.exitCode).not.toBe(0);
+    expect(result.stderr).toInclude('ErrorSecretsIsDirectory');
   });
   test('should concatenate multiple secrets', async () => {
     const vaultName = 'Vault3' as VaultName;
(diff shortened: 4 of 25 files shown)


src/utils/parsers.ts
```diff
@@ -89,25 +89,18 @@ function parseSecretPathOptional(
   return [vaultName, directoryPath, value];
 }
```
aryanjassal, line 117: Currently, I am returning `undefined` if a secret path isn't provided, and secrets can detect that to perform a separate operation if needed. Is this approach fine, or should the parser itself return `'/'` if no secret path is provided?

tegefaulkes: Have it return undefined since that part isn't actually provided. It's up to the commands themselves to decide how to handle it.

CMCDragonkai:
I think I prefer null in those scenarios? Undefined is usually meant to mean void.

tegefaulkes:
> I think I prefer null in those scenarios? Undefined is usually meant to mean void.

Undefined usually means it wasn't provided. Null would mean it was intentionally not provided. We always use undefined if it's optional.

CMCDragonkai:
I mean in TS, `undefined` is for `void` situations. I prefer null as an explicit `Maybe` type.

CMCDragonkai:
It depends, you need to review other places in the code where this has occurred.

CMCDragonkai:
Look at the other parsers.

aryanjassal:
While working on this PR, I actually discovered a bug in `secrets write`, where it doesn't properly handle writing to directories. I have made an appropriate bug report for that.

<b>[https://github.com/MatrixAI/Polykey-CLI/issues/311](Polykey-CLI#311)</b>

There could have been other issues like this which might have slipped past testing and reviewing. I will go through all the currently implemented commands and ensure that the tests properly test everything so similar bugs can't catch us by surprise.

src/utils/parsers.ts
```diff
@@ -65,40 +66,48 @@ function parseCoreCount(v: string): number | undefined {
...
+  if (!vaultNameRegex.test(vaultName)) {
     throw new commander.InvalidArgumentError(
-      `${secretPath} is not of the format <vaultName>[:<directoryPath>][=<value>]`,
+      `${vaultName} is not a valid vault name`,
     );
   }
-  const [, vaultName, directoryPath] = splitSecretPath.match(secretPathRegex)!;
-  return [vaultName, directoryPath, value];
+  // Make sure we don't accidentally return garbage data
+  return vaultName.match(vaultNameRegex)![0];
```
tegefaulkes: You shouldn't modify the vaultname here. For this parser we're just checking if its correct. It needs to remain as a string and unmodified.

src/utils/parsers.ts
```diff
@@ -65,40 +66,48 @@ function parseCoreCount(v: string): number | undefined {
...
 
-function parseSecretPath(secretPath: string): [string, string, string?] {
-  // E.g. If 'vault1:a/b/c', ['vault1', 'a/b/c'] is returned
-  //      If 'vault1', an error is thrown
-  const [vaultName, secretName, value] = parseSecretPathOptional(secretPath);
-  if (secretName === undefined) {
+// E.g. If 'vault1:a/b/c', ['vault1', 'a/b/c'] is returned
+//      If 'vault1', ['vault1, undefined] is returned
+//      If 'vault1:', an error is thrown
+//      If 'a/b/c', an error is thrown
```
tegefaulkes: This commentary needs to be updated.

src/utils/parsers.ts
```diff
@@ -65,40 +66,48 @@ function parseCoreCount(v: string): number | undefined {
...
+      ? secretPathPart
+      : secretPathPart.substring(0, equalIndex);
+  const value =
+    equalIndex === -1 ? undefined : secretPathPart.substring(equalIndex + 1);
+  if (splitSecretPath != null && !secretPathRegex.test(splitSecretPath)) {
     throw new commander.InvalidArgumentError(
-      `${secretPath} is not of the format <vaultName>:<directoryPath>[=<value>]`,
+      `${secretPath} is not of the format <vaultName>[:<secretPath>][=<value>]`,
     );
   }
```
tegefaulkes: When it comes to if statements that check a few boolean conditions at once, you always want to add a comment explaining in clear terms what you are checking for. It makes it easier to follow the code. If there was a bug in the logic which is very easy to do with boolean logic then it would be much easier to fix if the intention was known.

src/utils/parsers.ts
```diff
@@ -65,40 +66,48 @@ function parseCoreCount(v: string): number | undefined {
...
+  const vaultNamePart = secretPath.substring(0, colonIndex);
+  const secretPathPart = secretPath.substring(colonIndex + 1);
+  // Calculate contents after the `=` separator
+  const equalIndex = secretPathPart.indexOf('=');
+  const splitSecretPath =
+    equalIndex === -1
+      ? secretPathPart
+      : secretPathPart.substring(0, equalIndex);
+  const value =
+    equalIndex === -1 ? undefined : secretPathPart.substring(equalIndex + 1);
```
tegefaulkes: General formatting of ternary operators is that the main thing comes first then the fallback or secondary value comes second. 

This isn't a hard rule. Just how I usually structure it.

tegefaulkes (changes requested):

aryanjassal:
I just realised a potential issue that would crop up in the future. Lets say that I have a vault called `test`, and I have a local directory called `test`. If we add local filesystem support, then by running `pk secrets rm test --recursive`, it would be ambiguous what is being referred to, the local directory `test`, or all the vault contents under `test` (basically being an alias to `test:/`).

This same issue applies to basically all secrets commands which are expected to operate on the user's local file system. This needs to be addressed before we go about merging this PR.

@CMCDragonkai

aryanjassal:
I have observed an interesting behaviour with `secrets cat`. If we use `secrets write` to write a secret without a trailing newline (like how UNIX dictates it's standard should be), then the last line written using `secrets write` will fail to render.

```console
[aryanj@matrix-34xx:~]$ pk secrets write vault:test
some content
more content
i will press ctrl d on the next line
this line will not be shown

[aryanj@matrix-34xx:~]$ pk secrets cat vault:test
some content
more content
i will press ctrl d on the next line

[aryanj@matrix-34xx:~]$ pk secrets cat vault:test > tempfile

[aryanj@matrix-34xx:~]$ cat tempfile
some content
more content
i will press ctrl d on the next line
this line will not be shown%
```

Terminal emulators add a `%` to indicate that a file wasn't ended properly (no newline at end of file). Node's `process.stdout` actually doesn't interact with the terminal emulator's standard in and out. Rather, it writes directly to `/dev/stdout` (or something else. I'm not exactly sure on what it does, but it interacts more closely with IO than commands running on terminal emulators do). If a command ends without a newline on a terminal emulator, it knows that a new line needs to be added at the end to display the prompt and other things elegantly, so it outputs a `%` to let the users know that the command exited without a proper newline. This is strictly visual and doesn't impact the output or anything else, and is basically standard behaviour across all terminal emulators.

This is not supported by Node's `process.stdout` as it directly interacts with the standard in/out streams. As such, the terminal emulator doesn't know when the application finishes it's streams. As such, when that is the case, then the last line gets overwritten by the prompt. In my case, the prompt starts with a newline character to separate the output of the previous command better. This clears the previous line's contents and overwrites it with a blank newline.

As a workaround, currently we are manually writing a `\n` character to `stderr` so this issue doesn't happen, and doesn't break piping by providing incorrect output as compared to regular `cat`.

This still needs to be changed to be more intelligent, and detect if the last character was a newline, and do this only if it wasn't.

Another interesting discovery was made in regards to `secrets write`. When we press `<Ctrl-D>` on the keyboard, if characters have already been written on that line of stdin, then it would actually break to allow writing into a newline without actually inserting a newline. Then, when we press `<Ctrl-D>` again, that would signal the end of stream and close the stream. So, in many cases, we would need to press `<Ctrl-D>` twice in order to actually close the stream. This is not a bug with my code, but rather the behaviour of terminal emulators themselves. However, I'm not 100% certain about this, so more testing is actually needed here.

We should also consider writing a message to stderr informing the user that the message will now be written to the file, as writing to the file takes about a second, and the users won't be sure if the program needs another `<Ctrl-D>` or is just writing the file. In the future, when stdin will be directly streamed to the handler (currently it's a unary handler. this is being tracked in https://github.com/MatrixAI/Polykey/issues/822), this would become largely irrelevant, but would be good feedback to have right now.

aryanjassal:
This PR will take longer to finish, so the relevant part required for https://github.com/MatrixAI/Polykey-CLI/pull/289 has been broken off and migrated to it's own PR, which will be completed much sooner than this one.

https://github.com/MatrixAI/Polykey-CLI/pull/312

CMCDragonkai:
> I just realised a potential issue that would crop up in the future. Lets say that I have a vault called `test`, and I have a local directory called `test`. If we add local filesystem support, then by running `pk secrets rm test --recursive`, it would be ambiguous what is being referred to, the local directory `test`, or all the vault contents under `test` (basically being an alias to `test:/`).
> 
> This same issue applies to basically all secrets commands which are expected to operate on the user's local file system. This needs to be addressed before we go about merging this PR.
> 
> @CMCDragonkai

I think firstly there are only some commands that operate on both local and vault namespaces.

That's the `mv` and `cp` commands, all other commands only operate in vault namespace only. Therefore `test`, `test:` `test:/` all refer to the vault, vault and vault root directory path respectively.

So in the case of `mv` and `cp` you do have the situation of `cp test test` or `mv test test` where there's ambiguity of moving the test vault to the test local path or the test local path to the test vault.

**The solutions**:

1. **Syntax separation: `vault:` and `vault:/` always mean the vault, I think though I dislike `vault:` and should not be parsed as vault path, but should be a syntax error (that is `vault` is ok, and `vault:/` pr `vault:abc` is ok but not `vault`)**
2. **Assume that the parameter is a vault path, unless it is explicit that it is not. That is `secrets cp test test` should be assumed that `test` and `test` are both vault paths, and therefore this would an error, cannot copy vault into vault.**

Do note that you don't currently have PCC nor OCC with vault operations via RPC, until you've developed the stream-lifecycle representing an RPC conversation where the stream lifetime means 1 PCC lock. OCC requires MVCC and that's not possible without snapshots of the vault state which we cannot do at the moment.

---

Please ensure that:

```sh
pk secrets cp test test # this is an error
pk secrets cp -r testA testB # this means copying testA into testB, so `testA:/testB`
pk secrets mv test ./test # means local path
pk secrets mv test test/test # means moving test into test/test
```

Note that `mv` of a vault itself is equivalent to `cp -r` because you are not allowed to delete the vault entirely. That has to be a `vaults *` command and not a `secrets` command.

There's actually alot of constraints here. @tegefaulkes you should be working with @aryanjassal to ensure that all scenarios are being fast checked here and clearly in one big list.

I want to have a clear list here in the spec.

CMCDragonkai:
> Terminal emulators add a `%` to indicate that a file wasn't ended properly (no newline at end of file). Node's `process.stdout` actually doesn't interact with the terminal emulator's standard in and out. Rather, it writes directly to `/dev/stdout` (or something else. I'm not exactly sure on what it does, but it interacts more closely with IO than commands running on terminal emulators do). If a command ends without a newline on a terminal emulator, it knows that a new line needs to be added at the end to display the prompt and other things elegantly, so it outputs a `%` to let the users know that the command exited without a proper newline. This is strictly visual and doesn't impact the output or anything else, and is basically standard behaviour across all terminal emulators.

The `%` isn't necessarily true, it depends on the terminal emulator.

CMCDragonkai:
> As a workaround, currently we are manually writing a `\n` character to `stderr` so this issue doesn't happen, and doesn't break piping by providing incorrect output as compared to regular `cat`.

This is a bad idea.

CMCDragonkai:
> Another interesting discovery was made in regards to `secrets write`. When we press `<Ctrl-D>` on the keyboard, if characters have already been written on that line of stdin, then it would actually break to allow writing into a newline without actually inserting a newline. Then, when we press `<Ctrl-D>` again, that would signal the end of stream and close the stream. So, in many cases, we would need to press `<Ctrl-D>` twice in order to actually close the stream. This is not a bug with my code, but rather the behaviour of terminal emulators themselves. However, I'm not 100% certain about this, so more testing is actually needed here.

I've never seen this with other unix commands, you should check how you're taking in STDIN here.

CMCDragonkai:
> I have observed an interesting behaviour with `secrets cat`. If we use `secrets write` to write a secret without a trailing newline (like how UNIX dictates it's standard should be), then the last line written using `secrets write` will fail to render.

This is definitely a bug and not something terminal emulators fail at doing.

The reason for this is due to line buffering. You just need to flush the buffer explicitly at the end when the `\n` doesn't exist.

CMCDragonkai:
> > I have observed an interesting behaviour with `secrets cat`. If we use `secrets write` to write a secret without a trailing newline (like how UNIX dictates it's standard should be), then the last line written using `secrets write` will fail to render.
> 
> This is definitely a bug and not something terminal emulators fail at doing.
> 
> The reason for this is due to line buffering. You just need to flush the buffer explicitly at the end when the `\n` doesn't exist.

In fact you SHOULD ALWAYS be flushing the buffer explicitly at the end of every CLI command.

CMCDragonkai:
> Assume that the parameter is a vault path, unless it is explicit that it is not.

This a key constraint. You need to fast check this well.

The point is `abc` is just going to be interpreted as a vault path ALWAYS.

But `abc/` is not, it's going to be a local path.

Let's get some variants:

Here’s the updated table based on the clarified rules that vault paths must follow the `:/` pattern, with local paths distinguished by their explicit indicators (e.g., `./`, `/`, etc.):

| **Path**         | **Interpretation**  | **Rationale**                                             |
|------------------|---------------------|-----------------------------------------------------------|
| `abc`            | Vault path           | Simple string, assumed to be a vault path (`abc`).         |
| `abc:/`          | Vault root           | Explicitly refers to the root of the vault `abc`.          |
| `abc:/abc`       | Vault path           | Refers to subpath `abc` within the vault `abc`.            |
| `abc:/subdir/file` | Vault path         | Refers to a file within a subdirectory in the vault `abc`. |
| `abc:/xyz`       | Vault path           | Refers to `xyz` subpath in the vault `abc`.                |
| `abc:/xyz/`      | Vault path           | Refers to the `xyz/` directory inside vault `abc`.         |
| `./abc`          | Local path           | Prefixed with `./`, making it explicitly local.            |
| `abc/`           | Local path           | Ends with `/`, indicating it's a local directory.          |
| `/abc`           | Local path           | Absolute path starting with `/`, clearly a local path.     |
| `abc:xyz`        | Local path           | `:` allowed in Unix filenames, no `:/` pattern detected.   |
| `./abc/xyz`      | Local path           | Prefixed with `./`, clearly local with subdirectory `xyz`. |
| `/xyz/file`      | Local path           | Absolute path, explicitly local, referring to a file.      |

### Key Rules Recap:
1. **Vault Paths**:
   - Any path with `:/` is a vault path.
   - A simple path like `abc` is also treated as a vault path.
   
2. **Local Paths**:
   - Paths prefixed with `./`, `/`, or ending with `/` are explicitly local.
   - Paths with colons (`:`) but without the `:/` format are treated as local, given that Unix allows colons in filenames.

This table and rule set should now provide a clear, unambiguous interpretation of paths in the context of vault and local operations.

aryanjassal:
> Here’s the updated table based on the clarified rules that vault paths must follow the `:/` pattern, with local paths distinguished by their explicit indicators (e.g., `./`, `/`, etc.):

This would mean that vault names cannot have a `/` symbol, as otherwise we wouldn't be able to differentiate between the local file path or a vault name.

Also, what about `abc:.xyz`? That is technically an allowed unix path, but also an allowed vault path. How should names like this be captured?

Should we allow vault names to start with `.`? `.abc` is a local path, but `.abc:/xyz` is a valid vault path.

CMCDragonkai:
> > Here’s the updated table based on the clarified rules that vault paths must follow the `:/` pattern, with local paths distinguished by their explicit indicators (e.g., `./`, `/`, etc.):
> 
> This would mean that vault names cannot have a `/` symbol, as otherwise we wouldn't be able to differentiate between the local file path or a vault name.
> 
> Also, what about `abc:.xyz`? That is technically an allowed unix path, but also an allowed vault path. How should names like this be captured?
> 
> Should we allow vault names to start with `.`? `.abc` is a local path, but `.abc:/xyz` is a valid vault path.

Yes `/` is a reserved symbol anyway.

Under the rules then it will need to be `/.abc` to always start the resulting vault path.

Well we have to make some rules on a vault name - it would be more constrained than any path. At least 2 symbols are not allowed here: `:` and `/`.

We may also want to disallow non-printable characters and control characters but utf-8 symbols are ok.

Use a regex rule - ask chatgpt to help generate and use fastcheck and regex101 to check.

aryanjassal:
> We may also want to disallow non-printable characters and control characters but utf-8 symbols are ok.

Previously, and in this PR for now, only ASCII characters are allowed as the vault name. The previous regex followed this: `[a-zA-Z0-9_-]`.

https://github.com/MatrixAI/Polykey-CLI/blob/fd5bd1766e57c30f30986b013721d708602193dc/src/utils/parsers.ts#L11

Right now, this PR is following `[ -~^:]`, allowing all ASCII printable characters from `0x20` (space) to `0x7e` (~) except colon (:).

I will work on stricter rules and parsing for vault names in this PR. https://github.com/MatrixAI/Polykey-CLI/pull/312 will see a simple change to additionally allow dots in vault names, as that is required urgently. More extensive changes will be made in this PR.

aryanjassal:
> > Another interesting discovery was made in regards to `secrets write`. When we press `<Ctrl-D>` on the keyboard, if characters have already been written on that line of stdin, then it would actually break to allow writing into a newline without actually inserting a newline. Then, when we press `<Ctrl-D>` again, that would signal the end of stream and close the stream. So, in many cases, we would need to press `<Ctrl-D>` twice in order to actually close the stream. This is not a bug with my code, but rather the behaviour of terminal emulators themselves. However, I'm not 100% certain about this, so more testing is actually needed here.
> 
> I've never seen this with other unix commands, you should check how you're taking in STDIN here.

https://unix.stackexchange.com/a/701319
https://askubuntu.com/a/118762

It does appear to be consistent across unix commands. This behaviour is also seen on unix's `cat`, where pressing `<Ctrl-D>` sends the line into stdin, and pressing `<Ctrl-D>` again on the same line will close stdin. This means pressing `<Ctrl-D>` twice on the same line is required if a newline is not desired at the end of the output. Otherwise, add a newline, and press `<Ctrl-D>` once to close stdin. This is exactly how `secrets write` behaves.

@CMCDragonkai

aryanjassal:
> ### Key Rules Recap:
> 1. **Vault Paths**:
>    
>    * Any path with `:/` is a vault path.
>    * A simple path like `abc` is also treated as a vault path.
> 2. **Local Paths**:
>    
>    * Paths prefixed with `./`, `/`, or ending with `/` are explicitly local.
>    * Paths with colons (`:`) but without the `:/` format are treated as local, given that Unix allows colons in filenames.
> 
> This table and rule set should now provide a clear, unambiguous interpretation of paths in the context of vault and local operations.

The rules for local paths aren't relevant for this PR, as its scope is vault paths. I will make a new issue for implementation of local paths, and input all this information there.

src/utils/parsers.ts
aryanjassal: I was just trying to keep things consistent as in the condition before, it had a check like `equalIndex === -1`. Now, it looks like this:

```ts
  const secretPath =
    equalIndex === -1
      ? secretPathPart
      : secretPathPart.substring(0, equalIndex);
  const value =
    equalIndex !== -1 
      ? secretPathPart.substring(equalIndex + 1) 
      : undefined;
```

While there is nothing wrong with this approach, the conditions of consequent lines testing something very similar are flipped, which could also introduce some confusion.

I was kinda torn on how to handle this, but for now, I've implemented it the way you have recommended.

aryanjassal:
> The reason for this is due to line buffering. You just need to flush the buffer explicitly at the end when the `\n` doesn't exist.
>
> In fact you SHOULD ALWAYS be flushing the buffer explicitly at the end of every CLI command.

I tried to add a `process.stdout.end()`, which is designed to flush the stream, but to no avail. The only thing which worked was adding the newline at the end in `stderr`.

I'll make a separate issue for this, as I don't think that this can be done in this PR.

https://github.com/MatrixAI/Polykey-CLI/issues/316

aryanjassal:
As ENG-431 (#310) is a small fix, I will address that in this PR too.

src/secrets/CommandRemove.ts
```diff
@@ -20,9 +21,17 @@ class CommandRemove extends CommandPolykey {
...
-      );
+      for (let i = 0; i < secretPaths.length; i++) {
+        const value: string = secretPaths[i];
+        const parsedValue = binParsers.parseSecretPath(value);
+        // The vault root cannot be deleted
+        if (parsedValue[1] == null) {
+          throw new errors.ErrorPolykeyCLIRemoveSecret(
+            'EPERM: Cannot remove vault root',
+          );
+        }
```
aryanjassal, line 32: I am doing this here manually, as the generated errors aren't really descriptive. This needs to be changed in Polykey, but is not immediately required for this PR.

src/secrets/CommandRename.ts
```diff
@@ -13,13 +14,19 @@ class CommandRename extends CommandPolykey {
...
     this.addOption(binOptions.nodeId);
     this.addOption(binOptions.clientHost);
     this.addOption(binOptions.clientPort);
     this.action(async (secretPath, newSecretName, options) => {
+      // Ensure that a valid secret path is provided
+      if (secretPath[1] == null) {
+        throw new errors.ErrorPolykeyCLIRenameSecret(
+          'EPERM: Cannot rename vault root',
+        );
+      }
```
aryanjassal, line 29: Same here.

tests/secrets/cat.test.ts
```diff
@@ -61,6 +61,22 @@ describe('commandCatSecret', () => {
...
+    await polykeyAgent.vaultManager.withVaults([vaultId], async (vault) => {
+      await vaultOps.addSecret(vault, secretName, secretContent);
+    });
+    const command = ['secrets', 'cat', '-np', dataDir, vaultName];
+    const result = await testUtils.pkStdio(command, {
+      env: { PK_PASSWORD: password },
+      cwd: dataDir,
+    });
+    expect(result.exitCode).not.toBe(0);
+    expect(result.stderr).toInclude('ErrorSecretsIsDirectory');
```
aryanjassal, line 78: The error formatting is currently not consistent, as we recently started using error objects. This has only been implemented for `mkdir` and not for other handlers. An issue already exists to track this, and this will be done in other PRs.

tests/secrets/list.test.ts
```diff
@@ -41,124 +40,121 @@ describe('commandListSecrets', () => {
...
-      const result = await testUtils.pkStdio([...command], {
-        env: {
-          PK_PASSWORD: password,
-        },
-        cwd: dataDir,
-      });
-      expect(result.exitCode).toBe(0);
-      expect(result.stdout).toBe('MySecret1\nMySecret2\nMySecret3\n');
-    },
-    globalThis.defaultTimeout * 2,
```
aryanjassal, line 81: When I ran the tests, I saw no need for additional timeout, so I removed this line. This applies to all secrets subcommand tests, as most of them completed within 3 seconds, with longer ones taking up to 5, which is well under the 20 second timeout limit.

tests/secrets/mkdir.test.ts
```diff
@@ -56,6 +56,20 @@ describe('commandMkdir', () => {
       expect(stat.isDirectory()).toBeTruthy();
     });
   });
+  test('should fail when provided only the vault name', async () => {
+    const vaultName = 'vault' as VaultName;
+    const vaultId = await polykeyAgent.vaultManager.createVault(vaultName);
+    command = ['secrets', 'mkdir', '-np', dataDir, vaultName];
+    const result = await testUtils.pkStdio([...command], {
```
aryanjassal: This needs to change to `command` instead of `[...command]`. This actually needs to be consistent throughout the secrets tests.

tests/utils.test.ts
```diff
@@ -302,43 +295,87 @@ describe('bin/utils', () => {
...
+    .array(singleSecretPathArb, { minLength: 1, maxLength: 5 })
+    .map((segments) => path.join(...segments));
+  const valueFirstCharArb = fc.char().filter((c) => /^[a-zA-Z_]$/.test(c));
+  const valueRestCharArb = fc.stringOf(
+    fc.char().filter((c) => /^[\w]$/.test(c)),
+    { minLength: 1, maxLength: 100 },
+  );
+  const valueDataArb = fc
+    .tuple(valueFirstCharArb, valueRestCharArb)
+    .map((components) => components.join(''));
```
aryanjassal, line 350: This probably isn't the best way to generate testing data, but this is the only way I found to ensure the tested data will be valid at all times. Instead of using the actual regex, I might need to use some other arb rule, but it would become pretty complex pretty fast.

aryanjassal (commented):
This PR is more or less complete. I've put down some of my thoughts in my review, which would be useful for reference later.

src/secrets/CommandRemove.ts
aryanjassal: This is actually being tracked in https://github.com/MatrixAI/Polykey/issues/826

tests/secrets/cat.test.ts
aryanjassal: The tracked issue is https://github.com/MatrixAI/Polykey/issues/826

tests/secrets/mkdir.test.ts
aryanjassal: Secrets env and secrets mkdir had this issue. That has now been resolved. This has not been addressed for the vault tests, as they are out of scope for this.

aryanjassal:
All the tasks for this PR have been done. Brian and I have sanity checked the commands and ensured that they all work as expected. I will merge this PR and make a new one to address all the issues that cropped up during the development of this PR.

tests/utils.test.ts
CMCDragonkai: Regex is fine. What's wrong with regex?

tests/secrets/list.test.ts
CMCDragonkai: Usually we preserve the `globalThis.defaultTimeout` because that's the default.

Issue #38, opened 2023-10-19
Remove any non-deterministic use of `sleep()` in tests


tegefaulkes:
### Specification

We need to remove any non-deterministic use of `sleep()` in tests. Most if not all of these will be sleeps waiting for key change propagation to happen. 

To address this we need some way of knowing that the keys have finished changing. https://github.com/MatrixAI/Polykey/issues/594 addresses this in `Polykey`, but we need a change here.

The keys renew and reset commands need to wait for key changes to fully propagate before ending. This will be updated in the client handlers on `Polykey`.

### Additional context

* Blocked by https://github.com/MatrixAI/Polykey/issues/594
* Related https://github.com/MatrixAI/Polykey/issues/551

### Tasks

1. Remove any usage of `sleep()` waiting for keys changes and replace it with waiting for an event.

CMCDragonkai:
Some uses of sleep is to simulate time passing. However generally anything that relies on time should parameterise it so that the caller can pass in what it wants the time to be. This is a good practice not only for testing, but also for ensuring a bunch of operations is recorded with the same time.

CMCDragonkai:
That parameter can default to the now though for convenience. Removed a bunch of sleeps from `NodeGraph.test.ts` would reduce testing time by a significant amount. 80%

PR #320, opened 2024-10-29
Properly display non-Polykey errors instead of printing `undefined`

### LINKED ISSUE #311: Some `secrets` commands throw an `undefined` error
### Describe the bug

When writing to a directory, the command should provide appropriate feedback bail out. This should either be done immediately (by preemptively checking if the target is a valid file path or not), or later, by raising an appropriate error. 

Currently, it just returns this output, which is not really explicit or apparent in what went wrong:

```console
[aryanj@matrix-34xx:~]$ pk secrets write vault:dir
writing to dir <Ctrl-D>
undefined

[aryanj@matrix-34xx:~]$ echo $?
255
```

### To Reproduce

1. Try writing to a directory using stdin
2. Observe issue

### Expected behavior

As in `secrets cat`, an error could be raised:
```console
[aryanj@matrix-34xx:~]$ pk secrets cat vault:dir     
ErrorSecretsIsDirectory: dir: Is a directory
```

### Platform
 - Device: Dell Precision 3480
 - OS: NixOS
 - Version: `["0.10.0","1.14.0","1","1"]`

### Notify maintainers

@aryanjassal

### LINKED ISSUE #315: `secrets edit` should maintain the websocket as long as the editor session is alive
### Specification

When using `secrets edit`, it will timeout if the editor session isn't finished fast enough. I opened up the `secrets edit` and waited for a while, after trying to save, it failed:

```
cmcdragonkai ➜ matrix-framework-13-ryzen-7040  ➜ ~/Projects/Polykey-CLI
 $ ./dist/polykey.js secrets ed nix:/nix.conf
ERROR:polykey.PolykeyClient.WebSocketClient:ErrorWebSocketConnectionKeepAliveTimeOut: WebSocket Connection reached keep-alive timeout - 
ERROR:polykey.PolykeyClient.WebSocketClient.WebSocketConnection 0:ErrorWebSocketConnectionKeepAliveTimeOut: WebSocket Connection reached keep-alive timeout - 
ErrorWebSocketConnectionNotRunning
```

We need to make the websocket connection alive for as long as the editor session is alive, it shouldn't fail.

That being said it's also important not to start a vault commit transaction until the editor has in fact finished writing. We don't want this to block other secret operations. Concurrent testing needs to be checked too.

### Tasks

1. Use a stream life cycle to be maintained while the editor is there, it shouldn't expire on the keep alive timeout.
2. Test that secrets edit can work for a long running edit.
3. Test that while the editor is open it doesn't cause blocked concurrent secrets operations.

@tegefaulkes @aryanjassal

aryanjassal:
### Description

In Polykey, any non-Polykey errors get serialised as raw JSON, and are not able to be properly converted back into the original errors, or be printed on the output.

The only output we get if a non-Polykey error propagates through is `undefined` in the output with a non-zero exit code.

This PR should combat this by wrapping it in `ErrorPolykeyCLIUnexpectedError`, then rendering that error instead. In the future, this should help with identifying exactly what went wrong by providing more context instead of printing a useless `undefined` as the error message.

As #315 is a quick fix, I will make that change here too.

### Issues Fixed

* Fixes #311 (REF ENG-434)
* Fixes #315 (REF ENG-440)

### Tasks

- [x] 1. Wrap the malformed error
- [x] 2. Print the wrapped error using the errorFormatter
- [x] 3. Allow webstream to keep pinging in the background (`exec` async)
- [x] 4. Run all tests to ensure no regression

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased


aryanjassal:
A separate test is needed to ensure keeping the secrets edit open for around 60 seconds should not cause a websocket connection error.

Something like this can be used:
```ts
test(
  'timed test', 
  async () => {
    // launch the editor
    // sleep for 60 seconds
    // close the editor
  },
  timeout=70000
);
```

tegefaulkes:
We want to avoid using sleeps in tests generally now. https://github.com/MatrixAI/Polykey-CLI/issues/38

If you can, use promises to wait for the timeout to happen before proceeding.

aryanjassal:
> We want to avoid using sleeps in tests generally now. #38
> 
> If you can, use promises to wait for the timeout to happen before proceeding.

Well the sleep won't be in the actual TS test. Instead, the sleep will be inside the command which will be executed by the command as the `EDITOR` alias. This will simulate the editor being open for a duration without the JS/TS being blocked by a sleep.

Is this also troublesome, or would this be fine?

tegefaulkes:
Its better if we don't. Just have the editor scrip wait for an input on stdin. We can use that as a semaphore for the next step and then control timing from the test. Also look into using `pkExpect` for executing it.

CMCDragonkai:
Never rely on sleep. It's non-deterministic.

aryanjassal:
Currently, I am wrapping any unexpected errors in `ErrorPolykeyCLIUnexpectedError` with the actual offending error as the `data`, and then rendering the error as usual. This probably isn't the best way to do this, so I need some feedback on this.

This is the only real major change being done, so once this is addressed, then we should be able to get this merged pretty quickly.

aryanjassal:
Currently, the errors look like this. The code for determining the cause string was taken from https://github.com/MatrixAI/js-rpc/pull/69. This is what the undefined errors would look like now.

Note that for testing, I am inducing an undefined error as such: `throw 'testing'`

```
-> npm run polykey -- secrets write vault:test

> polykey-cli@0.13.0 polykey
> ts-node src/polykey.ts secrets write vault:test

z
ErrorPolykeyCLIUnexpectedError: Unexpected error - An unexpected error occurred
  timestamp	Mon Nov 11 2024 12:12:18 GMT+1100 (Australian Eastern Daylight Time)
  cause: Literal: testing
```

CMCDragonkai:
> Currently, the errors look like this. The code for determining the cause string was taken from [MatrixAI/js-rpc#69](https://github.com/MatrixAI/js-rpc/pull/69). This is what the undefined errors would look like now.
> 
> Note that for testing, I am inducing an undefined error as such: `throw 'testing'`
> 
> ```
> -> npm run polykey -- secrets write vault:test
> 
> > polykey-cli@0.13.0 polykey
> > ts-node src/polykey.ts secrets write vault:test
> 
> z
> ErrorPolykeyCLIUnexpectedError: Unexpected error - An unexpected error occurred
>   timestamp	Mon Nov 11 2024 12:12:18 GMT+1100 (Australian Eastern Daylight Time)
>   cause: Literal: testing
> ```

Why do we need a `Literal: ` prefix? Usually the literal value is a valid cause.

aryanjassal:
> Why do we need a `Literal: ` prefix? Usually the literal value is a valid cause.

I have done this to be consistent and clear. If we throw an error like `throw 'ReadableStream'`, then it is difficult to differentiate a string literal from an object. This is needed as we can throw anything, including objects, in TS.

With the prefix, it is much clearer:
```
ErrorPolykeyCLIUnexpectedError: Unexpected error - An unexpected error occurred
  timestamp	Mon Nov 11 2024 12:12:18 GMT+1100 (Australian Eastern Daylight Time)
  cause: Literal: ReadableStream
```

and

```
ErrorPolykeyCLIUnexpectedError: Unexpected error - An unexpected error occurred
  timestamp	Mon Nov 11 2024 12:12:18 GMT+1100 (Australian Eastern Daylight Time)
  cause: Object: ReadableStream
```

src/utils/utils.ts
```diff
@@ -425,20 +425,29 @@ function outputFormatterJson(json: string): string {
...
+        err.metadata != null
+      ) {
         output += '\n';
         for (const [key, value] of Object.entries(err.metadata)) {
           output += `${indent}${key}\t${value}\n`;
         }
+      } else {
+        output += '\n';
       }
       output += `${indent}timestamp\t${err.timestamp}\n`;
```
aryanjassal: Is the timestamp a part of the metadata? This will show for both `PolykeyCLI` errors and `PolykeyRemote` errors. I'm not sure if we need the entire exact timestamp of the error.

src/secrets/CommandCat.ts
```diff
@@ -80,28 +84,41 @@ class CommandGet extends CommandPolykey {
...
           // Print out incoming data to standard out
           let hasErrored = false;
           for await (const chunk of response.readable) {
-            if (chunk.error) {
+            if (chunk.type === 'error') {
               hasErrored = true;
-              process.stderr.write(chunk.error);
+              // TypeScript cannot properly decode this type, so we need to use
+              // the `as` keyword so it doesn't throw errors.
+              const error = chunk as ErrorMessage;
```
aryanjassal: We need to use the `as` keyword in every `secrets` command which needs to continue on error, like `mkdir`, `cat`, etc. and this can cause a sprawl of the `as` keyword. The keyword should be used as little as possible, but this way, it is actually being used quite often. I'm not sure how good is this idea in the long term.

src/secrets/CommandRemove.ts
```diff
@@ -59,25 +52,60 @@ class CommandRemove extends CommandPolykey {
...
+                process.stderr.write(
+                  `rm: cannot remove '${error.reason}': No such file or directory\n`,
+                );
+                continue;
+              } else if (error.code === 'EINVAL') {
+                // Attempt to delete vault root
+                process.stderr.write(
+                  `rm: cannot remove '${error.reason}': Permission denied\n`,
+                );
+              } else {
```
aryanjassal: When attempting to remove the vault root directory, the `VaultOps` operation throws an `EncryptedEFS` error with the code `'EINVAL'` for an invalid operation. For now, I have just captured this invalid operation as an output and am rendering it to the user as if they were trying to operate on a write-protected directory.

aryanjassal (commented):
Just some key points I wanted to mention.

src/polykey.ts
```diff
@@ -75,10 +75,14 @@ async function polykeyAgentMain(): Promise<number> {
       process.exitCode = e.exitCode;
     } else {
       // Unknown error, this should not happen
+      const wrappedError = new binErrors.ErrorPolykeyCLIUnexpectedError(
+        'An unexpected error occurred',
+        { cause: e },
+      );
```
tegefaulkes: The message needs to describe what the error was. Trust me we're going to want that because when just logging the problem we'll just get `ErrorPolykeyCLIUnexpectedError: An unexpected error occurred` message which will be absolutely useless for debugging. We'd have to trigger the error with modifications to the source just to have any idea what happened.

Everywhere you wrap an error like this you need to format the error message with what the wrapped error was.

src/polykey.ts
```diff
@@ -217,10 +221,14 @@ async function polykeyMain(argv: Array<string>): Promise<number> {
...
```
tegefaulkes: Same here, need a message that describes the actual error.

src/secrets/CommandCat.ts
tegefaulkes: The code doesn't build or run right now so I can't sanity check any alternative easily.

The type narrowing should work here. You might have to do something slightly different here or it might just be a language server problem.

src/utils/utils.ts
tegefaulkes: I'm not sure but that's something you can verify by testing it right?

As for why its included I'm not sure. IT would make it easy to verify the exact moment of the error across front and backend if you needed to compare it. But if its not causing and issue here and not in scope for the PR then don't worry about it.

tegefaulkes (changes requested):

aryanjassal:
All the tasks for this PR has been completed and the CI is also passing. Merging as completed.

aryanjassal:
All the tasks have been completed and the CI is also passing. Merging as completed.

CMCDragonkai:
Remember errors have the name, description and message. The message is instance context while description is static.

Issue #325, opened 2024-11-11
`secrets env` needs `--` to be passed twice to parse the command


aryanjassal:
### Specification

When running the `secrets env` command in the tests, we only need to pass `--` once, and that will delimit the commands and the vault names.

However, this is not the case when running the program on the CLI, as we need to write `--` twice to get it to work. For example, the following will work in the tests but not in CLI.

```bash
# In jest
[aryanj@matrix-34xx:~]$ polykey secrets env vault -- echo 'hi'
TESTING='top secret'
hi
 
# In shell
[aryanj@matrix-34xx:~]$ polykey secrets env vault -- echo 'hi'
ErrorPolykeyRemote: Remote error from RPC call
  ...
  cause: ErrorVaultsVaultUndefined: Vault does not exist
 
[aryanj@matrix-34xx:~]$ polykey secrets env vault -- -- echo 'hi'
TESTING='top secret'
hi
```

It is difficult to see which vault was the offender for commands containing multiple vaults, like `secrets` commands. For those commands at least, the `ErrorVaultsVaultUndefined` should also return the name of the offending vault. For example, a command like this can be difficult to find the invalid vault.

```bash
[aryanj@matrix-34xx:~]$ polykey secrets cat test.vault:file1 anothr.vault:file2
ErrorPolykeyRemote: Remote error from RPC call
  ...
  cause: ErrorVaultsVaultUndefined: Vault does not exist
```

Without reading the command properly, the user won't know that the offending vault was a typo in `another`.

### Additional context

* Needing `program.enablePositionalArguments` enabled to use `command.passThroughOptions`, but enabling positional arguments breaks the commands: [**Polykey-CLI#325**](https://linear.app/matrix-ai/issue/ENG-466/secrets-env-needs-to-be-passed-twice-to-parse-the-command#comment-8c1911de)  and [**Polykey-CLI#329 (comment)**](https://github.com/MatrixAI/Polykey-CLI/pull/329#discussion_r1843149664)
* Upstream discussion on a solution without needing to enable positional arguments: [**commander.js#2281**](https://github.com/tj/commander.js/issues/2281)

### Tasks

1. Allow `--` once in the CLI to run commands as opposed to twice
2. Provide feedback by listing which vault was incorrect

aryanjassal:
When printing `process.argv`, we can see that `--` is left to be parsed by the program. All the output is obtained when running this command.

```console
[...]

PR #329, opened 2024-11-12
Remove newlines by default when exporting contents using `secrets env`

### LINKED ISSUE #296: Update `secrets env` to omit single trailing `\n` from files which automatically include it
### Specification

Currently, most Unix files, and by extension, all files inside Polykey's vault will have a trailing `\n`. This is the standard way to end a file in Unix (see [**this question**](https://stackoverflow.com/questions/14171254/why-would-vim-add-a-new-line-at-the-end-of-a-file/31426524#31426524)). This poses an issue, as all secrets will also have this newline when it might not be intentional, breaking applications (see additional context below).

A long-term solution would require a vault schema to dictate the formatting of the secrets (see [**Polykey#222**](https://github.com/MatrixAI/Polykey/issues/222)). However, that is still some time away, and this issue needs to be solved before that can be merged. To do this, we can implement a flag which would explicitly tell the command to preserve the single trailing `\n` at the end of the file, otherwise the default behaviour would be to exclude it.

```bash
[aryanj@matrix-34xx:~]$ polykey secrets env vault:/ABC -el vault2:/ABC -el vault3:ABC -el -- command arg1
```

Keep in mind that **only a single, trailing newline must be removed**. Trailing white space must be preserved, and any trailing white space but the last one should also be preserved, unless the user explicitly specifies that the newline can be kept.

While we technically can use `polykey secrets write` to write without entering the newline, it is not ideal, as a vault is a file system, and in most cases, users will paste their secrets using an editor, which would add the newline at the end. This is why we need this to work right now.

### Additional context

> Line break is bad. I have an example, things like passwords to Postgres databases don't like newlines, and in fact if you use them, it will just sometimes silently fail the authentication.
>
> So you have to be careful about whether the secret string actually has a newline or not.
>
> This has implications on things like secrets write command. So for example, the standard `EOF` character is `\n` for saved files on disk. That's by convention. However if you are saving a password into a vault, and we store it as a file, then that password may not have a `\n` at the end of the string. Leaving a `\n` as convention then becomes ambiguous.
>
> Suppose you have something like `./DB_PASSWORD` and inside is `helloworld\n` (due to `\n` conventions and even most commands like `echo helloworld > ./DB_PASSWORD`, by default this will usually write a newline along with it). To get it without newline you do `echo -n helloworld >./DB_PASSWORD`. On my terminal that shows up as `helloworld%` after `cat`
>
> IMO, we need to be quite strict about this... which is that whatever comes in via `stdin` should be preserved as-is, and to not add newlines into files that isn't supposed to have it.
>
> The problem comes from implicit newlines. For example, `polykey secrets ed ./DB_PASSWORD`. If this opens up `vim`, it's likely `vim` will save the newline `EOF` - depending on the configuration of `vim`. Then later with `polykey secrets cat ./DB_PASSWORD`, this will produce everything including the newline. If this is injected using `polykey secrets env`, then you may have a newline in the output like `DB_PASSWORD='helloworld\n'`, which may be inaccurate.
>
> I think this problem fundamentally derives from crossing 2 abstractions - the file system abstraction and the key-value string abstraction
>
> File systems/files usually use newlines to indicate end of a line and end of a file, many tools operating on files will naturally automate this. But many secrets are key value strings, and therefore don't assume newlines would exist. Like for example when you hit `ENTER` on the keyboard in a text input, it's a newline, but when you hit `ENTER` for a text field input, it means save it/input it.
>
> So we have this situation whether in one context, vaults are file systems, but in another context they are key-value strings
>
> We need to figure out an elegant way of dealing with the cross over. Right now we can use `polykey secrets write` to create the secret without the newline. Later this would be solved by a vaults schema.
>
> For vaults schema to solve it, it would need to have trim constraints. But, I think we need to document the kind of workflows involving compositions with various CLI commands where this can be an issue.
>
> \- Conversation between @cmcdragonkai and @tegefaulkes

* Discussion about vault schema in [**Polykey#222**](https://github.com/MatrixAI/Polykey/issues/222)
* `vim` adding newlines by default at the end of files in this [**Stackoverflow question**](https://stackoverflow.com/questions/14171254/why-would-vim-add-a-new-line-at-the-end-of-a-file/31426524#31426524).

### Tasks

1. Automatically trim single trailing newline from all secrets
2. Add an option to disable this behaviour by default (ideally this should apply to each secret path rather than all the provided secrets)

aryanjassal:
### Description

The standard to end all files in Unix is adding a newline at the end of the file. All editors, including `vim` do it. However, adding a stray newline at the end of a secret, like a password, can have adverse effects.

As such, this PR will, by default, trim the single trailing newline, with an option to disable it if needed.

### Issues Fixed

* Fixes #296 (REF ENG-416)
* Related to #325 (REF ENG-466)

### Tasks

- [x] 1. Remove newlines by default
- [x] 2. Add a new option to allow preserving newlines
- [x] 3. Add tests ensuring this behaviour

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/src/errors.ts
+++ b/src/errors.ts
@@ -84,6 +84,11 @@ class ErrorPolykeyCLIUnexpectedError<T> extends ErrorPolykeyCLI<T> {
   exitCode = sysexits.SOFTWARE;
 }
 
+class ErrorPolykeyCLISubprocessFailure<T> extends ErrorPolykeyCLI<T> {
+  static description = 'A subprocess failed to exit gracefully';
+  exitCode = sysexits.UNKNOWN;
+}
+
 class ErrorPolykeyCLINodePath<T> extends ErrorPolykeyCLI<T> {
   static description = 'Cannot derive default node path from unknown platform';
   exitCode = sysexits.USAGE;
@@ -191,6 +196,7 @@ export {
   ErrorPolykeyCLIUncaughtException,
   ErrorPolykeyCLIUnhandledRejection,
   ErrorPolykeyCLIUnexpectedError,
+  ErrorPolykeyCLISubprocessFailure,
   ErrorPolykeyCLIAsynchronousDeadlock,
   ErrorPolykeyCLINodePath,
   ErrorPolykeyCLIClientOptions,
--- a/src/identities/CommandDiscover.ts
+++ b/src/identities/CommandDiscover.ts
@@ -103,8 +103,8 @@ class CommandDiscover extends CommandPolykey {
                   case 'queued':
                     queuedSet.add(vertex);
                     break;
-                  case 'processed':
-                  case 'cancelled':
+                  case 'processed': // Fallthrough
+                  case 'cancelled': // Fallthrough
                   case 'failed':
                     queuedSet.delete(vertex);
                     break;
--- a/src/secrets/CommandEnv.ts
+++ b/src/secrets/CommandEnv.ts
@@ -1,11 +1,12 @@
 import type PolykeyClient from 'polykey/dist/PolykeyClient';
+import type { ParsedSecretPathValue } from '../types';
 import path from 'path';
 import os from 'os';
 import * as utils from 'polykey/dist/utils';
+import CommandPolykey from '../CommandPolykey';
 import * as binProcessors from '../utils/processors';
 import * as binUtils from '../utils';
 import * as binErrors from '../errors';
-import CommandPolykey from '../CommandPolykey';
 import * as binOptions from '../utils/options';
 import * as binParsers from '../utils/parsers';
 
@@ -14,7 +15,7 @@ class CommandEnv extends CommandPolykey {
     super(...args);
     this.name('env');
     this.description(
-      `Run a command with the given secrets and env variables using process replacement. If no command is specified then the variables are printed to stdout in the format specified by env-format.`,
+      `Run a command with the given secrets and env variables. If no command is specified then the variables are printed to stdout in the format specified by env-format.`,
     );
     this.addOption(binOptions.nodeId);
     this.addOption(binOptions.clientHost);
@@ -22,18 +23,14 @@ class CommandEnv extends CommandPolykey {
     this.addOption(binOptions.envFormat);
     this.addOption(binOptions.envInvalid);
     this.addOption(binOptions.envDuplicate);
+    this.addOption(binOptions.preserveNewline);
     this.argument(
       '<args...>',
-      'command and arguments formatted as [envPaths...][-- cmd [cmdArgs...]]',
+      'command and arguments formatted as <envPaths...> -- [ cmd [cmdArgs...]]',
       binParsers.parseEnvArgs,
     );
-    this.passThroughOptions(); // Let -- pass through as-is to parse as delimiter for cmd
     this.action(
-      async (
-        args: [Array<[string, string?, string?]>, Array<string>],
-        options,
-      ) => {
-        args[1].shift();
+      async (args: [Array<ParsedSecretPathValue>, Array<string>], options) => {
         const { default: PolykeyClient } = await import(
           'polykey/dist/PolykeyClient'
         );
@@ -46,13 +43,22 @@ class CommandEnv extends CommandPolykey {
           envDuplicate: 'keep' | 'overwrite' | 'warn' | 'error';
           envFormat: 'auto' | 'unix' | 'cmd' | 'powershell' | 'json';
         } = options;
+        // Populate a set with all the paths we want to preserve newlines for
+        const preservedSecrets = new Set<string>();
+        for (const [vaultName, secretPath] of options.preserveNewline) {
+          // The vault name is guaranteed to have a value.
+          // If a secret path is undefined, then the newline preservation was
+          // targeting the secrets of the entire vault. Otherwise, the target
+          // was a single secret.
+          if (secretPath == null) preservedSecrets.add(vaultName);
+          else preservedSecrets.add(`${vaultName}:${secretPath}`);
+        }
         // There are a few stages here
         // 1. parse the desired secrets
         // 2. obtain the desired secrets
         // 3. switching behaviour here based on parameters
         //   a. exec the command with the provided env variables from the secrets
         //   b. output the env variables in the desired format
-
         const [envVariables, [cmd, ...argv]] = args;
         const clientOptions = await binProcessors.processClientOptions(
           options.nodePath,
@@ -160,7 +166,26 @@ class CommandEnv extends CommandPolykey {
                     utils.never();
                 }
               }
-              envp[newName] = secretContent;
+
+              // Find if we need to preserve the newline for this secret
+              let preserveNewline = false;
+              // If only the vault name is specified to be preserved, then
+              // preserve the newlines of all secrets inside the vault.
+              // Otherwise, if a full secret path has been specified, then
+              // preserve that secret path.
+              if (
+                preservedSecrets.has(nameOrId) ||
+                preservedSecrets.has(`${nameOrId}:${newName}`)
+              ) {
+                preserveNewline = true;
+              }
+
+              // Trim the single trailing newline if it exists
+              if (!preserveNewline && secretContent.endsWith('\n')) {
+                envp[newName] = secretContent.slice(0, -1);
+              } else {
+                envp[newName] = secretContent;
+              }
               envpPath[newName] = {
                 nameOrId,
                 secretName,
@@ -175,14 +200,26 @@ class CommandEnv extends CommandPolykey {
           // Here we want to switch between the different usages
           const platform = os.platform();
           if (cmd != null) {
-            // If a cmd is| provided then we default to exec it
+            // If a cmd is provided then we default to exec it
             switch (platform) {
-              case 'linux':
-              // Fallthrough
+              case 'linux': // Fallthrough
               case 'darwin':
                 {
                   const { exec } = await import('@matrixai/exec');
-                  exec.execvp(cmd, argv, envp);
+                  try {
+                    exec.execvp(cmd, argv, envp);
+                  } catch (e) {
+                    if ('code' in e && e.code === 'GenericFailure') {
+                      throw new binErrors.ErrorPolykeyCLISubprocessFailure(
+                        `Command failed with error ${e}`,
+                        {
+                          cause: e,
+                          data: { command: [cmd, ...argv] },
+                        },
+                      );
+                    }
+                    throw e;
+                  }
                 }
                 break;
               default: {
--- a/src/utils/parsers.ts
+++ b/src/utils/parsers.ts
@@ -1,5 +1,6 @@
 import type { Host, Hostname, Port } from 'polykey/dist/network/types';
 import type { SeedNodes } from 'polykey/dist/nodes/types';
+import type { ParsedSecretPathValue } from '../types';
 import commander from 'commander';
 import * as validationUtils from 'polykey/dist/validation/utils';
 import * as validationErrors from 'polykey/dist/validation/errors';
@@ -56,9 +57,9 @@ function parseCoreCount(v: string): number | undefined {
   switch (v) {
     case 'all':
       return 0;
-    case 'none':
-    case 'no':
-    case 'false':
+    case 'none': // Fallthrough
+    case 'no': // Fallthrough
+    case 'false': // Fallthrough
     case 'null':
       return undefined;
     default:
@@ -81,7 +82,7 @@ function parseVaultName(vaultName: string): string {
 //      If 'vault1:', an error is thrown
 //      If 'a/b/c', an error is thrown
 // Splits out everything after an `=` separator
-function parseSecretPath(inputPath: string): [string, string?, string?] {
+function parseSecretPath(inputPath: string): ParsedSecretPathValue {
   // The colon character `:` is prohibited in vaultName, so it's first occurence
   // means that this is the delimiter between vaultName and secretPath.
   const colonIndex = inputPath.indexOf(':');
@@ -116,7 +117,7 @@ function parseSecretPath(inputPath: string): [string, string?, string?] {
   return [vaultName, secretPath, value];
 }
 
-function parseSecretPathValue(secretPath: string): [string, string?, string?] {
+function parseSecretPathValue(secretPath: string): ParsedSecretPathValue {
   const [vaultName, directoryPath, value] = parseSecretPath(secretPath);
   if (value != null && !secretPathValueRegex.test(value)) {
     throw new commander.InvalidArgumentError(
@@ -126,7 +127,7 @@ function parseSecretPathValue(secretPath: string): [string, string?, string?] {
   return [vaultName, directoryPath, value];
 }
 
-function parseSecretPathEnv(secretPath: string): [string, string?, string?] {
+function parseSecretPathEnv(secretPath: string): ParsedSecretPathValue {
   const [vaultName, directoryPath, value] = parseSecretPath(secretPath);
   if (value != null && !environmentVariableRegex.test(value)) {
     throw new commander.InvalidArgumentError(
@@ -202,30 +203,29 @@ const parseSeedNodes: (data: string) => [SeedNodes, boolean] =
  */
 function parseEnvArgs(
   value: string,
-  prev: [Array<[string, string?, string?]>, Array<string>] | undefined,
-): [Array<[string, string?, string?]>, Array<string>] {
-  const current: [Array<[string, string?, string?]>, Array<string>] = prev ?? [
-    [],
-    [],
-  ];
-  if (current[1].length === 0) {
+  prev: [Array<ParsedSecretPathValue>, Array<string>, boolean] | undefined,
+): [Array<ParsedSecretPathValue>, Array<string>, boolean] {
+  const current: [Array<ParsedSecretPathValue>, Array<string>, boolean] =
+    prev ?? [[], [], false];
+  const [secretsList, commandList, parsingCommandCurrent] = current;
+  let parsingCommand = parsingCommandCurrent;
+  if (!parsingCommand) {
     // Parse a secret path
     if (value !== '--') {
-      current[0].push(parseSecretPathEnv(value));
+      secretsList.push(parseSecretPathEnv(value));
     } else {
-      current[1].push(value);
-      return current;
+      parsingCommand = true;
     }
   } else {
     // Otherwise we just have the cmd args
-    current[1].push(value);
+    commandList.push(value);
   }
-  if (current[0].length === 0 && current[1].length > 0) {
+  if (secretsList.length === 0 && commandList.length > 0) {
     throw new commander.InvalidArgumentError(
       'You must provide at least 1 secret path',
     );
   }
-  return current;
+  return [secretsList, commandList, parsingCommand];
 }
 
 export {
--- a/tests/secrets/env.test.ts
+++ b/tests/secrets/env.test.ts
@@ -1,4 +1,5 @@
 import type { VaultName } from 'polykey/dist/vaults/types';
+import type { ParsedSecretPathValue } from '@/types';
 import path from 'path';
 import fs from 'fs';
 import fc from 'fast-check';
@@ -17,6 +18,8 @@ describe('commandEnv', () => {
   const vaultName = 'vault' as VaultName;
   let dataDir: string;
   let polykeyAgent: PolykeyAgent;
+
+  const secretContentNewlineArb = fc.stringMatching(/^[ -~]+\n$/).noShrink();
 
   beforeEach(async () => {
     dataDir = await fs.promises.mkdtemp(
@@ -59,6 +62,7 @@ describe('commandEnv', () => {
       'unix',
       `${vaultName}:SECRET`,
       '--',
+      '--',
       'node',
       '-e',
       'console.log(JSON.stringify(process.env))',
@@ -85,6 +89,7 @@ describe('commandEnv', () => {
       'unix',
       `${vaultName}:SECRET1`,
       `${vaultName}:SECRET2`,
+      '--',
       '--',
       'node',
       '-e',
@@ -115,6 +120,7 @@ describe('commandEnv', () => {
       'unix',
       `${vaultName}:dir1`,
       '--',
+      '--',
       'node',
       '-e',
       'console.log(JSON.stringify(process.env))',
@@ -141,6 +147,7 @@ describe('commandEnv', () => {
       '--env-format',
       'unix',
       `${vaultName}:SECRET=SECRET_NEW`,
+      '--',
       '--',
       'node',
       '-e',
@@ -169,6 +176,7 @@ describe('commandEnv', () => {
       '--env-format',
       'unix',
       `${vaultName}:dir1=SECRET_NEW`,
+      '--',
       '--',
       'node',
       '-e',
@@ -203,6 +211,7 @@ describe('commandEnv', () => {
       `${vaultName}:SECRET2`,
       `${vaultName}:dir1`,
       '--',
+      '--',
       'node',
       '-e',
       'console.log(JSON.stringify(process.env))',
@@ -230,6 +239,7 @@ describe('commandEnv', () => {
       '--env-format',
       'unix',
       `${vaultName}:SECRET1`,
+      '--',
       '--',
       'node',
       '-e',
@@ -266,6 +276,7 @@ describe('commandEnv', () => {
       `${vaultName}:SECRET2=SECRET1`,
       `${vaultName}:SECRET3=SECRET4`,
       `${vaultName}:dir1`,
+      '--',
       '--',
       'node',
       '-e',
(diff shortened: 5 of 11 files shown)


aryanjassal:
We are using `spawnSync` to spawn the env command. This will result in everything else being blocked, and in cases of a long-running command, will also lead to websocket timeouts. To avoid this, I will change it to use `spawn` much like what I did in `secrets edit`.

Also, there was a minor inaccuracy in the command, where the command would fail without any environment variables being provided. This is different than the Unix `env`, which just runs the command with all the system variables instead. was this intentional or should I fix it?

@tegefaulkes @CMCDragonkai

CMCDragonkai:
In our case `polykey secrets env` doesn't make sense to run with no variables. But... What do you mean by runs with system variables?

aryanjassal:
> In our case `polykey secrets env` doesn't make sense to run with no variables. But... What do you mean by runs with system variables?

By default, when no variables are provided to `env`, it exports all the current shell variables into the command. This can be previewed by running `env` in the terminal, which would output all the current variables. That's what gets passed on to any command. Any other variables we define gets appended to that list.

In our case, it does not make much sense, but is a deviation from the standard Unix behaviour, which is why I wanted to ask and confirm this.

aryanjassal:
> We are using `spawnSync` to spawn the env command. This will result in everything else being blocked, and in cases of a long-running command, will also lead to websocket timeouts. To avoid this, I will change it to use `spawn` much like what I did in `secrets edit`.

After talking about this with Brian, I have realised that the Polykey client is actually stopped before the commands are even run. This means that, without a connection to Polykey, the websocket errors won't happen, and that is a non-issue in this case.

I did a manual sanity check by using a long-running command in `secrets env`, and it didn't throw any errors.

aryanjassal:
The syntax to preserve newlines for a secret is as follows:

```console
[aryanj@matrix-34xx:~]$ polykey secrets env vault:SECRET1 vault:SECRET2
SECRET1='testing123'
SECRET2='testing123'

[aryanj@matrix-34xx:~]$ polykey secrets env -pn vault:SECRET1 vault:SECRET2
SECRET1='testing123
'
SECRET2='testing123'

[aryanj@matrix-34xx:~]$ polykey secrets env --preserve-newline VAULT:SECRET3
SECRET3='newlines-after-this

'

[aryanj@matrix-34xx:~]$ polykey secrets env VAULT:SECRET3
SECRET3='newlines-after-this
'
```

If the newlines should be preserved, then the flag `-pn --preserve-newline` should be used before providing a secret.

aryanjassal:
How should I handle providing the same secret in two different contexts? If `SECRET1` has been provided with both a preserve newline and without, what should take precedence?

What about cases where we export entire vaults? Currently, the flag does preserve newlines of all the secrets exported from the vault, but what if we need one secret to have newlines, and the rest can be trimmed? Or vice-versa, where all secrets need to have preserved newlines, but one secret needs to be trimmed? How would be handle those cases?

With the way how the command is currently structured, is it not possible to preserve the order of the secrets, so they can't be used for prioritisation.

Currently, the preserving newline takes precedence over trimming them. For example, the following will preserve newlines for all secrets:

```console
[aryanj@matrix-34xx:~]$ polykey secrets env -pn vault vault:SECRET1
SECRET1='testing123
'
SECRET2='testing123
'
```

Should this be the intentional behaviour? @CMCDragonkai

src/secrets/CommandEnv.ts
```diff
@@ -33,7 +35,23 @@ class CommandEnv extends CommandPolykey {
...
+        options.preserveNewline = options.preserveNewline ?? [];
+
+        // There needs to be at least one argument or preserved argument
+        // provided. Otherwise, print the help text and error out.
+        if (args[0].length === 0 && options.preserveNewline.length === 0) {
+          this.outputHelp()
+          throw new commander.InvalidArgumentError(
+            'You must provide at least one secret path.',
+          );
+        }
```
aryanjassal: I don't like doing it this way, but I don't know any other way to make this work. I have set the newline preservation as an option. This means that we have zero secrets provided only when both `args` and `options.preserveNewline` are zero.

If there were no args, then the array will not get properly initialised. The same is the case for `options.preserveNewline`, so I had to manually create an empty array for both.

Then, `this.help()` will print the help page and exit with an error code of zero, so that can't be used for printing help text for an error, and the help text isn't printed for errors thrown from the action. It is only printed when it is thrown in the parsing step. So, I have to manually output the help page, and exit with the error so the correct error code gets set.

This approach just doesn't sit well with me, and there is got to be a better method to do this.

src/utils/parsers.ts
```diff
@@ -208,6 +208,8 @@ function parseEnvArgs(
     [],
     [],
   ];
+  console.error('val', value)
+  console.error('cur', current)
   if (current[1].length === 0) {
     // Parse a secret path
     if (value !== '--') {
```
aryanjassal, line 214: I have tried to think up of many solutions, but I can't find a way to consistently use `'--'` as a delimiter between the secrets and a command using commander. The simplest way to fix this would be to use commander to process and validate the arguments, and use `process.env` to get the raw arguments, which will include the `'--'`. We can parse the raw arguments and use them reliably, but I don't like the fact that we are basically bypassing the literal purpose of commander just to get something like this working.

This is kind of broken right now due to the option for removing newlines. If there are zero arguments, but the command doesn't fail, that means that there were some paths provided with the option to preserve the newlines. However, in this case, the `this.passThroughOptions()` doesn't pass through `'--'`, as it only does that after we have started listing the arguments, which didn't happen in this case, as we are listing individual paths inside the preserve newline option, so commander gobbles up the `'--'`. This can be seen in this command:

```console
[aryanj@matrix-34xx:~]$ polykey secrets env --preserve-newline vault:SECRET -- command
# args = [['command', undefined, undefined], []]
# the -- gets parsed by commander instead of the parser
# now the program thinks that 'command' is a vault name
```

We need to address this problem in this PR, but I'm not sure how I can even approach this. I need inputs.

aryanjassal (commented):
I've run into a roadblock and need some inputs from @CMCDragonkai and @tegefaulkes to help me with some issues.

src/utils/parsers.ts
aryanjassal: Related issue: https://github.com/MatrixAI/Polykey-CLI/issues/325

CMCDragonkai:
You can have a global flag or the flags are applied per spec. I would do something like that.

CMCDragonkai:
By default secret env is going to need to trim a single trailing newline.

Repeating the spec is an interesting situation I would argue that you maintain order in line with whatever the order of the specification is. In the case of env shells the last one takes priority because of how shell processes them.

aryanjassal:
After a discussion with Brian, we have come to the following conclusion:

Instead of using a separate option like `--preserve-newline`, then merging the secrets passed in to that and the original arguments, we instead pass in all the secret paths as arguments. The flag can be used as a modifier to the argument list instead of acting as a separate argument list to keep track of.

This means that all secrets specified are, by default, trimmed. However, if a secret name is present in the list to preserve newlines, then the newlines will be preserved. This approach does cause duplication to the secret path, but that's not a huge concern, as most of the times, the newlines would be trimmed as usual.

This approach solves all the issues. The newline preservation takes precedence. If the newlines must be trimmed, then don't specify it to be preserved. As the arguments need to be populated anyways, they will never be empty, and the usual checks can be done from within the parser without need of manually creating an empty object or throwing an error from the action.

This also, to an extent, treats the symptoms of `'--'` being weird to use. That is an exceptionally weird issue, as they work perfectly in the tests, but break when used via the command line. I can start by using `shell` in `pkExec`, which is being used in the tests. This will let me find out if the shell is misbehaving or not.

src/utils/parsers.ts
aryanjassal: Answered in https://github.com/MatrixAI/Polykey-CLI/pull/329#issuecomment-2472332972

src/secrets/CommandEnv.ts
aryanjassal: Answered in https://github.com/MatrixAI/Polykey-CLI/pull/329#issuecomment-2472332972

CMCDragonkai:
Can you also review https://github.com/MatrixAI/Polykey/issues/222#issuecomment-2452549365? I want to make sure we're laying foundations to enable egress schema implementation.

aryanjassal:
> Can you also review [MatrixAI/Polykey#222 (comment)](https://github.com/MatrixAI/Polykey/issues/222#issuecomment-2452549365)? I want to make sure we're laying foundations to enable egress schema implementation.

Just to confirm, this egress schema refers to a schema which would allow formatting of the secrets in a particular way at the time of egress, which would be using the `secrets env` command in this case? For example, the schema would specify if a secret needs to preserve trailing newline or not. If so, then yes, foundation for supporting this is being laid in this PR.

src/utils/parsers.ts
aryanjassal, line 214: For now, we are also returning the `'--'` in the returned command arguments as such, but it doesn't seem ideal.

```ts
> // Command setup
> // Calling with `polykey secrets env vault:SECRET -- command`
> console.log(args);
[['vault', 'SECRET', undefined'], ['--', 'command']]
```

Currently, this is being removed at the beginning of the `env` command's action, but this shouldn't be the case in the first place. How should we deal with it, if at all?

One idea I had was to store some state inside `prev` as a JSON object, so we can extract the args. The parser can set and get the previously-set state to switch contexts between parsing vaults to parsing commands.

Another one is to store this state out of the function, so the `prev` doesn't get messed up, but storing the state of a function outside the actual function just feels weird.

Should this be addressed, or the command removing `'--'` is fine and the behaviour doesn't need modification?

tests/secrets/env.test.ts
```diff
@@ -694,7 +721,40 @@ describe('commandEnv', () => {
...
+      '--env-format',
+      'unix',
+      '--preserve-newline',
+      `${vaultName}:${secretName}`,
+      `${vaultName}:${secretName}`,
+      '--',
+      'node',
+      '-e',
+      'console.log(JSON.stringify(process.env))',
+    ];
```
aryanjassal: Very interestingly, if the order of the flag is changed to the following, the flag gets parsed as a vault name instead. I tried to replicate this behaviour on the command line manually, but that didn't work, and it worked as expected.

```ts
const command = [
  ...
  `${vaultName}:${secretName}`,
  '--preserve-newline',
  `${vaultName}:${secretName}`,
  ...
];
```

First, for the tests, specifying a single `'--'` works but shell needs two of them, now this inconsistency. Something weird is going on. I will investigate this when I work next.

aryanjassal: So, apparently, this issue is resolved by removing the `this.passThroughOptions()`, which makes the tests work as on the CLI, but that has the side-effect of the tests also needing the double `'--'` to be parsed properly. I think we might need to rely on `process.argv` to process the secret paths and the commands separately.

src/utils/parsers.ts
aryanjassal: We can't rely on any checks like `if (current[1].length === 1) { current[1].shift(); }`, as once it has been removed, it will trigger the condition again, and if the user intends to pass `'--'` as an actual value, then this approach will also incorrectly remove that, too. As the array is of strings, I cannot use values like `null` or `undefined` as placeholder values to be removed later, and I also can't use an empty string as the placeholder because `commander` can return any combination of a string.

tests/secrets/env.test.ts
tegefaulkes: Were the `--preserve-newline` options being set before the arguments of the command?  The comment above seems to suggest that you were setting the options after the arguments. Which `passThroughOptions` will treat as arguments because it's telling commander to treat everything after the first argument as more aguments.

aryanjassal: Interestingly enough, doing it like that actually works on the CLI, and the CLI needs two `'--'`, as if `passThroughOptions()` isn't having an effect on the actual CLI of the command, but just the tests.

If `passThroughOptions()` doesn't even work on the CLI, and only on the tests (as what we can see here), should I just remove it and look for alternative solutions for the `'--'` issue, like trying to use `process.argv` for arguments and parsing instead of `commander`?

I enabled the shell mode in `pkExec` in the tests, and the behaviour was consistent with the actual shell, in that I needed two of `'--'` and the options were working in any order.

aryanjassal: After digging into the reason for needing double `--` and the inconsistency between the tests and the CLI, I discovered that I needed an option, `this.enablePositionalOptions()` to be set. Without it, I cannot pass through the options and they are parsed as normal.

However, enabling the option breaks other things, especially the option availability on all subcommand layers. That doesn't work, and my tries of using ChatGPT to fix it have succeeded a little, but ultimately failed to fix all the issues.

I will make an issue on `commanderjs` repo, asking them to add a feature to allow this, or provide a workaround.

aryanjassal: Until then, to run commands, just use double `--` instead of a single one.

```console
[aryanj@matrix-34xx:~]$ polykey secrets env vault:SECRET -- -- bash -c 'echo $SECRET'
this is the secret contents
```

src/utils/parsers.ts
aryanjassal: To address this, we are storing a state flag in the parser `prev`. If we encounter the `--`, then we set the flag and continue. This will then result in the following iterations appending to the command list instead. We can simply ignore this state when deconstructing the array anywhere else.

aryanjassal:
Key changes in this PR:
- Added a new error which is thrown whenever a subprocess fails. It is used in `secrets env` (if the command fails) and `secrets edit` (if the editor fails).
- Properly removed handlers from editor process in `secrets edit`
- Automatically trimming newlines when exporting secrets using `secrets env`
- Preserving newlines of any secrets listed in `-pn` option. (should there be a short form of this option?)
- Created type for parsed secret path and value (`[string, string?, string?]` like `[vaultName, secretPath, value]`)
- Added tests for trimming and preserving newlines using fastcheck

Need input on:
- Is the addition and usage of `ErrorPolykeyCLISubprocessFailure` reasonable?
- Should `-pn --preserve-newline` have a short form of the option?

@tegefaulkes

CMCDragonkai:
> Key changes in this PR:
> 
> * Added a new error which is thrown whenever a subprocess fails. It is used in `secrets env` (if the command fails) and `secrets edit` (if the editor fails).
> * Properly removed handlers from editor process in `secrets edit`
> * Automatically trimming newlines when exporting secrets using `secrets env`
> * Preserving newlines of any secrets listed in `-pn` option. (should there be a short form of this option?)
> * Created type for parsed secret path and value (`[string, string?, string?]` like `[vaultName, secretPath, value]`)
> * Added tests for trimming and preserving newlines using fastcheck
> 
> Need input on:
> 
> * Is the addition and usage of `ErrorPolykeyCLISubprocessFailure` reasonable?
> * Should `-pn --preserve-newline` have a short form of the option?
> 
> @tegefaulkes

If we are using exec, subprocess failure should be temporary on Unix systems, because of process replacement. It would be more accurate to say that you were not able to execute the subprocess. Also I prefer the name ChildProcess.

On windows however there is no process replacement. It's always a child process. Did you do cross platform sanity checks?

aryanjassal:
> Did you do cross platform sanity checks?

Oh, it completely slipped my mind. I will do this today before merging.

src/secrets/CommandEnv.ts
```diff
@@ -160,7 +157,27 @@ class CommandEnv extends CommandPolykey {
...
+                for (const pair of options.preserveNewline) {
+                  if (
+                    pair[0] === nameOrId &&
+                    (pair[1] === newName || pair[1] == null)
+                  ) {
+                    preserveNewline = true;
+                    break;
+                  }
+                }
+              }
```
tegefaulkes: This is poorly done. 

First we only ever do `if(thing)` if thing is a boolean. Otherwise you need to be stricter about it. In this case we're checking if `options.preserveNewLine` is defined. We always check this with `if(options.preserveNewLine != null)`. 

Secondly, if you're checking if something is within a set, you need to use a map rather than iterate over the whole set. I think you're checking if a path or a whole vault for that path is preserving the newline here? That should have a comment explaining that as well. But for the most part rather than iterate over the whole set of `options.preserveNewLine` You should structure it as a map and use that instead. To check if something is in a map is log complexity vs the linear complexity of iterating over a list.

src/secrets/CommandEnv.ts
```diff
@@ -175,14 +192,26 @@ class CommandEnv extends CommandPolykey {
           // Here we want to switch between the different usages
           const platform = os.platform();
           if (cmd != null) {
-            // If a cmd is| provided then we default to exec it
+            // If a cmd is provided then we default to exec it
             switch (platform) {
               case 'linux':
-              // Fallthrough
```
tegefaulkes, line 181: Why did you remove the `// Fallthrough` comment? This is like a `this page is intentionally left blank`. With switch case statements a common problem is forgetting the `break:` between cases, this is letting people know it's intentional and not a mistake.

tests/utils/utils.ts
```diff
@@ -112,6 +112,13 @@ const cmdArgsArrayArb = fc
   })
   .noShrink();
 
+const vaultNameArb = fc
+  .stringOf(
+    fc.char().filter((c) => binParsers.vaultNameRegex.test(c)),
+    { minLength: 1, maxLength: 100 },
+  )
+  .noShrink();
```
tegefaulkes: You should generate the string directly from the regex here.

tegefaulkes (changes requested):

src/secrets/CommandEnv.ts
aryanjassal: Oh. I just did it for clarity, as in other places, the fallthrough comment has been omitted. Especially in error composition in Polykey, Polykey-CLI, and js-rpc. I was just being consistent with that. I can change this back if needed, but how do I handle the other switch statements with that?

aryanjassal:
> On windows however there is no process replacement. It's always a child process. Did you do cross platform sanity checks?

https://github.com/MatrixAI/Polykey-CLI/issues/331

src/secrets/CommandEnv.ts
tegefaulkes: Add it back in. If you see it missing in other places then add it in there. I thought the linter complains about this but it might just be my IDE.

aryanjassal: Yeah, I think your IDE does some extra processing over the regular standards we have. Can you take a look at `Polykey:src/network/utils.ts:487` and see if your IDE still complains for that? If so, then there are a couple places where we would need to add the fallthrough comment.

This has also been done in `Polykey-CLI:src/utils/utils.ts:194`, and adding comments would make it more difficult to read.

```ts
switch (lastChar) {
  case 'n':
    return '\n';
  case 'r':
    return '\r';
  case 't':
    return '\t';
  case 'v':
    return '\v';
  case 'f':
    return '\f';
  case '"':
  case "'":
  case '`':
  case '\\':
    return lastChar;
}
// versus
switch (lastChar) {
  case 'n':
    return '\n';
  case 'r':
    return '\r';
  case 't':
    return '\t';
  case 'v':
    return '\v';
  case 'f':
    return '\f';
  case '"':
  // Fallthrough
  case "'":
  // Fallthrough
  case '`':
  // Fallthrough
  case '\\':
    return lastChar;
}

```

src/secrets/CommandEnv.ts
aryanjassal: > Secondly, if you're checking if something is within a set, you need to use a map rather than iterate over the whole set. I think you're checking if a path or a whole vault for that path is preserving the newline here? That should have a comment explaining that as well. But for the most part rather than iterate over the whole set of `options.preserveNewLine` You should structure it as a map and use that instead. To check if something is in a map is log complexity vs the linear complexity of iterating over a list.

I'm not sure how to achieve this. `options.preserveNewline` is an array of parsed environment paths like the ones parsed in the actual arguments. (like `[ ['vaultName', 'secretPath', 'value'] ]`). Perhaps the naming being 'pair' is kind of misleading. How would I structure this in a map?

Do you mean doing something like putting vault name as the key, and all the provided paths as the values, then use that for the checks here?

src/secrets/CommandEnv.ts
tegefaulkes: Seems my IDE doesn't complain about that. All the same messing up a fallthrough in a switch case is a common bug. and annoying to debug because intention can be unclear. Its best to be explicit about the fallthough.

If you're worried about readability then just put the comment on the same line
```ts
switch (thing) {
// ...
    case '"': // Fallthrough
    case "'": // Fallthrough
    case '`': // Fallthrough
// ...
}
```

aryanjassal: Okay, yeah, this looks clean and is also informative. Ill make this change across all relevant cases in this repo.

src/secrets/CommandEnv.ts
tegefaulkes:
```ts
const set = new Set<string>();

// The options.preserveNewline fills the set.
for (const [vaultName, path] of options.preserveNewLine) {
  set.add(`${vaultName}${path != null ? `:${path}` : ``}`)
}

// Then when checking you can check the path or the vault name.
function checkPreserveNewline(vaultName: string, path: string): boolean {
  const preserveVault = set.has(vaultName);
  const preservePath = set.has(`${vaultName}:${path}`);
  return preserveVault || preservePath;
}

for (const [vaultName, path] of secrets) {
  if (checkPreserveNewline(vaultName, path)) {
    // Do the thing.
  }
}
```

Can you guess why this would be preferred?

tests/utils/utils.ts
aryanjassal: If I try to do this with `stringMatching`, then the tests fail to run with the following error:

```
Assertions of kind Lookahead not implemented yet!

      113 |   .noShrink();
      114 |
    > 115 | const vaultNameArb = fc.stringMatching(binParsers.vaultNameRegex).noShrink();
```

As such, I will retain the previous approach of checking individual characters and concatenating them.

aryanjassal: The previous approach looked very bad and also worked kinda weirdly. So, I have updated the code to utilise this instead of the previous approach.

```ts
const vaultNameArb = fc
  .string({ minLength: 1, maxLength: 100 })
  .filter((str) => binParsers.vaultNameRegex.test(str))
  .noShrink();
```

tegefaulkes (approved):

aryanjassal:
The CI is passing, and the PR has been approved. I will be merging it now.

src/errors.ts
```diff
@@ -84,6 +84,11 @@ class ErrorPolykeyCLIUnexpectedError<T> extends ErrorPolykeyCLI<T> {
   exitCode = sysexits.SOFTWARE;
 }
 
+class ErrorPolykeyCLISubprocessFailure<T> extends ErrorPolykeyCLI<T> {
+  static description = 'A subprocess failed to exit gracefully';
+  exitCode = sysexits.UNKNOWN;
+}
+
```
CMCDragonkai, line 91: Can you change the description and name according to https://github.com/MatrixAI/Polykey-CLI/pull/329#issuecomment-2481400933

src/identities/CommandDiscover.ts
```diff
@@ -103,8 +103,8 @@ class CommandDiscover extends CommandPolykey {
                   case 'queued':
                     queuedSet.add(vertex);
                     break;
-                  case 'processed':
-                  case 'cancelled':
+                  case 'processed': // Fallthrough
+                  case 'cancelled': // Fallthrough
```
CMCDragonkai, line 107: Why need these comments?

aryanjassal: They are not needed, but are recommended to be inserted. This will ensure that the case is intentionally left blank as a fallthrough and is not a mistake, where it was intended to fill in the content later.

aryanjassal: The basic linter (eslint) doesn't care, but IDEs (like WebStorm) actually report this as an error if there is a blank case without a fallthrough comment.

src/errors.ts
aryanjassal: I had actually merged this PR and had forgotten to make this change. So, I made this change in another PR I was working on. This error has been renamed to `ErrorPolykeyCLIChildProcessFailure` with the updated description in https://github.com/MatrixAI/Polykey-CLI/pull/336.

CMCDragonkai: Is this not specific to the env command? If it's called `ErrorPolykeyCLIChildProcessFailure` that would imply in general the idea of a child process failure to PK CLI.

Unless you're also using this in any situation you're going to create a child process?

aryanjassal: This error is not specific to the `env` command. As `secrets edit` also spawns the editor in a child process, any failure of the child process is wrapped in this error.

So, yes, in any situation where a child process might fail, this error should be used to wrap it, unless there is a better error in that context.

CMCDragonkai:
This was a significant change to the UX of secrets env. Now it is actually usable.

PR #336, opened 2024-11-21
Use a single `--` instead of needing double `--` to separate commands

### LINKED ISSUE #325: `secrets env` needs `--` to be passed twice to parse the command

aryanjassal:
### Description

Previously, we needed to use two `--` to separate commands from arguments. Now, that is no longer needed, and we can use a single `--` to parse arguments.

`polykey secrets env vault:SECRET -- cmd arg1 arg2`

Note that you still can't do this, as it would be parsed by the shell instead of the actual command.

```
$ polykey secrets env vault:SECRET -- cmd arg1 arg2 && cmd2
'cmd' run by polykey
'cmd2' run by shell
```

### Issues Fixed

* Fixes #325 (FIX ENG-466)

### Tasks

- [x] 1. Move parsing to action and emulate what it does
- [x] 2. Remove old parser
- [x] 3. Update tests

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/src/secrets/CommandEnv.ts
+++ b/src/secrets/CommandEnv.ts
@@ -2,6 +2,7 @@ import type PolykeyClient from 'polykey/dist/PolykeyClient';
 import type { ParsedSecretPathValue } from '../types';
 import path from 'path';
 import os from 'os';
+import commander from 'commander';
 import * as utils from 'polykey/dist/utils';
 import CommandPolykey from '../CommandPolykey';
 import * as binProcessors from '../utils/processors';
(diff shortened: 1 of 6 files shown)


aryanjassal:
I ran a couple of manual sanity checks to check this behaviour.

```console
$ polykey secrets env testvault:SECRET1 -- bash -c 'echo $SECRET1'
testing

$ polykey secrets env testvault --preserve-newline testvault -- bash -c echo "$SECRET1"
testing


$ polykey secrets env testvault -- node -e 'console.log(JSON.stringify(process.env["SECRET1"]))'
"testing"

$ polykey secrets env testvault -- -- echo 'hi'
ErrorPolykeyCLIChildProcessFailure: A child process failed to exit gracefully - Command failed with error Error: 2
  data: {"command":["--","echo","hi"]}
  cause: Error: 2

$ polykey secrets env -- echo 'hi'
You must provide at least 1 secret path
Usage: polykey secrets env [options] <args...>

Run a command with the given secrets and env variables. If no command is specified then the variables are printed to stdout in the format specified by env-format.

Arguments:
  args                                command and arguments formatted as <envPaths...> [-- cmd [cmdArgs...]]

Options:
  -np, --node-path <path>             Path to Node State (default: "/home/aryanj/.local/share/polykey", env: PK_NODE_PATH)
  -pf, --password-file <path>         Path to Password
  -f, --format <format>               Output Format (choices: "human", "json", default: "human")
  -v, --verbose                       Log Verbose Messages (default: 0)
  -ni, --node-id <id>                  (env: PK_NODE_ID)
  -ch, --client-host <host>           Client Host Address (env: PK_CLIENT_HOST)
  -cp, --client-port <port>           Client Port (env: PK_CLIENT_PORT)
  -ef --env-format <envFormat>        Select how the env variables are formatted on stdout if no command is specified (choices: "auto", "json", "unix", "cmd", "powershell", default: "auto")
  -ei --env-invalid <envInvalid>      How invalid env variable names are handled when retrieving secrets. `error` will throw, `warn` will log a warning and drop and `ignore` will silently drop. (choices: "error", "warn", "ignore", default: "error")
  -ed --env-duplicate <envDuplicate>  How duplicate env variable names are handled. `keep` will keep the exising secret, `overwrite` will overwrite existing with the new secret, `warn` will log a warning and overwrite and `error` will throw. (choices: "keep", "overwrite", "warn", "error", default: "overwrite")
  --preserve-newline <path>           Preserve the last trailing newline for the secret content (default: [])
  -h, --help                          display help for command
```

So, I think this is ready for merge now.

src/secrets/CommandEnv.ts
```diff
@@ -26,297 +27,324 @@ class CommandEnv extends CommandPolykey {
...
+        cmd = args[cmdIndex];
+        argv = args.slice(cmdIndex + 1);
+      }
+      if (envVariables.length === 0) {
+        this.addHelpText('before', 'You must provide at least 1 secret path');
+        this.outputHelp();
+        throw new commander.InvalidArgumentError(
+          'You must provide at least 1 secret path',
         );
+      }
```
aryanjassal, line 91: These are the changes made my this PR. Are these changes okay, or do they need some touch-ups?

tegefaulkes: It looks fine. If there's anything wrong with it I wont spot it from the code alone. We're entering weird territory when it comes to processing the args. So you need to do a bunch of manual testing and check all the edge cases you can think of to make sure it works as expected.

aryanjassal: All the manual testing I did results in all the tests to pass and also the manual sanity checks I ran to pass. See https://github.com/MatrixAI/Polykey-CLI/pull/336#issuecomment-2489954819.

I also used print statements to print the values of `envVariables`, `cmd`, and `argv`. Nothing stood out from my testing, so I think we can go ahead with the merge.

aryanjassal:
Everything seems to be done for this PR. The CI is also passing. Merging.

CMCDragonkai:
This will require some UX testing. @brynblack can you start integrating this.

CMCDragonkai:
> $ polykey secrets env vault:SECRET -- cmd arg1 arg2 && cmd2

This is generally correct, since the subprogram isn't a shell. People who want to do this usually have to make `cmd` the shell and the quote the script they want to run, but that's less common.

PR #339, opened 2024-11-28
ci: merge staging to master

maxwell-aiden:
This is an automatic PR generated by the CI/CD pipeline. This will be automatically fast-forward merged if successful.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -106,7 +106,7 @@
             shellHook = ''
               echo "Entering $(npm pkg get name)"
               set -o allexport
-              . ./.env
+              . <(pk secrets env Polykey-CLI)
               set +o allexport
               set -v
               ${lib.optionalString ci ''
--- a/tests/utils/utils.ts
+++ b/tests/utils/utils.ts
@@ -115,6 +115,7 @@ const cmdArgsArrayArb = fc
 const vaultNameArb = fc
   .string({ minLength: 1, maxLength: 100 })
   .filter((str) => binParsers.vaultNameRegex.test(str))
+  .filter((str) => !str.startsWith('-'))
   .noShrink();
 
 export {


maxwell-aiden:
Pipeline Attempt on 12061525762 for 2b4c00c0eafcae462ac8a6609cc71227e674168f

https://github.com/MatrixAI/Polykey-CLI/actions/runs/12061525762

tegefaulkes:
@aryanjassal its failing on tests you added recently. You need to dig into this.

aryanjassal:
I pulled the repo and tried to run the tests, and they all passed. It could be due to fastcheck discovering an edge case.

aryanjassal:
This is happening due to fastcheck generating `-` as the vault name, and commander trying to parse any values starting with `-` as an option. I can update the vault regex to disallow `-`, or use quotation marks around the vault name.

The first option is the easiest, and the second option will take some time as this change needs to be implemented across all tests, but would be more robust in the future. Which one should I implement, @tegefaulkes?

maxwell-aiden:
Pipeline Attempt on 12078305328 for c6f7614cb27a9c9b14d55dd5bf4dd645e1155f52

https://github.com/MatrixAI/Polykey-CLI/actions/runs/12078305328

maxwell-aiden:
Pipeline Succeeded on 12078305328 for c6f7614cb27a9c9b14d55dd5bf4dd645e1155f52

https://github.com/MatrixAI/Polykey-CLI/actions/runs/12078305328

CMCDragonkai:
> This is happening due to fastcheck generating `-` as the vault name, and commander trying to parse any values starting with `-` as an option. I can update the vault regex to disallow `-`, or use quotation marks around the vault name.
> 
> The first option is the easiest, and the second option will take some time as this change needs to be implemented across all tests, but would be more robust in the future. Which one should I implement, @tegefaulkes?

I would say that yes, if quoted, it's a legal name, if not quoted, it's not a legal name. So in some cases, the fast check generation should be more precise, or you should be expecting a parse error in those scenarios.

PR #348, opened 2024-12-18
feat: added export option to secrets env #342

### LINKED ISSUE #342: Add export formats to `secrets env` command
### Specification

Currently the `secrets env` command contains different formats that get generated when it's not running a command. We need to add variants to these formats that include export within the file.

For example the `unix` format is structured as 

```
ENV_ONE="one";
ENV_TWO="two";
```

But the new variant will include the export keyword

```
export ENV_ONE="one";
export ENV_TWO="two";
```

This should apply to all the existing formats where applicable. So if the cmd and powershell if they have a equavalent way of exporting env variables within a script then we should have a format version for them as well.

### Tasks

1. Add an export variant format for the `unix` format
2. Add an export variant for all other formats that can support it.

Sooryasanand:
### Description
I have introduced an export flag in options.ts to determine whether the export keyword should be prefixed to environment variables for shell usage. This is relevant for Unix-based systems, where export is required to make the variables available to child processes. On Windows, however, the export keyword or an equivalent is unnecessary because a child process automatically inherits the environment variables of its parent process. This behavior is documented on the official Microsoft website: [Environment Variables](https://learn.microsoft.com/en-us/windows/win32/procthread/environment-variables).

### Issues Fixed

* Fixes #342

### Tasks

- [x] 1. Add export option for json
- [x] 2. Add Tests

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/src/secrets/CommandEnv.ts
+++ b/src/secrets/CommandEnv.ts
@@ -24,6 +24,7 @@ class CommandEnv extends CommandPolykey {
     this.addOption(binOptions.envFormat);
     this.addOption(binOptions.envInvalid);
     this.addOption(binOptions.envDuplicate);
+    this.addOption(binOptions.envExport);
     this.addOption(binOptions.preserveNewline);
     this.argument(
       '<args...>',
@@ -280,22 +281,24 @@ class CommandEnv extends CommandPolykey {
           switch (format) {
             case 'unix':
               {
-                // Formatting as a .env file
                 let data = '';
                 for (const [key, value] of Object.entries(envp)) {
-                  data += `${key}='${value}'\n`;
+                  if (options.envExport) {
+                    data += `export ${key}='${value}'\n`;
+                  } else {
+                    data += `${key}='${value}'\n`;
+                  }
                 }
                 process.stdout.write(
                   binUtils.outputFormatter({
                     type: 'raw',
-                    data,
+                    data: data,
                   }),
                 );
               }
               break;
             case 'cmd':
               {
-                // Formatting as a .bat file for windows cmd
                 let data = '';
                 for (const [key, value] of Object.entries(envp)) {
                   data += `set "${key}=${value}"\n`;
@@ -303,22 +306,25 @@ class CommandEnv extends CommandPolykey {
                 process.stdout.write(
                   binUtils.outputFormatter({
                     type: 'raw',
-                    data,
+                    data: data,
                   }),
                 );
               }
               break;
             case 'powershell':
               {
-                // Formatting as a .bat file for windows cmd
                 let data = '';
                 for (const [key, value] of Object.entries(envp)) {
-                  data += `\$env:${key} = '${value}'\n`;
+                  if (options.envExport) {
+                    data += `$env:${key} = '${value}'\n`;
+                  } else {
+                    data += `$${key} = '${value}'\n`;
+                  }
                 }
                 process.stdout.write(
                   binUtils.outputFormatter({
                     type: 'raw',
-                    data,
+                    data: data,
                   }),
                 );
               }
--- a/src/utils/options.ts
+++ b/src/utils/options.ts
@@ -246,6 +246,11 @@ const envDuplicate = new commander.Option(
   .choices(['keep', 'overwrite', 'warn', 'error'])
   .default('overwrite');
 
+const envExport = new commander.Option(
+  '-ee --env-export',
+  'If enabled, exports the set environment variables to child environments',
+).default(false);
+
 const discoveryMonitor = new commander.Option(
   '--monitor',
   'Enabling monitoring will cause discover to output discovery events as they happen and will exit once all children are processed',
@@ -363,6 +368,7 @@ export {
   envFormat,
   envInvalid,
   envDuplicate,
+  envExport,
   discoveryMonitor,
   seekStart,
   seekEnd,
--- a/tests/secrets/env.test.ts
+++ b/tests/secrets/env.test.ts
@@ -416,10 +416,10 @@ describe('commandEnv', () => {
       env: { PK_PASSWORD: password },
     });
     expect(result.exitCode).toBe(0);
-    expect(result.stdout).toContain(`$env:SECRET1 = 'this is the secret1'`);
-    expect(result.stdout).toContain(`$env:SECRET2 = 'this is the secret2'`);
-    expect(result.stdout).toContain(`$env:SECRET3 = 'this is the secret3'`);
-    expect(result.stdout).toContain(`$env:SECRET4 = 'this is the secret4'`);
+    expect(result.stdout).toContain(`$SECRET1 = 'this is the secret1'`);
+    expect(result.stdout).toContain(`$SECRET2 = 'this is the secret2'`);
+    expect(result.stdout).toContain(`$SECRET3 = 'this is the secret3'`);
+    expect(result.stdout).toContain(`$SECRET4 = 'this is the secret4'`);
   });
   test('should output json format', async () => {
     const vaultId = await polykeyAgent.vaultManager.createVault(vaultName);
@@ -924,4 +924,33 @@ describe('commandEnv', () => {
     expect(result.stdout).toContain("SECRET3='this is the secret3'");
     expect(result.stdout).toContain("SECRET4='this is the secret4'");
   });
+  test.each(['unix', 'cmd', 'powershell'])(
+    'should export all secrets in %s format to child processes',
+    async (format) => {
+      const vaultId = await polykeyAgent.vaultManager.createVault(vaultName);
+      await polykeyAgent.vaultManager.withVaults([vaultId], async (vault) => {
+        await vaultOps.addSecret(vault, 'SECRET1', 'this is the secret1');
+      });
+      const command = [
+        'secrets',
+        'env',
+        '-np',
+        dataDir,
+        '--env-format',
+        format,
+        '--env-export',
+        `${vaultName}:SECRET1`,
+      ];
+      const result = await testUtils.pkExec(command, {
+        env: { PK_PASSWORD: password },
+      });
+      expect(result.exitCode).toBe(0);
+      const formatResult = {
+        unix: `export SECRET1='this is the secret1'\n`,
+        cmd: `set "SECRET1=this is the secret1"\n`,
+        powershell: `$env:SECRET1 = 'this is the secret1'\n`,
+      };
+      expect(result.stdout).toEqual(formatResult[format]);
+    },
+  );
 });


src/secrets/CommandEnv.ts
```diff
@@ -283,7 +283,11 @@ class CommandEnv extends CommandPolykey {
                 // Formatting as a .env file
                 let data = '';
                 for (const [key, value] of Object.entries(envp)) {
-                  data += `${key}='${value}'\n`;
+                  if (options.envExport) {
+                    data += `export ${key}='${value}'\n`;
+                  } else {
+                    data += `${key}='${value}'\n`;
+                  }
```
Sooryasanand, line 290: The if statement applies to Unix systems. On Windows, an equivalent check is unnecessary because the export keyword (or equivalent) isn’t needed. On Windows, child processes automatically inherit the parent process's environment variables, as documented on the official Microsoft website: [Environment Variables](https://learn.microsoft.com/en-us/windows/win32/procthread/environment-variables).

aryanjassal: If that is the case, then instead of making this an option, we can probably make this another env format - something like `--env-format unix-export`. Or we can also rename this to be `--env-format unix-local` for non-export and `--env-format unix-global` for exported environment variables. Thoughts, @CMCDragonkai?

CMCDragonkai: PowerShell does have an export like semantic. But I forgot the details.

I think we should automate the default expected behaviour as much as possible and minimise configuration options.

Sooryasanand: There are three ways to manage environment variables in PowerShell.

### Solution 1: Setting Environment Variables at the Machine Level
You can set environment variables at the machine level using the following command:
``[Environment]::SetEnvironmentVariable('${key}', '${value}', 'Machine')``

However, changes made this way require a **restart of the terminal session** or **new processes to take effect**, which might not be ideal for immediate changes.

### Solution 2: Using a Function to Add Paths to $env:Path
You can create a PowerShell function that:

- Updates $env:Path in the current session so changes take effect immediately.
- Persists the change for future sessions.
- Avoids adding duplicate paths if they already exist.

Here’s an example function that implements similar function from [stack overflow](https://stackoverflow.com/a/34844707):
```  powershell
function Add-EnvPath {
    param(
        [Parameter(Mandatory=$true)]
        [string] $Path,

        [ValidateSet('Machine', 'User', 'Session')]
        [string] $Container = 'Session'
    )

    if ($Container -ne 'Session') {
        $containerMapping = @{
            Machine = [EnvironmentVariableTarget]::Machine
            User = [EnvironmentVariableTarget]::User
        }
        $containerType = $containerMapping[$Container]

        $persistedPaths = [Environment]::GetEnvironmentVariable('Path', $containerType) -split ';'
        if ($persistedPaths -notcontains $Path) {
            $persistedPaths = $persistedPaths + $Path | where { $_ }
            [Environment]::SetEnvironmentVariable('Path', $persistedPaths -join ';', $containerType)
        }
    }

    $envPaths = $env:Path -split ';'
    if ($envPaths -notcontains $Path) {
        $envPaths = $envPaths + $Path | where { $_ }
        $env:Path = $envPaths -join ';'
    }
}
```
This solution ensures that the environment variables are updated for both the current session and future sessions, without causing duplication.

### Solution 3: Dynamic Update Based on envExport Flag
If you want to set the environment variables immediately in the current session and optionally persist them system-wide based on the envExport flag, you can use the following approach:

- If envExport is true, the variable is set at the machine level and immediately available in the current session.
- If envExport is false, the variable is only set for the current session.

Here’s the implementation:
``` js
case 'powershell': {
  let data = '';
  for (const [key, value] of Object.entries(envp)) {
    if (options.envExport) {
      // Export environment variable at the machine level (system-wide)
      data += `[Environment]::SetEnvironmentVariable('${key}', '${value}', 'Machine')\n`;
    } else {
      // Set environment variable for current session
      data += `$env:${key} = '${value}'\n`;
    }
  }

  // After setting the variables, fetch and update them in the current session
  for (const key of Object.keys(envp)) {
    data += `$env:${key} = [System.Environment]::GetEnvironmentVariable('${key}', 'Machine')\n`;
  }

  process.stdout.write(
    binUtils.outputFormatter({
      type: 'raw',
      data,
    }),
  );
}
break;
```
This solution allows flexibility: you can choose to export the variable system-wide or just for the current session, with immediate effect in the session.

Summary of Solutions:

- Set machine-wide variables using [Environment]::SetEnvironmentVariable (requires restart of terminal or processes).
- Use a PowerShell function to add paths to $env:Path, making changes immediately available in the current session and persistent for future sessions, while avoiding duplicates.
- Use the envExport flag to control whether the variable is set system-wide (machine level) or just for the current session, updating the current session immediately. 

@CMCDragonkai What do you prefer me to do?

tests/secrets/env.test.ts
```diff
@@ -924,4 +924,40 @@ describe('commandEnv', () => {
...
+      const jsonOut = JSON.parse(result.stdout);
+
+      // Depending on the format, check that the variable is exported in the correct format
+      if (format === 'unix') {
+        expect(result.stdout).toContain("SECRET='this is the secret'");
+      } else if (format === 'cmd') {
+        expect(result.stdout).toContain('set "SECRET=this is the secret"');
+      } else if (format === 'powershell') {
+        expect(result.stdout).toContain(`$env:SECRET = 'this is the secret'`);
+      }
```
tegefaulkes: If you can, avoid if else chains like this. a `switch()` block is better suited and more readable.

Also, for the `unix` format shouldn't it include the `export` keyword if the `--export` option is set?

src/utils/options.ts
```diff
@@ -318,6 +318,11 @@ const parents = new commander.Option(
   'If enabled, create all parent directories as well. If the directories exist, do nothing.',
 ).default(false);
 
+const envExport = new commander.Option(
+  '--export',
+  'If enabled, the export parameter sets the environment variable system-wide.',
```
tegefaulkes: This description doesn't seem correct. Using `Export` just means that the child shells inherit the variable. Where not using `export` means only the current shell has access to it. Export it does not mean setting it system wide.

src/secrets/CommandEnv.ts
tegefaulkes: If the point of the `--export` option is to have the variables inherited by children then I don't think there's anything to change for Powershell, it seems to be done by default. I don't know if there's a way to disable that. The cmd `set` possibly does the same.

CMCDragonkai: In PowerShell, to set a variable that is scoped to the current shell session and not exported to child processes, you simply define it as a normal variable. PowerShell variables are not exported to child processes unless you explicitly mark them as environment variables.

Here’s how you can do it:

1. Set a Normal Variable

$myVariable = "This is a local variable"

This variable is only accessible within the current session and scope. It will not be available to any child processes.

2. Verify It's Not Exported to Child Processes

You can test this by launching a child process (e.g., cmd.exe) and checking if the variable is accessible:

$myVariable = "This is a local variable"
cmd.exe /c echo %myVariable%

The result will be blank, as $myVariable is not exported.

3. Exporting a Variable to Child Processes (if Needed)

To explicitly export a variable to child processes, you need to use the environment variable scope:

$env:myVariable = "This is an environment variable"

This variable will then be available to child processes.

To summarize: by default, variables in PowerShell are not exported to child processes unless you specifically set them as environment variables using $env:. Just define your variable as usual to keep it local.

CMCDragonkai: So just do it the way chatgpt said ^.

tests/secrets/env.test.ts
Sooryasanand: I have changed the if statement to a switch block as it's better and more readable like you told. 

Added the export keyword as it was needed when testing for unix systems

is this good, @tegefaulkes?

src/utils/options.ts
Sooryasanand: I have updated the export description to "Make the environment variable accessible to child processes".  

is this a better, @tegefaulkes?

src/secrets/CommandEnv.ts
Sooryasanand: Thanks for the feedback, Roger. I've updated the PowerShell code to correctly handle variable scoping based on the envExport option:

- If envExport is true, variables are defined using $env:{key}, ensuring they are exported to child processes.
- If envExport is false, variables are defined as normal PowerShell variables (${key}), keeping them scoped to the current shell session.

is this better, @CMCDragonkai

tests/secrets/env.test.ts
tegefaulkes: Yep, looking good.

src/utils/options.ts
tegefaulkes: A child process is something else. You should probably call it `child environments`.

tests/secrets/env.test.ts
```diff
@@ -924,4 +924,46 @@ describe('commandEnv', () => {
...
+        'env',
+        '-np',
+        dataDir,
+        '--env-format',
+        format,
+        '--export',
+        `${vaultName}:SECRET`,
+        '--',
+        'node'
+      ];
```
aryanjassal: At the end of the command, we do a `-- node -e 'console.log(JSON.stringify(process.env))'`. This runs the specified command (separated by a `--`) in the shell with the secrets exported.

In this case, we are testing the `stdout` of the command instead, so this would not work like this. Instead, remove the last two lines (containing `--` and `node`).

tests/secrets/env.test.ts
```diff
@@ -924,4 +924,46 @@ describe('commandEnv', () => {
...
+
+      const result = await testUtils.pkExec(command, { env: { PK_PASSWORD: password } });
+
+      expect(result.exitCode).toBe(0);
+      const jsonOut = JSON.parse(result.stdout);
```
aryanjassal: Here, you attempt to get `jsonOut`, but that is not used anywhere. The current command is ill-suited to even return JSON, so this should be removed.

tests/secrets/env.test.ts
```diff
@@ -924,4 +924,46 @@ describe('commandEnv', () => {
...
+          break;
+        case 'cmd':
+          expect(result.stdout).toContain('set "SECRET=this is the secret"');
+          break;
+        case 'powershell':
+          expect(result.stdout).toContain(`$env:SECRET = 'this is the secret'`);
+          break;
+        default:
+          throw new Error(`Unsupported format: ${format}`);
+      }
```
aryanjassal: It is fine here, but normally I would do a `expect().toEqual()` and make sure the entire string matches the exact value we are after. Note that you would probably need to include the `'\n'` symbol in the expected string.

Instead of throwing a plain error in the default case, look into `utils.never` from Polykey. Perhaps doing a fuzzy search over the code base can give you some examples of its usage?

aryanjassal: https://github.com/MatrixAI/Polykey-CLI/blob/ca15d2436ae06497130117cc0c9236d1b4ea35b3/src/secrets/CommandEnv.ts#L156-L174

This could be an example of doing such an operation.

CMCDragonkai:
If review comments are fixed, resolve the threads with a comment, don't leave it open.

tests/secrets/env.test.ts
aryanjassal: Instead of doing it this way, soorya suggested using a JSON object storing all the expected results like so.

```ts
  test.each(['unix', 'cmd', 'powershell'])(
    'should export all secrets in %s format to child processes',
    async (format) => {
      // ...
      const formatResult = {
        unix: `export SECRET1='this is the secret1'\n`,
        cmd: `set "SECRET1=this is the secret1"\n`,
        powershell: `$env:SECRET1 = "this is the secret1"\n`,
      };
      expect(result.stdout).toEqual(formatResult[format]);
    },
  );
```

This is more compact and elegant than before.

src/secrets/CommandEnv.ts
aryanjassal: I noticed a minor issue in your implementation, Soorya. In powershell, instead of assigning variables by doing `$KEY = 'value'`, you were doing `KEY = 'value'`. This was probably due to the template string needing a `${}` at the beginning, confusing you with the dollar sign.

aryanjassal:
I have gone through the PR and made any required changes. Linting gives the all-clear, and the tests are also all passing. This should now be ready for merging.

CMCDragonkai:
Can someone add a comment to #342 explaining what this PR brings in from a UI/UX perspective.

CMCDragonkai:
Also not all conversations here have been resolved.

src/secrets/CommandEnv.ts
CMCDragonkai: What's the resolution?

aryanjassal: I have implemented the fix, but had forgotten to resolve this thread.

tegefaulkes (approved):

CMCDragonkai:
@aryanjassal #342 or @Sooryasanand please summarise the feature in #342.

aryanjassal:
Everything has finished for this PR. It was just hanging because I had other priorities. I have gone through this PR and can confirm that it is now ready for merging. Will make a new release for this.

PR #357, opened 2025-01-28
Updating types to match Polykey#846

### LINKED ISSUE #832: Review and refactor RPC handlers to handle cancellation properly
### Specification

Some RPC handlers are preforming long running async tasks. In these cases they need to make use of the provided `ctx` and handle cancellation gracefully to prevent any resource leaks. A clear example of this is the vaults pull and clone handlers.

So we need to review handlers and update the ones that need CTX handling applied.

### Additional context

- Related https://github.com/MatrixAI/Polykey-CLI/issues/264 REF ENG-397

### Tasks

1. Review all RPC handlers and identify ones that preform complex async  or long running tasks.
2. Refactor theses handlers to handle cancellation in a responsive and graceful manor.
3. Apply testing to see if timing out during these tasks causes any problems. Possibly apply regression testing for this.

aryanjassal:
### Description

There have been a lot of changes in https://github.com/MatrixAI/Polykey/pull/846 which require changes in Polykey CLI too. This PR makes those changes, ensuring the CLI remains up to date.

### Issues Fixed

* Fixes https://github.com/MatrixAI/Polykey/issues/832

### Tasks

- [x] 1. Update types to the tagged variants
- [x] 2. Remove usage of `VaultsSecretsGet` to prepare for its removal

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/src/secrets/CommandEdit.ts
+++ b/src/secrets/CommandEdit.ts
@@ -27,10 +27,10 @@ class CommandEdit extends CommandPolykey {
       const secretPath = fullSecretPath[1] ?? '/';
       const os = await import('os');
       const { spawn } = await import('child_process');
-      const vaultsErrors = await import('polykey/dist/vaults/errors');
       const { default: PolykeyClient } = await import(
         'polykey/dist/PolykeyClient'
       );
+      const { never } = await import('polykey/dist/utils');
       const clientOptions = await binProcessors.processClientOptions(
         options.nodePath,
         options.nodeId,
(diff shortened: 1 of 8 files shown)


aryanjassal:
Forgot that working on an 'unreleased' polykey version will cause test failures. Oops.

src/secrets/CommandEdit.ts
```diff
@@ -66,41 +65,48 @@ class CommandEdit extends CommandPolykey {
...
             });
-            try {
-              let rawSecretContent: string = '';
-              for await (const chunk of response) {
+            await writer.close();
+            let rawSecretContent: string = '';
+            for await (const chunk of response.readable) {
+              if (chunk.type === 'SuccessMessage') {
                 rawSecretContent += chunk.secretContent;
+              } else {
```
tegefaulkes: We need to open a file descriptor to the temp file and write to that directly in chunks. We want to avoid keeping the whole secret in memory just to write it to a file in one go.

tegefaulkes (changes requested):

src/secrets/CommandEdit.ts
aryanjassal: I have done this. However, I have also stumbled upon an optimisation issue. 

When the secret is being read from disk, it is accumulating the data like in here and sending it over in one call. The relevant RPC, `VaultsSecretsWriteFile`, actually is a unary call. My first idea was to utilise a similar method to `VaultsSecretsRemove` where we use a client streaming and a header-content communication protocol. However, that will be a lot of time and effort investment on something which is working fine for now and is bound to change a little bit down the line.

This also affects another issue, which relates to not updating the secret if it has not changed. However, doing it the way we are already doing it will require on average double the memory, perhaps even more depending on the changes. Instead, I tried to stream the file contents to compare them, but that is too much complexity, and they end up needing to be accumulated anyways.

I have made a note of this, but we need to work on converting these RPCs to raw streams before we can handle something like this.

CMCDragonkai: Didn't we discuss change all those unary calls to streaming calls? Any IO regarding secret files should be streaming not unary.

PR #366, opened 2025-02-03
Added `secrets touch` command

### LINKED ISSUE #244: Implement `secrets touch` command
### Specification

`touch` serves two purposes
1. Will create a new empty file if no file exists at that path.
2. Will update the timestamps of a file if it already exists.

For our `secrets touch` command we need to support these two actions while working with normal file paths and secret paths. It should also support multiple paths for touching multiple files. Wild cards are supported as well but will not create files.

So we have two levels of functionality.
1. Create files.
2. Update file access times without modifying contents.

I don't know if we want to include updating the files access times. Most of the commands complexity will come from that.

### Additional context

Related #32

### Tasks

1. Implement a `secrets touch` command.
4. Must create a new empty secret or file if no file exists at the provided path.
5. If secret or file exists then we update the updated and access time without modifying contents.
6. It takes a varadic argument of file paths, so we can touch multiple files at a time.
7. File paths and secret paths should be interchangeable. 
8. Touching to create a file can't operate recusively. So it won't create a directory for a touched file.
9. Wildcards can be used for updating access times of files, but not for creating files.

aryanjassal:
### Description

Added `secrets touch` command to create a file if it doesn't exist, otherwise update the `mtime` if it does. The `mtime` is updated for directories too.

### Issues Fixed

* Fixes #244 

### Tasks

- [x] 1. Add `secrets touch`
- [x] 2. Add tests for `secrets touch`

### Final checklist

* [ ] Domain specific tests
* [ ] Full tests
* [ ] Lint fixed
* [ ] Squash and rebased

### DIFF
--- a/npmDepsHash
+++ b/npmDepsHash
@@ -1,1 +1,1 @@
-sha256-m1ZpBUrLygWELYlgShs1oWs+o3qf3e6G4rhd5ZzsOhE=
+sha256-yH5mPEv7drHBiLeI+KzMWv4ZQqqqJyiEzDdl6ntMRvQ=
--- a/package-lock.json
+++ b/package-lock.json
@@ -42,7 +42,7 @@
         "nexpect": "^0.6.0",
         "node-gyp-build": "^4.4.0",
         "nodemon": "^3.0.1",
-        "polykey": "^1.18.0",
+        "polykey": "^1.19.0",
         "prettier": "^3.0.0",
         "shelljs": "^0.8.5",
         "shx": "^0.3.4",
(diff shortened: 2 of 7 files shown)


aryanjassal:
Note that merging this PR will introduce the `secrets touch` command, but not remove the `secrets new` command, as that command is responsible for copying the contents of a file on disk to a path on the vault. This behaviour will be covered by the `secrets cp` or `secrets mv` command, which will eventually make its way into PKCLI. Until then, there would be no real 1:1 command replacement.

aryanjassal:
All the tests are passing and the command seems to be working. I verified this with manual sanity checks. So, this will now be merged.

CMCDragonkai:
Hey checklist should be checked.

CMCDragonkai:
> Note that merging this PR will introduce the `secrets touch` command, but not remove the `secrets new` command, as that command is responsible for copying the contents of a file on disk to a path on the vault. This behaviour will be covered by the `secrets cp` or `secrets mv` command, which will eventually make its way into PKCLI. Until then, there would be no real 1:1 command replacement.

Yes eventually new needs to be removed.

PR #381, opened 2025-04-03
ESM migration

### LINKED ISSUE #375: Polykey-CLI: Migration to ESM
### Specification

This issue is a part of a body of work for converting all relevant repository to ESM. This focuses on converting Polykey-CLI to esm

### Additional context

- Project: https://linear.app/matrix-ai/project/migration-to-esm-8d106e215b5c/overview

### Tasks

1. Update all dependencies to ESM
2. Migrate tests
3. Migrate package.json
4. Migrate src code

tegefaulkes:
### Description

This PR addresses converting `Polykey-CLI` to ESM.

### Issues Fixed

* Fixes #375 

### Tasks

- [x] 1. Migrate to ESM.
- [x] 2. Check that Nix build works.
- [x] 3. Check that the docker image runs.

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased


tegefaulkes:
I have a functional bundled code now. Right now I'm running into problems with tests. Specifically mocking. It seems that how mocks of modules and imports are handled are changed with ESM. There is a new major constraint causing problems.

1. The ESM imports are read-only now. So we can't do any cheeky overriding of functionality globally.
2. Its possible to mock modules that are imported. HOWEVER you must mock then before they are imported. Since most imports are static and done before any code execution these can't be mocked. Only dynamically imported code can be mocked now.

So the usual methods can't be done anymore. I need to review and think of alternative ways to handle mocking.

CMCDragonkai:
Amy mentioned that mocking doesn't work the same way as before with ESM, when she was working on js-ws and js-rpc.

Generally speaking mocking isn't the best way to do testing. DI is preferred in most cases. Mocking is a hack, for very static scenarios, you're reinjecting dynamism when it was supposed to be static.

In CJS, it was dynamic masquerading as static, but in ESM, it's static properly now.

Can you give a list of locations where mocking is a problem, you can use R1 to track and list all locations with a description after indexing the codebase, as well as a fuzzy search.

Then in most cases we should be able to convert to DI instead. And other cases, we may just do a proper integration.

tegefaulkes:
The problem here is we have a clear boundary of the cli running as its own process or pretending to in some cases. We can't DI across that boundary.

The options here are 
1. Remove all mocks
2. Special case testing code with flags.
3. Update code that needs to be mocked with dynamic imports so mocks can be applied before running them.

Given the CLI already depends on dynamic importing for optimisation then option 3 seems the least annoying.

tegefaulkes:
Everything is fixed up now. I just need to check the nix build works and the docker image works.

PR #394, opened 2025-04-28
Fix `yao-pkg` bundling for post-ESM Polykey CLI

### LINKED ISSUE #391: Fix ESM pkg packaged output
### Specification

The ESM migration is complete. but rather predicibly `pkg` which we use to bundle everything into a single executable is giving us issues. Right now it doesn't fully support ESM. There are some work-arounds written on in one of their issues here.

https://github.com/yao-pkg/pkg/issues/16#issuecomment-1945486658

Right now I think the best solution is to make our esbuild scrip more comprehensive in bundling everything together in a single directory. We need to make better use of plugins and patches to do a better job of this. There is an example of this here https://github.com/zwave-js/zwave-js-ui/blob/master/esbuild.js

Alternatively we could explore better methods of generating a single executable.

In any case, right now I'm stuck on `sodium-native` either not being bundled properly or not being imported properly. This results in the CI failing https://github.com/MatrixAI/Polykey-CLI/actions/runs/14512736158/job/40715279046 . I'm leaning towards it not importing `sodium-native` properly. Maybe it needs to be imported with `sodium-native/index.js`?

Some more prototyping needs to be done but I'm running out of time before the easter break. This either needs to be put on hold till I get back or someone else needs to take it over.

### Additional context

- yao-pkg issue outlining the ESM problem: https://github.com/yao-pkg/pkg/issues/16#issuecomment-1945486658
- Example of a more comprehensive Esbuild script: https://github.com/zwave-js/zwave-js-ui/blob/master/esbuild.js
- Failing CI: https://github.com/MatrixAI/Polykey-CLI/actions/runs/14512736158/job/40715279046

### Tasks

1. explore a more comprehensive esbuild script.
2. explore potential patches or quick fixes for importing sodium-native properly.

aryanjassal:
### Description

Currently, `yao-pkg` does not work well with ESM packages. Thus, to solve this issue, we first have to compile the application in CJS mode instead, then include the native modules.

First, we need to make sure the native modules are being bundled correctly in the binary, then we need to fix up the paths, as many libraries use `__dirname` to refer to the current dir, which when compiled, points to the incorrect directory.

### Issues Fixed

* Fixes #391 (FIX ENG-584)

### Tasks

- [x] 1. Compile to CJS via esbuild
- [x] 2. Confirm the bundle has native modules
- [x] 3. Fix up the import paths

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased


scripts/pkg.mjs
CMCDragonkai, line 1: Would be good to have commentary about what is happening.

pkg.json
aryanjassal, line 1: This file appears to be an artifact and not strictly required in the code. This can probably be deleted.

scripts/pkg.mjs
tegefaulkes: Not much to say really. The fix comprises three main things.

1. Changing the output to common js so pkg has a better time dealing with it.
2. native modules that loaded with `node-gyp` needed their load paths overridden. This was done with a esbuild plugin.
3. Certain globals used by ether esm or cjs needed to be overridden with shims to prevent them being undefined.

scripts/build.mjs
```diff
@@ -96,7 +173,11 @@ async function main(argv = process.argv) {
   console.error('Renaming worker script');
   childProcess.execFileSync(
     'mv',
-    ['dist/polykeyWorkerManifest.mjs', 'dist/polykeyWorkerManifest.js'],
+    [
+      `dist/polykeyWorkerManifest.${isPkg ? 'cjs' : 'mjs'}`,
+      'dist/polykeyWorkerManifest.js',
+    ],
+    // ['dist/polykeyWorkerManifest.mjs', 'dist/polykeyWorkerManifest.js'],
```
aryanjassal: This comment here is also a remnant of the previous code I commented out so I can easily recover it if my approach didn't work out. This should also probably be removed.

scripts/pkg.mjs
```diff
@@ -143,7 +144,6 @@ async function main(argv = process.argv) {
     `--config=${pkgConfigPath}`,
     `--targets=node${nodeVersion}-${pkgPlatform}-${pkgArch}`,
     '--no-bytecode',
-    '--no-native-build',
```
aryanjassal: I'm not sure if this option interfered with the pkg output or not. We might be able to reintroduce it?

pkg.json
tegefaulkes: Removed.

scripts/build.mjs
tegefaulkes: I missed that one, removed.

scripts/pkg.mjs
tegefaulkes: I've reverted this file.

Issue #408, opened 2025-05-16
Create a `polykey auth login` command


tegefaulkes:
### Specification

This command now falls under the new `polykey auth` subcommand.

As part of the PKE PK integration we need to create a new command to help with logging in to PKE. At its core this command just triggers the node to connect to a target node and hand over a token it was given. Since this handover is done over a TLS secured connection protected by the nodes certificate we both prove identity of the node securely an link the node to the login attempt.

The command will be `polykey auth login <token>`, it will take in the provided token and trigger all the steps to login with that token. The token itself will contain the node it needs to connect to and send the token to.

This breaks down into two parts.

1. The creation of the command on `Polykey-CLI`.
2. The creation of the client RPC handler for `Polykey` to handle the command. This needs to handle signing the JWT token and calling the proper agent RPC handler.

How the overall login flow works as designed is under change. We're still reviewing how the flow works and what the best way to approach it is. @shafiqihtsham Is reviewing the options right and and will make a write up comparing them. From there we will decide on the next steps in [ENG-619](https://linear.app/matrix-ai/issue/ENG-619).

### Additional context

Parent issue: REF [ENG-580](https://linear.app/matrix-ai/issue/ENG-580)

### Tasks

1. Create a client RPC handler that triggers calling the appropriate agent-RPC handler for logging into a PKE node.
2. Create the `polykey-cli` command `polykey auth login <token>`.

CMCDragonkai:
Are you sure this command makes sense, according to the plan here.

![Image](https://github.com/user-attachments/assets/2a6f50aa-e0c8-4986-a941-26a2fb54e6de)

CMCDragonkai:
Like I don't understand what `polykey login` is for? Logging into what? It should just be part of the `agent start` command and integrated into that process. I don't see the need for an independent login command.

tegefaulkes:
The intention is for it to login to PKE using your node. It's hard to read but I think we have a PK authenticate command that fills that role at the top of the plan. But in any case we need a secure way to tie a user session on the website to a PK node that you log in with.

CMCDragonkai:
This makes it seem like `tailscale`, `pulumi`, `docker`... etc.

1. What does it mean to login to `enterprise.polykey.com`?
2. What does it mean to login to `matrixai.enterprise.polykey.com`?
[...]

PR #423, opened 2025-06-18
Added `polykey auth login` command

aryanjassal:
### Description

This PR adds the `polykey auth login` command, which takes in an encoded JWT, signs it, and returns the new token. This is meant to verify the identity of the node, which would be as part of a returned field labelled `nodeId`.

### Issues Fixed

* Addresses #408 

### Tasks

- [x] 1. Create `polykey auth` subcommand
- [x] 2. Create `polykey auth login` command
- [x] 3. Integrate `AuthSignToken` handler from Polykey
- [x] 4. Write tests for this behaviour

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/src/auth/CommandLogin.ts
+++ b/src/auth/CommandLogin.ts
@@ -1,0 +1,109 @@
+import type PolykeyClient from 'polykey/PolykeyClient.js';
+import CommandPolykey from '../CommandPolykey.js';
+import * as binProcessors from '../utils/processors.js';
+import * as binUtils from '../utils/index.js';
+import * as binOptions from '../utils/options.js';
+
+class CommandLogin extends CommandPolykey {
+  constructor(...args: ConstructorParameters<typeof CommandPolykey>) {
+    super(...args);
+    this.name('login');
+    this.description('Login to a platform with Polykey identity');
+    this.argument('<token>', 'Token provided by platform for logging in');
+    this.addOption(binOptions.nodeId);
+    this.addOption(binOptions.clientHost);
+    this.addOption(binOptions.clientPort);
+    this.action(async (token, options) => {
+      const { default: PolykeyClient } = await import(
+        'polykey/PolykeyClient.js'
+      );
+      const { default: Token } = await import('polykey/tokens/Token.js');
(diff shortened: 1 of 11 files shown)


src/auth/CommandLogin.ts
```diff
@@ -0,0 +1,109 @@
...
+        if (!result.ok) {
+          throw new binErrors.ErrorPolykeyCLILoginFailed(
+            'Return url returned failure',
+            {
+              data: {
+                code: result.status,
+              },
+            },
+          );
+        }
```
tegefaulkes: Have the error message include the code that it received.

src/auth/CommandLogin.ts
```diff
@@ -0,0 +1,109 @@
...
+          token.split('.');
+        const incomingTokenEncoded = {
+          payload: payload as TokenPayloadEncoded,
+          signatures: [
+            {
+              protected: protectedHeader as TokenProtectedHeaderEncoded,
+              signature: signature as TokenSignatureEncoded,
+            },
+          ],
+        };
```
tegefaulkes: This is lacking basic parsing checks. This will fail with a bad error message if the token is ever misformed. WE should have better parsing and generator utility functions for this format.

tegefaulkes (changes requested):
Some small comments.

src/auth/CommandLogin.ts
aryanjassal: Do I remove it from the data or keep the error code in both the data and the message?

PS. `result.status` is the HTTP code it passed (or failed in this case) with

src/auth/CommandLogin.ts
aryanjassal: This was meant to be more for testing but I missed it in my glance over the code. I'll probably write a new parser for this.

src/auth/CommandLogin.ts
aryanjassal: I have removed the data block completely, instead embedding the status code in the message string itself.

PR #428, opened 2025-06-23
Aligned command with tailscale login flow

aryanjassal:
### Issues Fixed

* Addresses https://github.com/MatrixAI/Polykey-CLI/issues/408 (REF ENG-620)

### Tasks

- [x] 1. Update command to match RPC
- [x] 2. Update tests
- [x] 3. Add auto-open browser or optionally print the URL if browser can't be automatically detected

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased


src/auth/CommandLogin.ts
```diff
@@ -1,35 +1,24 @@
...
-import * as binParsers from '../utils/parsers.js';
 import * as binUtils from '../utils/index.js';
 import * as binOptions from '../utils/options.js';
-import * as errors from '../errors.js';
 
 class CommandLogin extends CommandPolykey {
   constructor(...args: ConstructorParameters<typeof CommandPolykey>) {
     super(...args);
     this.name('login');
     this.description('Login to a platform with Polykey identity');
```
shafiqihtsham: This should be `Login to a platform with your Polykey identity.', just for user clarity.

src/auth/CommandLogin.ts
```diff
@@ -58,51 +47,33 @@ class CommandLogin extends CommandPolykey {
...
 
         // Send the returned JWT to the returnURL provided by the initial token
         const compactHeader = binUtils.jsonToCompactJWT(response);
-        const incomingPayload =
-          tokensUtils.parseTokenPayload<IdentityRequestData>(payload);
-        let result: Response;
+        const targetURL = new URL(
+          url.endsWith('/') ? url.slice(0, url.length) : url,
+        );
+        const subPath: string = options.returnURLPath ?? '/oauth2/oidc';
```
shafiqihtsham, line 62: I don't think this check is needed anymore, because the command only expects the url now. So this can just be 
`const subPath = '/oauth2/oidc`. Or is this for future cases, where the user should also provide the specific endpoint as well?

src/auth/CommandLogin.ts
```diff
@@ -58,51 +47,33 @@ class CommandLogin extends CommandPolykey {
...
+        );
+
+        // Try to open the URL in the browser
         try {
-          result = await fetch(incomingPayload.returnURL, {
-            method: 'POST',
-            headers: { 'Content-Type': 'application/json' },
-            body: JSON.stringify({ token: compactHeader }),
-          });
+          process.stderr.write('Opening URL in browser...\n');
```
shafiqihtsham: This is all client facing code as the user will be reading this of course. So I think for clarity this should be changed to 
`Attempting to open this URL in your browser...` just so the flow of the command is easier to understand for the user.

tsconfig.json
```diff
@@ -13,7 +13,7 @@
     "allowSyntheticDefaultImports": true,
     "resolveJsonModule": true,
     "moduleResolution": "NodeNext",
-    "module": "ESNext",
+    "module": "NodeNext",
```
shafiqihtsham, line 16: I think I mentioned this already in https://github.com/MatrixAI/Polykey/pull/912#pullrequestreview-2960377506, but yeah I believe it should stay `ESNext`.

shafiqihtsham (changes requested):
Looks good. I just left some questions about things I was confused with. Just a bit of nitpicking as well with the command description and what is shown to the user. Otherwise this looks good to merge.

src/auth/CommandLogin.ts
aryanjassal: I did this for future use cases, where the endpoint isn't what is implemented in PKE. This was a small change so I felt I could sneak it in this PR.

tsconfig.json
aryanjassal: As mentioned in the linked comment, it doesn't make a practical impact on the application's runtime and I am no longer getting linter errors in my projects on VS Code, so I will keep this change as-is. If an issue is linked to this update, it can be reverted pretty easily, but the tests were not able to catch any regression, and neither could me using it (semi-) regularly.

aryanjassal:
Everything has been done for this PR and it is all working as intended. Merging.

